PR #13, opened 2022-10-09
add loaders config

Tianhao-Gu:


loaders/gtdb/gtdb_config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "LOAD_VERSION": 207,
+  "KBASE_COLLECTION": "gtdb",
+  "LOAD_FILES": [
+    "~/SCIENCE/taxonomy/gtdb/bac120_taxonomy_r207.tsv",
+    "~/SCIENCE/taxonomy/gtdb/ar53_taxonomy_r207.tsv"
+  ]
+}
```
MrCreosote, line 8: As an alternative option we discussed, given the relatively small number of inputs, just having CLI params rather than a config. What made you decide to go with a config file?

MrCreosote, line 8: I would strongly recommend something other than JSON for the config format - it sucks to edit for a human and doesn't have comments. YAML or TOML are much better

loaders/loaders_config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "production": {
+    "CONFIG_FILE": "~/SCIENCE/taxonomy/gtdb/load_config.toml",
```
MrCreosote, line 3: Whatever's needed from these config files should just be moved into the config, don't reference separate config files.

loaders/gtdb/gtdb_config.json
Tianhao-Gu: 👍 a CLI param should be the way to go.

Tianhao-Gu: 👍

loaders/loaders_config.json
Tianhao-Gu: 👍 make sense. updated config file.

loaders/gtdb/gtdb_config.json
MrCreosote: Ok, I'm confused then - if we're going to use CLI params do we need this PR at all?

Tianhao-Gu: hmmm, we probably still need a global config for DB connections. maybe some default params that go into config files?

MrCreosote: But if we use arangoimport we don't need DB connection config, right?

PR #14, opened 2022-10-11
add arango class

Tianhao-Gu:


MrCreosote (changes requested):
This looks fine other than
* Adding a test file that at least imports this file and runs a no-op test so that when we move on from prototyping well see this class needs tests
* Maybe this should go in /common rather than making another directory at the root?

MrCreosote:
Oh, pipfile.lock should be updated as well, missed that

MrCreosote:
I'm sort of wondering now if all the code should just go in `/src` rather than several top level dirs

Tianhao-Gu:
@MrCreosote 👍 addressed comments.

MrCreosote (approved):
LGTM

PR #15, opened 2022-10-14
restruct

Tianhao-Gu:


Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY ./ /kb/collections/
 
 # Write the git commit for the service
 ARG VCS_REF=no_git_commit_passed_to_build
-RUN echo "GIT_COMMIT=\"$VCS_REF\"" > common/git_commit.py
+RUN echo "GIT_COMMIT=\"$VCS_REF\"" > src/common/git_commit.py
```
MrCreosote, line 18: The entrypoint command will need an update as well

src/service/app.py
```diff
@@ -3,8 +3,8 @@
 '''
 import os
 
-from common.git_commit import GIT_COMMIT
-from common.version import VERSION
+from src.common.git_commit import GIT_COMMIT
+from src.common.version import VERSION
```
MrCreosote, line 7: I'm not sure whether the common python practice is to include src in the import path or not. If not, we'd need to set the PYTHONPATH to /src. 

Personally I don't really care either way as long as we're consistent.

MrCreosote (changes requested):
Just needs a docker file fix

Dockerfile
Tianhao-Gu: 👍

PR #17, opened 2022-10-18
First crack at a collections op design

MrCreosote:


design/core_collection_operations.md
```diff
@@ -0,0 +1,167 @@
...
+* Name (opaque string)
+* Version (opaque string)
+* Icon url
+  * Could probably use FastAPI's built in static server to serve these for now
+* Creation date
+* (If active) activation date
+* List of available data products with the active version of each data product (see example
+  below)
+  * Generally speaking, each data product will support a view in the collections UI
+  * The version information allows creating a new version of the collection document but only
```
MrCreosote, line 71: Maybe a schema version for the data should be specified as well

design/core_collection_operations.md
```diff
@@ -0,0 +1,167 @@
...
+    updating some of the associated data products, while leaving others at their current
+    version
+  * Similar to the collections version, it also allows for easy rollbacks, and any version
+    semantics are up to the user.
+
+```
+[
+    {
+        "product": "taxa_freq",
+        "version": "r207.kbase.3"
```
MrCreosote, line 81: `source_version` might also be useful, e.g. `r207` in this case

design/core_collection_operations.md
dakotablair: 👍

Tianhao-Gu (approved):
GLTM

design/core_collection_operations.md
MrCreosote: Added this

design/core_collection_operations.md
MrCreosote: Added this

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+
+* Collection information can change at any time, and previous versions are not necessarily
+  accessible.
+  * They should be accessible, at least, to database admins, and possibly via admin API methods.
+* Collections can theoretically be reverted to prior versions at any time.
+* We expect < 1000 collections over the lifetime of the collections infrastructure.
+* There may be many versions of collections as they are updated and adjusted over time.
+  * Including adding new data products to collections.
+* Data product data structures / UIs / etc for collections will usually, but not always,
+  be applicable to other collections as well.
```
dauglyon, line 16: What is an example of the "not always" case here?

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+  * Administration only operation.
+  * Once saved, a collection version is immutable and cannot be overwritten.
+    * Trying to save a collection with an existing version will result in an error.
+  * A collection version *is not* immediately active.
+    * At some point we may wish to supply admin endpoints for accessing alternate collection
+      versions.
+  * The user specifies the version, which to the collections API is an opaque string. Any
+    semantics are supplied by the user.
+    * Note that this means semantic versions will not sort in order - the DB and API would need
+      specific schemas / code to do that.
```
dauglyon, line 30: I don't love this. Would prefer to have some sort of ordering in the versions. Even if that means having a "version string" (collections-creating-user provided) and a "version number". I suppose storing the date a version was created could implicitly store the order, but I'm not sure how that would interact with reverted/rollbacked/pre-activated versions.

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+* Activate a collection version.
+  * Administration only operation.
+  * Can activate any version at any time.
+  * The active version is the only version accessible to users.
+  * Separating the saving and activation steps means we can save a collection document,
+    load data products for that collection, and then activate the collection when the data
+    products are ready.
+  * It also means we can trivially roll back changes.
+* Get a collection version.
+  * Public operation.
```
dauglyon, line 40: public-public or requiring kbase login?

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+  * Initially, this is only the active collection.
+    * At some point we may wish to supply admin endpoints for accessing alternate collection
+      versions.
```
dauglyon, line 43: should we just build this in from the beginning? It seems like a light lift now, but it could maybe become a heavier lift if we do it as a refactor later on

design/core_collection_operations.md
dauglyon: Actually, if we have "source version" as described below, I don't see why we shouldn't enforce strictly ordered versions

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+For a minimal start, more will be added later (e.g. data comparison methods, contributors, etc.)
+
+* ID (lowercase alpha only string, no punctuation other than underscores, otherwise opaque)
+* Name (opaque string)
+* Version (opaque string)
+* Source version (opaque string)
+  * The source version may be the same from collection version to collection version if the
+    same source data is reloaded to, say, fix bugs or add features.
+  * This implies a particular collection version should not mix data from different source
+    versions.
```
dauglyon, line 68: We should write this in big bold text somewhere for collections admins

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+* Icon url
+  * Could probably use FastAPI's built in static server to serve these for now
+* Creation date
+* (If active) activation date
+* List of available data products with the active version of each data product (see example
```
dauglyon, line 73: can we define "data product"? Is this the language we are going to use for the "widgets"/visualizations going forward

dauglyon: i.e. is a "data product" the schema/api for each of those views, or is a "data product" the cached data sent/consumed by the api/ui?

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+  * Similar to the collections version, it also allows for easy rollbacks, and any version
+    semantics are up to the user.
+  * In the future, we could add a schema version if the schema needs to change. For now, no
+    schema version implies a schema version of 1.
+    * An altenative would just be to add a version to the product name, like `taxa_freq2`.
+
+```
+[
+    {
+        "product": "taxa_freq",
```
dauglyon, line 88: is this the "data_product_id" specified in the endpoint below? if so should we maybe call that path argument "data_product_type"? Or is there going to be an ID field that is unique across all collections that is listed in the `/collection/{collection_id}` endpoint?

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+```
+GET /collection
+```
+
+#### Save a collection
+
+* Requires a token with the `COLLECTIONS_ADMIN` (or something) `auth2` role.
+
+```
+POST /collection/{collection_id}/version/{version}
```
dauglyon, line 146: how does one get the version ID for this call? Will `/collection` return a list? I'm not seeing that specified here. Alternatively we could have a `GET /collection/{collection_id}/versions`

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Data products
+
+* General endpoint structure, exact path and query params will depend on the data product
+
+```
+GET /collection/{collection_id}/data/{data_product_id}/...
+```
+* Extendable to allow getting a specific collection version (which specifies the data product
+  version) in the future.
```
dauglyon, line 175: super minor nit, plurals seem to be more common in rest APIs to help differentiate between "functional" endpoints (like `.../activate`) and "object" endpoints (like `/collections`):

````suggestion
### Endpoints

* Only includes immediately relevant endpoints.

#### List collections

* For now returns all active collections based on the < 1000 collections assumption.

```
GET /collections
```

#### Save a collection

* Requires a token with the `COLLECTIONS_ADMIN` (or something) `auth2` role.

```
POST /collections/{collection_id}/versions/{version}

BODY contains collection contents.
```

#### Activate a collection

* Requires a token with the `COLLECTIONS_ADMIN` (or something) `auth2` role.

```
PUT /collections/{collection_id}/versions/{version}/activate
```

#### Get a collection

```
GET /collections/{collection_id}
```

* Extendable to allow getting a specific version in the future.

#### Data products

* General endpoint structure, exact path and query params will depend on the data product

```
GET /collections/{collection_id}/data/{data_product_id}/...
```
* Extendable to allow getting a specific collection version (which specifies the data product
  version) in the future.
````

**Review by dauglyon — COMMENTED** [review]
minor stuff

File: design/core_collection_operations.md (hunk 3 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:16]: This is more of a CYA because I don't think I know enough at this point to assert that a particular data structure / UI / API for a data product is 100% going to work for all future collections

File: design/core_collection_operations.md (hunk 4 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:30]: The creation date of the version implies an ordering. The problem with ordering based on the version tag is that we then have to enforce a specific format for the version tag, which I think of as basically like a git tag. I'd prefer to allow the users of the service to choose their version format.

The rollback mechanism also makes having a version tag with an enforced order not make a lot of sense, since based on the design the version can jump around in the order at will.

> Actually, if we have "source version" as described below, I don't see why we shouldn't enforce strictly ordered versions
I'm not understanding the logic here.

File: design/core_collection_operations.md (hunk 5 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:40]: no login required at all

File: design/core_collection_operations.md (hunk 6 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:43]: No, since it's not required for the first demo UI. I'd expect this to happen sooner rather than later though. I'm not concerned about it being a heavy lift

File: design/core_collection_operations.md (hunk 7 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:68]: As I implement endpoints I'll add stuff like this to the docs for sure

File: design/core_collection_operations.md (hunk 8 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:73]: Data product is the vocabulary I've been using, but I'm not attached to it at all and if there's better terminology I'm all for it.

> i.e. is a "data product" the schema/api for each of those views, or is a "data product" the cached data sent/consumed by the api/ui?

Yes to both; I think of it as the data generated from the source data that supports the UI, the schema for that data in Arango, the API, and the UI

Comment (MrCreosote) [reply design/core_collection_operations.md:73]: I.e. the full stack for a particular widget

File: design/core_collection_operations.md (hunk 10 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:146]: The user supplies the version

File: design/core_collection_operations.md (hunk 9 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:88]: > is this the "data_product_id" specified in the endpoint below?

Right

> if so should we maybe call that path argument "data_product_type"? 

Hmm, what abour just `data_product`?

File: design/core_collection_operations.md (hunk 11 above)
Comment (MrCreosote) [reply design/core_collection_operations.md:175]: I don't really care one way or the other, but it looks like the plurals have it: https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources

File: design/core_collection_operations.md (hunk 4 above)
Comment (dauglyon) [reply design/core_collection_operations.md:30]: Having an easy way to say "this version is before this other one" without passing dates around // display dates to users is quite a convenience boost client-side.

> I'm not understanding the logic here.

Letting the users set the version string arbitrarily makes sense to me from a perspective of "trying to have the version name match the version of the source data", but if we're going to store the source version anyway (i.e. in "source version"), I don't see what benefit we get from arbitrary version names. It seems like we would get more benefit from some sort of sequential versioning scheme, unless I'm missing something.

semver in our collections could provide admins nice visual/scannable inidcators when working with versions. Totally rough-draft spitballing off that, semver would enable guidelines something like the following.
```plaintext
2.11.3
^ ^  ^
| |  |
| |  Increment when a change which does not effect data content is made (description, name, icon) changes
| Increment when collection data content changes, but not the source data version (e.g. updated ETL for that collection)
Increment when collection data content changes to reflect a new source version
```

design/core_collection_operations.md
dauglyon: > Hmm, what abour just data_product?
perfect

design/core_collection_operations.md
dauglyon: how does the user query for list of what versions exist, so that they might supply it?

MrCreosote: They don't provide an existing version, and for now the user could just go to the arango aardvark API if they want to be sure the version they supply doesn't exist

design/core_collection_operations.md
MrCreosote: anyway, done

design/core_collection_operations.md
MrCreosote: done

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+BODY contains collection contents.
+```
+
+#### Activate a collection
+
+* Requires a token with the `COLLECTIONS_ADMIN` (or something) `auth2` role.
+
+```
+PUT /collections/{collection_id}/versions/{version}/activate
+```
```
dauglyon, line 157: Sorry, more nits I just spotted. `PUT` doesn't feel appropriate here as running `GET` on the same URL will not return the "PUTted" resource. More of a GET or a POST (if a body is required) IMO

MrCreosote: I'm using PUT because PUTs are idempotent and POSTS are not. GET isn't appropriate because the endpoint can result in a state change.

design/core_collection_operations.md
```diff
@@ -0,0 +1,175 @@
...
+GET /collections
+```
+
+#### Save a collection
+
+* Requires a token with the `COLLECTIONS_ADMIN` (or something) `auth2` role.
+
+```
+POST /collections/{collection_id}/versions/{version}
```
dauglyon, line 146: `PUT` _is_ actually more appropriate here, as `GET /collections/{collection_id}/versions/{version}` would be our likely way to fetch a specific collection version, and restful `POST`s usually generate IDs/resource-locations whereas `PUT` typically explicitly specify them

MrCreosote: This operation isn't idempotent, so a PUT isn't correct here

dauglyon: This looks pretty idempotent to me, we should discuss

design/core_collection_operations.md
dauglyon: `PUT` is pretty explicitly supposed to replace the contents at a resource. `GET` feels most appropriate here, excluding the elusive `PATCH`, which is probably correct but tended not to be well supported as-of the last time I worked on a REST API.

design/core_collection_operations.md
dauglyon: see https://stackoverflow.com/questions/51953943/what-is-idempotency-in-http-put-can-i-disallow-overwriting-of-a-resource

unless I'm misunderstanding what the endpoint does

design/core_collection_operations.md
MrCreosote: > GET feels most appropriate here

The problem with a GET here, as I mentioned, is that this causes a state change, which GET requests are not supposed to do

dauglyon: update, the more I think about `GET` the less correct it feels

design/core_collection_operations.md
MrCreosote: Huh, TIL. https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

That doesn't match with the standard CS definition of idempotent as I understand it, which is confusing if so.

Fair enough then.

design/core_collection_operations.md
dauglyon: Could there be a case for "deactivating" the active version (so as to hide a collection) with `PUT /collections/{collection_id}/versions/{version}/activate`?

Also, potentially in the future: `GET /collections/{collection_id}/versions/{version}/activate` returns a bool
seems fine-ish? 

happiest path seems to be:
```HTTP
PUT /collections/{collection_id}/versions/{version}/activate
Content-Type: application/json

<boolean>
```
```
200 OK
```
Later:
```HTTP
GET /collections/{collection_id}/versions/{version}/activate
```
``` 
200 OK

<boolean>
```

design/core_collection_operations.md
dauglyon: Yeah... AFAIK web folks tend to use idempotent to mean `idempotent with respect to the recipient` and ignore the existence of the sender

design/core_collection_operations.md
MrCreosote: By process of elimination:
HEAD, OPTIONS - yeah, no
GET - this op does a state change, so no
PATCH - means we're replacing part of an entity, which is not what's happening here
POST - means we're creating a new entity, which is not what's happening here
PUT - means we're replacing an entity, which is not what's happening here exactly, but we *are* replacing the *active* collection entity

To me PUT looks like the least bad choice

MrCreosote: > Could there be a case for "deactivating" the active version (so as to hide a collection) with PUT

Hmm, my KJR is that once a collection is published it shouldn't be unpublished... if this is an operation we want I think this probably needs discussion with the PO.

And wouldn't it be a DELETE?

MrCreosote: > Also, potentially in the future: GET /collections/{collection_id}/versions/{version}/activate returns a bool
seems fine-ish?

I'm not sure what the purpose of that would be, since the active collection (and version) can be retrieved via
GET /collections/{collection_id}

MrCreosote: I'm not sure what the purpose of returning a bool from the PUT is - it'll always be `true` unless there's an error, in which case the user will get an entirely different error structure

design/core_collection_operations.md
MrCreosote: resolving per @dauglyon

design/core_collection_operations.md
MrCreosote: resolve as per @dauglyon

design/core_collection_operations.md
MrCreosote: Resolution per out of band conversations:
* We'll go with PUT
* We won't worry about the deactivation stuff (including PUT bodies, GET etc.) for now and cross that bridge when we come to it

design/core_collection_operations.md
MrCreosote: Resolution per out of band conversations:

In the future we should have a GET version of this endpoint so a UI can check that a version exists before trying to POST and potentially failing. This is already mentioned in the doc so no changes needed at this point. As mentioned above, for the first demo version we can just check the DB directly.

design/core_collection_operations.md
MrCreosote: Per out of band discussions, I'm going to write up a doc presenting options for handling collection versions, run it past @dauglyon, and then have the full team discuss it.

MrCreosote: Added the doc and a decision, so closing this now

Tianhao-Gu (approved):
I am okay with merging this PR.

dauglyon (approved):

PR #19, opened 2022-10-18
Add basic infrastructure for auth & service

MrCreosote:
input checkers, errors, user ID, etc.

Some of this stuff maybe should be moved to common but we can cross that bridge when we come to it.


Tianhao-Gu (approved):
mostly from sample_service, LGTM

PR #20, opened 2022-10-19
Add auth client

MrCreosote:
```
In [1]: with open('/home/<user>/.kbase_token_ci') as f:
   ...:     token = f.read().strip()
   ...:

In [2]: from src.service.kb_auth import KBaseAuth

In [3]: ka = KBaseAuth("https://ci.kbase.us/services/auth", ["ws_read_admin"], c
   ...: ache_expiration=5)

In [4]: ka.get_user(token)
Out[4]: (<AdminPermission.FULL: 10>, UserID("gaprice"))

In [8]: ka = KBaseAuth("https://ci.kbase.us/services/auth", ["ws_fake_admin"], c
   ...: ache_expiration=5)

In [9]: ka.get_user(token)
Out[9]: (<AdminPermission.NONE: 1>, UserID("gaprice"))
```

### DIFF
--- a/src/service/kb_auth.py
+++ b/src/service/kb_auth.py
@@ -5,7 +5,15 @@ A client for the KBase Auth2 server.
 # Mostly copied from https://github.com/kbase/sample_service with a few tweaks.
 # TODO make a KBase auth library?
 
+from cacheout.lru import LRUCache
 from enum import IntEnum
+import logging
+import requests
+import time
+from typing import List, Sequence, Tuple
+
+from src.service.arg_checkers import not_falsy as _not_falsy
+from src.service.user import UserID
 
 
 class AdminPermission(IntEnum):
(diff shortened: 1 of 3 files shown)


src/service/kb_auth.py
```diff
@@ -16,3 +24,89 @@ class AdminPermission(IntEnum):
...
     FULL = 10
 
+
+class KBaseAuth:
+    ''' A client for contacting the KBase authentication server. '''
+
+    def __init__(
+            self,
+            auth_url: str,
+            full_admin_roles: List[str] = None,
```
Tianhao-Gu, line 34: shouldn't we make the default `full_admin_roles` to some fixed set just in case someone inits this class with a wrong `full_admin_roles` list?

Tianhao-Gu (approved):
just have one question. But overall LGTM.

src/service/kb_auth.py
MrCreosote: Not totally sure I understand what you're asking, but `full_admin_roles` should be configurable, so they need to be an input arg.

As to why they need to be configurable, we may wish to rename roles, or add new roles that work for several (but not all services). If they're hardcoded we can't do that.

I'm going to go ahead and merge since you approved, but if you want to keep talking about it we can do it here or in slack and make any changes we come up with in a future PR

MrCreosote: Another option is just returning the roles and letting the caller figure out what do to with them I guess

PR #21, opened 2022-10-19
Make uvicorn receive SIGTERM

MrCreosote:
Never really understood why there were two syntaxes for ENTRYPOINT, and now I really don't understand since one of them will cause signals to not make it to your application.

With the ENTRYPOINT change the container now shuts down immediately rather than waiting for 10 seconds and getting a SIGKILL.

### DIFF
--- a/src/loaders/gtdb/gtdb_taxa_freq_loader.py
+++ b/src/loaders/gtdb/gtdb_taxa_freq_loader.py
@@ -91,8 +91,8 @@ def main():
                         help='GTDB taxonomy files')
 
     # Required flag argument
-    parser.add_argument('--load_version', required=True, type=str, nargs=1,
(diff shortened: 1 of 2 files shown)


src/loaders/gtdb/gtdb_taxa_freq_loader.py
```diff
@@ -91,8 +91,8 @@ def main():
                         help='GTDB taxonomy files')
 
     # Required flag argument
-    parser.add_argument('--load_version', required=True, type=str, nargs=1,
-                        help='GTDB load version')  # TODO parse from load_files name
```
MrCreosote, line 95: I don't think we actually want to do this TODO since the load versions could be things like
```
207-dev
207.kbase.1
207.kbase.2
```

Tianhao-Gu (approved):
LGTM

PR #25, opened 2022-10-21
Don't create app as a global

MrCreosote:
Instead use a factory method. This has a couple of benefits:

1. The app isn't created on import
2. in the future we could potentially pass mocks into the create_app function to mock out app dependencies.


Tianhao-Gu (approved):
👍

PR #28, opened 2022-10-24
Add configuration handling for service

MrCreosote:


src/service/kb_auth.py
```diff
@@ -59,7 +59,7 @@ async def _check_error(r):
             j = await r.json()
         except Exception:
             err = ('Non-JSON response from KBase auth server, status code: ' +
-                    str(r.status_code))
+                    str(r.status))
```
MrCreosote, line 62: oopc

Tianhao-Gu (approved):
minimal test is required as agreed.

PR #31, opened 2022-10-26
Add collection pydantic model

MrCreosote:
very slightly tested manually for now. If we start finding too many bugs when running the service I'll add a test suite.

### DIFF
--- a/src/service/models.py
+++ b/src/service/models.py
@@ -1,0 +1,149 @@
+"""
+Global models for the service that cross API and storage boundaries. These models define
+both
+
+* The structure and keys accepted by and returned to the API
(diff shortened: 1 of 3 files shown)


src/service/models.py
```diff
@@ -0,0 +1,149 @@
+"""
```
Tianhao-Gu, line 1: structurally it looks good to me. But without tests, I didn't examine the logic for each op.

Tianhao-Gu (approved):
it looks good to me. But again, without tests, I did examine the code logic.

src/service/models.py
MrCreosote: Not quite sure what you mean - you can check the logic of the code without tests...

PR #32, opened 2022-10-27
RE2022-52: Add start of arango wrapper

MrCreosote:
```
In [1]: from src.service.storage_arango import create_storage

In [2]: import aioarango

In [3]: cli = aioarango.ArangoClient(hosts='http://localhost:8529')

In [4]: db = await cli.db("collections_test")

In [5]: arst = await create_storage(db, create_collections_on_startup=True)

In [6]: await arst.get_next_version("gtdb")
Out[6]: 1

In [7]: await arst.get_next_version("gtdb")
Out[7]: 2

In [8]: await arst.get_next_version("gtdb")
Out[8]: 3

In [9]: arst = await create_storage(db, create_collections_on_startup=True)

In [10]: await arst.get_next_version("gtdb")
Out[10]: 4
```


src/service/storage_arango.py
```diff
@@ -0,0 +1,103 @@
...
+
+    db - the database where the data is stored. The DB must exist.
+    create_collections_on_startup - on starting the wrapper, create all collections and indexes
+        rather than just checking for their existence. Usually this should be false to
+        allow for system administrators to set up sharding to their liking, but auto
+        creation is useful for quickly standing up a test service.
+    """
+    if create_collections_on_startup:
+        # TODO DB create indexes and collections
+        await _create_collection(db, _COLL_COUNTERS)  # no indexes necessary
```
Tianhao-Gu, line 49: should `_create_collection` and `_check_collection_exists` take a list of default collections? or `kbcoll_coll_counters` is the only collection we care about?

src/service/storage_arango.py
```diff
@@ -0,0 +1,103 @@
...
+        raise ValueError(f"Collection {col_name} does not exist")
+
+async def _create_collection(db: StandardDatabase, col_name: str):
+    """
+    Create a collection, ignoring a duplicate name error
+    """
+    try:
+        await db.create_collection(col_name)
+    except CollectionCreateError as e:
+        if not e.error_code == 1207:  # duplicate name error code
```
Tianhao-Gu, line 66: it's ok. the error code is for `name duplicate`. but why not check collection existence before calling create_collection? That will be easier to read than looking for the error code.

src/service/storage_arango.py
```diff
@@ -0,0 +1,103 @@
...
+        # TODO DB make some sort of pluginish system for data products
+        # TODO NOW document endpoint in new endpointsfile (in /docs/endpoint.md)
+
+    async def get_next_version(self, collection_id: str) -> int:
+        """ Get the next available version number for a collection. """
+        bind_vars = {
+            _FLD_COLLECTION_ID: collection_id,
+            f"@{_FLD_COLLECTION}": _COLL_COUNTERS
+        }
+        cur = await self._db.aql.execute(_QUERY_GET_NEXT_VERSION, bind_vars=bind_vars)
```
Tianhao-Gu, line 85: I didn't try but if `collection_id` doesn't exist, will `cur.next()` still work?

Tianhao-Gu (approved):
overall looks good to me. Just a few questions/comments.

src/service/storage_arango.py
MrCreosote: I'll add collections as needed. Currently I'm just adding a method call for each collection since there's only 3 total in my WIP at this point. We could change it to a list later if needed.

src/service/storage_arango.py
MrCreosote: Yep, the INSERT part of the UPSERT gets run if that's the case

src/service/storage_arango.py
MrCreosote: Because since it's two steps, it'd be a race condition, like how you shouldn't check for a file's existence and then open the file. I'd prefer it if the API allowed a `silent_if_exists` or something option, but there isn't one.

It's not a race condition that is likely to occur, but still, better to keep to best practices.

MrCreosote: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

MrCreosote: Also just one http connection to arango instead of 2

src/service/storage_arango.py
Tianhao-Gu: 👍

PR #37, opened 2022-10-31
Option to not connect to ext services

MrCreosote:
also close arango client properly

### DIFF
--- a/src/service/app.py
+++ b/src/service/app.py
@@ -23,7 +23,7 @@ from src.service import models
 from src.service.arg_checkers import contains_control_characters
 from src.service.config import CollectionsServiceConfig
 from src.service.http_bearer import KBaseHTTPBearer, KBaseUser
-from src.service.storage_arango import create_storage
+from src.service.storage_arango import create_storage, ARANGO_ERR_NAME_EXISTS, ArangoStorage
 
 
 # TODO LOGGING - log all write ops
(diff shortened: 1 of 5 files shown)


MrCreosote:
The codecov UI has really become unreliable, it's a different problem almost every time I go there
image

src/service/app.py
```diff
@@ -60,37 +60,52 @@ def create_app(noop=False):
...
+                await sysdb.create_database(cfg.arango_db)
+            except aioarango.exceptions.DatabaseCreateError as e:
+                if e.error_code != ARANGO_ERR_NAME_EXISTS:  # ignore, db exists
+                    raise
+        db = await _get_arango_db(cli, cfg.arango_db, cfg)
+        storage = await create_storage(db, create_collections_on_startup=cfg.create_db_on_startup)
+        return cli, storage
+    except:
+        await cli.close()
+        raise
```
Tianhao-Gu, line 103: will it provide us with enough traceback to troubleshoot?

Tianhao-Gu (approved):
Just have a few questions. But overall looks good to me.

src/service/app.py
```diff
@@ -23,7 +23,7 @@
 from src.service.arg_checkers import contains_control_characters
 from src.service.config import CollectionsServiceConfig
 from src.service.http_bearer import KBaseHTTPBearer, KBaseUser
-from src.service.storage_arango import create_storage
+from src.service.storage_arango import create_storage, ARANGO_ERR_NAME_EXISTS, ArangoStorage
```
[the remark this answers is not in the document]
MrCreosote: The reason for `create_storage` is that async operations need to happen at the initialization of `ArangoStorage`, but __init__ can't be aysnc, so it has to be a different function. I happened to make it a module function, but it just as easily could be a `@classmethod` so the signature would be something like `ArangoStorage.create()`. To me it doesn't make a big difference, but if you think it's important I have no issues swapping to a class method

src/service/app.py
MrCreosote: I don't understand the concern - the `raise` will raise the exact same error the was caught so we're not losing any info...?

MrCreosote:
Since you approved I'm going to go ahead and merge, but we can continue with the conversations and I'll make any changes coming out of that in a later PR

src/service/app.py
Tianhao-Gu: yea. yea. you are right.

MrCreosote:
Oop, you beat me to it :P

src/service/app.py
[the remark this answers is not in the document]
MrCreosote: I actually think the classmethod is a bit more readable on second thought, I'll probably do that at some point

PR #38, opened 2022-10-31
Document error structure

MrCreosote:
Also handle 404 errors correctly


MrCreosote:
Tests failed due to codecov upload failing, what the heck is going on with them

Tianhao-Gu (approved):
👍  rerun tests for 4 times but codecov still failed.

PR #40, opened 2022-10-31
Add collection version activation routes

MrCreosote:
Turns out that getting the collection version was just a suboperation of activation, so added those too


Tianhao-Gu (approved):
💯

PR #44, opened 2022-10-31
Add list versions route

MrCreosote:


src/service/storage_arango.py
```diff
@@ -210,4 +224,43 @@ async def get_collection_version_by_num(self, collection_id: str, ver_num: int
...
+            bind_vars[_FLD_VER_NUM] = max_ver
+            aql += f"""
+                    FILTER d.{models.FIELD_VER_NUM} <= @{_FLD_VER_NUM}
+                    """
+        aql += f"""
+                    SORT d.{models.FIELD_VER_NUM} DESC
+                    LIMIT @{_FLD_LIMIT}
+                    RETURN d
+                """
+        cur = await self._db.aql.execute(aql, bind_vars=bind_vars)
```
Tianhao-Gu, line 264: is the query tested locally?

src/service/storage_arango.py
```diff
@@ -210,4 +224,43 @@ async def get_collection_version_by_num(self, collection_id: str, ver_num: int
...
+        List versions of a collection, sorted from highest to lowest.
+
+        collection_id - the ID of the collection.
+        max_ver - the maximum version to return. This can be used to page through the results.
+        limit - the maximum number of versions to return. Default and maximum value is 1000.
+        """
+        if not limit or limit < 1:
+            limit = 1000
+        if limit > 1000:
+            raise errors.IllegalParameterError("Limit must be <= 1000")
```
Tianhao-Gu, line 242: I am okay with raise. But should we consider doing `limit = min(limit, 1000)`?

Tianhao-Gu (approved):
Looks good overall. But few questions.

src/service/storage_arango.py
MrCreosote: I tested it via running the server via `docker-compose up`, and also modified the code to hardcode a limit and tested again. Is that what you were looking for?

Tianhao-Gu: okay. I think that's good for now. I just didn't test it locally.  In case it errors.

src/service/storage_arango.py
MrCreosote: Hmm, I'm ok with modifying a user's inputs when it's clear they're just being lazy or utterly wrong (like 0 or negative limits), but a user might not read the documentation correctly and think that specifying a higher limit is ok, and then think they got all the results. In that case I think it's better to expressly tell them the input is invalid

Tianhao-Gu: 👍

Tianhao-Gu (approved):

src/service/storage_arango.py
MrCreosote: That being said, the limit isn't exposed in the server api for now

PR #47, opened 2022-11-01
Split up collections router

MrCreosote:
Will allow adding data products endpoints before the collections admin endpoints

Also simplifies method metadata declarations slightly


MrCreosote:
Codecov upload failing again

Tianhao-Gu (approved):

PR #49, opened 2022-11-01
Add endpoints to get collection IDs

MrCreosote:
No way to figure out what inactive collections are available otherwise

### DIFF
--- a/src/service/routes_collections.py
+++ b/src/service/routes_collections.py
@@ -107,6 +107,13 @@ class CollectionsList(BaseModel):
     colls: list[models.ActiveCollection]
 
 
+class CollectionIDs(BaseModel):
+    colids: list[str] = Field(
+        example=["GTDB", "PMI"],
+        description="A list of collection IDs"
+    )
+
+
 class CollectionVersions(BaseModel):
     counter: int = Field(
         example=42,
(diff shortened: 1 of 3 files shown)


src/service/routes_collections.py
```diff
@@ -192,6 +205,13 @@ async def save_collection(
     return sc
 
 
+@ROUTER_COLLECTIONS_ADMIN.get("/collectionids/all/", response_model = CollectionIDs)
```
MrCreosote, line 208: I'm not totally sold on the endpoint uris. If you have a better idea let me know

src/service/storage_arango.py
```diff
@@ -27,23 +27,6 @@
 ARANGO_ERR_NAME_EXISTS = 1207
 _ARANGO_ERR_UNIQUE_CONSTRAINT = 1210
 
-_QUERY_GET_NEXT_VERSION = f"""
```
MrCreosote, line 30: Easier to read the methods if the queries are inline. If the same aql is used for multiple methods then a module global might make sense

src/service/storage_arango.py
```diff
@@ -123,15 +107,22 @@ async def create(
 
     def __init__(self, db: StandardDatabase):
         self._db = db
-        # TODO DB make some sort of pluginish system for data products
```
MrCreosote, line 126: Duplicated by a TODO in app.py

MrCreosote:
*)&*))8&(*^&^) codecov

Tianhao-Gu (approved):
👍

PR #57, opened 2022-11-02
RE2022-55: Document collection names / fields in `common`

MrCreosote:
Need to have the loaders and service speak the same language wrt collection names and fields


Tianhao-Gu (approved):
👍

PR #58, opened 2022-11-02
RE2022-55: Support optional auth

MrCreosote:


src/service/app.py
```diff
@@ -134,10 +134,6 @@ def _format_error(
 
 
 
-# TODO REFACTOR auth handling. Calling check_admin anyway, might was well do it all there
-# rather than with an overly complex Depends... but that's needed for OpenAPI to know about
-# the header. hmm. Also may need to be optional for some cases, now it's all or nothing
-# https://stackoverflow.com/questions/70926257/how-to-pass-authorization-header-from-swagger-doc-in-python-fast-api
```
MrCreosote, line 140: AFAICT the only way to get the OpenAPI docs to allow setting a token is via the dependency mechanism we're currently using, and since we're already using that we might as well leave the header processing in KBaseHTTPBearer. So this PR only adds optional auth to the header processing.

MrCreosote:
Codecov seems to be confused. There were only 2 lines added, so +18 hits is impossible

Tianhao-Gu (approved):

PR #61, opened 2022-11-03
RE2022-55: Add a minimal framework for adding data products

MrCreosote:
This will almost certainly change as I implement the first data product, but it's a reasonable starting point

### DIFF
--- a/src/common/storage/collection_and_field_names.py
+++ b/src/common/storage/collection_and_field_names.py
@@ -48,10 +48,10 @@ COLL_SRV_ACTIVE = _SRV_PREFIX + "active"
 
 ## Data product collections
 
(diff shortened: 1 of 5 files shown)


src/common/storage/collection_and_field_names.py
```diff
@@ -48,10 +48,10 @@
 
 ## Data product collections
 
-### Taxa frequency
+### Taxa counts
```
MrCreosote, line 51: The data product doesn't actually have frequency info, counts is the accurate terminology

Tianhao-Gu: 👍

Tianhao-Gu (approved):

PR #62, opened 2022-11-03
RE2022-55: Add support for data products to arango wrapper

MrCreosote:
Not really tested other than making sure the server starts up and creates indexes correctly. Will test when a DP is added


Tianhao-Gu (approved):
👍

PR #67, opened 2022-11-05
RE2022-55: Allow overriding taxa count load version

MrCreosote:


src/service/data_products/common_functions.py
```diff
@@ -0,0 +1,45 @@
+"""
```
MrCreosote, line 1: I had to split out the functions into a separate module from the common modules, as this function imports ArangoStorage, but that module imports stuff from the common data products module, causing a circular import

src/service/data_products/taxa_count.py
```diff
@@ -55,12 +65,20 @@ class Ranks(BaseModel):
...
+async def get_ranks(
+    r: Request,
+    collection_id: str = PATH_VALIDATOR_COLLECTION_ID,
+    load_ver_override: str | None = QUERY_VALIDATOR_LOAD_VERSION_OVERRIDE,
+    user: KBaseUser = Depends(_OPT_AUTH)
+):
     store = app_state.get_storage(r)
-    ac = await store.get_collection_active(collection_id)
-    load_ver = get_load_version(ac, ID)
+    load_ver = await get_load_version(store, collection_id, ID, load_ver_override, user)
```
Tianhao-Gu, line 81: can you remind a use case when an admin wants to use an overridden load_ver?

Tianhao-Gu (approved):
💯

src/service/data_products/taxa_count.py
MrCreosote: Without an override, the data is only viewable in the API for active collections, so only one load version of the data is viewable per collection. If you load new data, you can override the load versions to see it.

I considered allowing specification of a collection version instead, but that just effectively overrides the load version as well, so I figured we might as well allow that directly so we can view data that doesn't even have an inactive collection yet.

PR #70, opened 2022-11-07
Rename genome_stats -> genome_attribs

MrCreosote:
Attributes is a more general name; stats doesn't really fit what's actually going in the DB


MrCreosote:
codecov causing the tests to fail again

Tianhao-Gu (approved):

PR #71, opened 2022-11-08
Add genome_attribs data product

MrCreosote:
Needed to unmunge the loader keys as arango doesn't like whitespace in keys

### DIFF
--- a/src/loaders/gtdb/gtdb_genome_attribs_loader.py
+++ b/src/loaders/gtdb/gtdb_genome_attribs_loader.py
@@ -5,6 +5,7 @@ import pandas as pd
 import gtdb_genome_attribs_helper as helper
 from gtdb_loader_helper import convert_to_json
 
+import src.common.storage.collection_and_field_names as names
 from src.common.hash import md5_string
 from src.common.storage.collection_and_field_names import (
     FLD_ARANGO_KEY,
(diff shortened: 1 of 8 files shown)


src/loaders/gtdb/gtdb_genome_attribs_loader.py
```diff
@@ -100,11 +100,7 @@ def _rename_col(df, header_mapper):
          None (updates dataframe in place)
     """
 
-    # default mapper (e.g. {'aa_bb_cc': 'Aa Bb Cc'})
-    # Note: all columns are following 'a_b_c' format in the original GTDB metadata file
-    mapper = {col: ' '.join(list(map(lambda x: x.capitalize(), col.split('_')))) for col in df.columns}
```
Tianhao-Gu, line 105: I am okay not doing this. I was thinking when the attribute table was displayed, the header will be presented as `Genome Name` rather than `genome_name`.  If we store attri name as `genome_name`, will it add more work for the front end?

Tianhao-Gu (approved):
👍 just one question.

src/loaders/gtdb/gtdb_genome_attribs_loader.py
MrCreosote: It will, but we have to store the keys without spaces or AQL breaks. We could munge the keys every time when returning from the DB, but IMO we should let the front end decide how to display the data

If the FE devs complain, we can figure something else out - maybe we do something similar to `ranks` for the taxa counts and add an endpoint that returns a header mapper. Let's see how the plain old keys work first, maybe we're fixing a problem that doesn't exist

Tianhao-Gu: ah right. that makes sense.

PR #72, opened 2022-11-08
do not keep NA value

Tianhao-Gu:


MrCreosote (approved):
I'm not familiar with that option, but assuming it means we don't interpret strings as NaN regardless of their contents this LGTM

Tianhao-Gu:
> 

yup. that's exactly what it does.

PR #77, opened 2022-11-11
Just use built in defaults vs coding them

MrCreosote:
duh.


dauglyon (approved):

dauglyon:
Pretty straightforward, but it looks like I don't count as an approving reviewer

MrCreosote:
now you do

PR #79, opened 2022-11-12
Add an optional table format for genome attributes return

MrCreosote:
Default to table since the payload is ~1/2 the size

### DIFF
--- a/src/service/data_products/genome_attributes.py
+++ b/src/service/data_products/genome_attributes.py
@@ -16,6 +16,7 @@ from src.service.data_products.common_models import (
 from src.service.http_bearer import KBaseHTTPBearer, KBaseUser
 from src.service.routes_common import PATH_VALIDATOR_COLLECTION_ID
 from src.service.storage_arango import ArangoStorage, remove_arango_keys
+from typing import Any
 
 # Implementation note - we known FLD_GENOME_ATTRIBS_GENOME_NAME is unique per collection id / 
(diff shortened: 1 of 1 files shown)


src/service/data_products/genome_attributes.py
```diff
@@ -140,11 +158,24 @@ async def get_ranks(
...
+        if output_table:
+            data.append([d[k] for k in sorted(_remove_keys(d))])
+        else:
+            data.append({k: d[k] for k in sorted(_remove_keys(d))})
+    fields = []
+    if d:
+        if sort_on not in d: 
+            raise errors.IllegalParameterError(
+                f"No such field for collection {collection_id} load version {load_ver}: {sort_on}")
+        fields = [{"name": k} for k in sorted(d)]
```
Tianhao-Gu, line 177: why not just assign `fields = sorted(d)`?

Tianhao-Gu (approved):
just have one question. Otherwise, looks good to me.

src/service/data_products/genome_attributes.py
dauglyon: I think this is to allow adding other field properties (i.e. sort/filter type, human-readable name) in the future without it being a breaking change (correct me if I'm wrong @MrCreosote)

MrCreosote: right

Tianhao-Gu: 👍

PR #81, opened 2022-11-22
download ncbi files

Tianhao-Gu:


MrCreosote:
I think the file directories should be structured in a somewhat different way - to me it doesn't make sense to have the kbase collection and load version as part of the folder path since multiple collections could be using NCBI genomes as the source, and a particular load version for a collection could use the exact same genomes as a different load version, but differ only by post download calculations.

I think there are two routes we could take here:
1) Just have an NCBI folder where we stick all the NCBI genomes and then hard/soft link to create collections in different directories.
2) Since we're dealing primarily with GTDB, download only the GTDB genomes and store them in folders by the GTDB release version. For other collections, hard/soft link as usual, but for the GTDB collection just link the directory.

Right now I lean towards 2) since it simplifies things for our use case

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,75 @@
...
+PROTOTYPE - Download genome files from NCBI FTP server.
+
+usage: ncbi_downloader.py [-h] --download_file_ext
+                          DOWNLOAD_FILE_EXT
+                          [DOWNLOAD_FILE_EXT ...] --release_ver
+                          {207,202,95,89,86,83,80}
+                          [--root_dir ROOT_DIR]
+                          [--collection COLLECTION]
+                          [--threads THREADS]
+                          [--chuck_size CHUCK_SIZE] [--overwrite]
```
MrCreosote, line 11: `s/chuck/chunk/g`

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,75 @@
...
+
+optional arguments:
+  --root_dir ROOT_DIR   Root directory. (default: /global/cfs/cdi
+                        rs/kbase/collections/sourcedata)
+  --collection COLLECTION
+                        Collection (default: GTDB)
+  --threads THREADS     Number of threads. (default: half of
+                        system cpu count)
+  --chuck_size CHUCK_SIZE
+                        Number of genomes per thread
```
MrCreosote, line 31: This is confusing - why not just always distribute evenly rather than allowing the user to do something weird?

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,75 @@
...
+/global/cfs/cdirs/kbase/collections/sourcedata/ -> GTDB -> [GTDB_release_version] -> [genome_id] -> genome files
 
+e.g.
+/global/cfs/cdirs/kbase/collections/sourcedata/GTDB -> r207 -> GB_GCA_000016605.1 -> genome files
+                                                            -> GB_GCA_000200715.1 -> genome files
+                                                       r202 -> GB_GCA_000016605.1 -> genome files
+                                                            -> GB_GCA_000200715.1 -> genome files
+                                                       r80  -> GB_GCA_000016605.1 -> genome files
+                                                            -> GB_GCA_000200715.1 -> genome files
+
```
MrCreosote, line 49: Presumably at some point we'll want to update this script to softlink to earlier GTDB versions when the genome is the same

MrCreosote (changes requested):
The way the output is structured looks good, but I have some questions about the input args and how things are processed

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: And what happens when chunk_size * threads is < genome_count?

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,75 @@
...
+  --download_file_ext DOWNLOAD_FILE_EXT [DOWNLOAD_FILE_EXT ...]
+                        Download only files that match given
+                        extensions.
+  --release_ver {207,202,95,89,86,83,80}
+                        GTDB release version
+
+optional arguments:
+  --root_dir ROOT_DIR   Root directory. (default: /global/cfs/cdi
+                        rs/kbase/collections/sourcedata)
+  --collection COLLECTION
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: removed chuck/chunk

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: 👍

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -34,18 +94,95 @@ def _download_genome_file(download_dir: str, gene_id: str, target_file_ext: list
...
-def download_genome_files(gene_ids: list[str], target_file_ext: list[str], result_dir='ncbi_genomes') -> list[str]:
+def _form_gtdb_taxonomy_file_url(release_ver):
+    # form GTDB taxonomy URL for specific GTDB version.
+    # e.g. https://data.gtdb.ecogenomic.org/releases/release207/207.0/ar53_taxonomy_r207.tsv
+
+    gtdb_server = 'https://data.gtdb.ecogenomic.org/releases/'
+    file_dir_url = gtdb_server + f'release{release_ver}/' + f'{release_ver}.0/'
+
+    file_urls = list()
+    if release_ver in ['207', '202', '95', '89']:
```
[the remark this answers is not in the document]
MrCreosote: Hmm, ok. Depending on the page source is fragile at best so it seems we have two not-great choices (although the source is pretty simple). Depending on the page source means possibly less trouble in the future since it should work for new releases, but if the page source changes things might break (but that doesn't seem super likely). Doing it the way we're doing it now means less trouble now but a guarantee things'll break the next time a version is released. 

I'll leave it up to you, have fun

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍  parsing GTDB web page for releasing versions now.

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,91 @@
 """
-Download genome files from NCBI FTP server
-"""
+PROTOTYPE - Download genome files from NCBI FTP server.
+
+usage: ncbi_downloader.py [-h] --download_file_ext
+                          DOWNLOAD_FILE_EXT
+                          [DOWNLOAD_FILE_EXT ...]
+                          --release_ver
+                          {202,207,80,83,86,89,95}
```
MrCreosote, line 8: This is going to change over time, should probably be removed unless I'm missing something

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,91 @@
...
+                          [--threads THREADS] [--overwrite]
+
+options:
+  -h, --help            show this help message and exit
+
+required named arguments:
+  --download_file_ext DOWNLOAD_FILE_EXT [DOWNLOAD_FILE_EXT ...]
+                        Download only files that match given
+                        extensions.
+  --release_ver {202,207,80,83,86,89,95}
```
MrCreosote, line 20: same re allowed versions

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,91 @@
...
+    links = soup.find_all('a')
 
-def _download_genome_file(download_dir: str, gene_id: str, target_file_ext: list[str]) -> None:
+    # release links follow format like 'latest/', 'release202/', 'release83/', etc.
+    release_ver = [link.get('href').split('release')[-1][:-1] for link in links if 'release' in link.get('href')]
+
+    return release_ver
+
+
+GTDB_RELEASE_VER = _parse_gtdb_release_vers()  # available GTDB release versions
```
MrCreosote, line 85: This will trigger an internet call every time this file is imported, pretty please don't do that. Move the call to `main()` at least and pass around the list as necessary

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -34,18 +110,92 @@ def _download_genome_file(download_dir: str, gene_id: str, target_file_ext: list
...
 
-def download_genome_files(gene_ids: list[str], target_file_ext: list[str], result_dir='ncbi_genomes') -> list[str]:
+def _form_gtdb_taxonomy_file_url(release_ver):
+    # form GTDB taxonomy URL for specific GTDB version.
+    # e.g. https://data.gtdb.ecogenomic.org/releases/release207/207.0/ar53_taxonomy_r207.tsv
+
+    if release_ver not in GTDB_RELEASE_VER:
+        raise ValueError(f'Unsupported GTDB release version: {release_ver}')
+
+    file_dir_url = GTDB_DOMAIN + f'release{release_ver}/' + f'{release_ver}.0/'
```
MrCreosote, line 132: Hmm, we really should support dot releases... but the only release with dot releases is 86, and the chance of use using that is very low. How much work would it be to support them?

MrCreosote: another thing that could be done in a different PR maybe

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -34,18 +110,92 @@ def _download_genome_file(download_dir: str, gene_id: str, target_file_ext: list
...
+    # download GTDB taxonomy files and then parse genome_ids from the GTDB taxonomy files
+    genome_ids = list()
+
+    taxonomy_urls = _form_gtdb_taxonomy_file_url(release_ver)
+
+    for taxonomy_url in taxonomy_urls:
+        # download GTDB taxonomy file to work_dir
+        url = urlparse(taxonomy_url)
+        taxonomy_file = os.path.join(work_dir, os.path.basename(url.path))
+        _ = request.urlretrieve(taxonomy_url, taxonomy_file)
```
MrCreosote, line 164: Is there a reason to have the `=` at all?

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -79,3 +229,64 @@ def download_genome_files(gene_ids: list[str], target_file_ext: list[str], resul
...
+
+    work_dir = _make_work_dir(root_dir, source, release_ver)
+
+    genome_ids = _fetch_genome_ids(source, release_ver, work_dir)
+
+    if not threads:
+        threads = max(int(multiprocessing.cpu_count() * min(SYSTEM_UTILIZATION, 1)), 1)
+    print(f"Start downloading genome files with {threads} threads")
+
+    chunk_size = max(len(genome_ids) // (threads - 1), 1)  # distribute genome ids evenly across threads
```
MrCreosote, line 278: div0 if threads = 1

MrCreosote (changes requested):

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,15 +1,75 @@
...
+from urllib.parse import urlparse
 
 import ftputil
 
+# Fraction amount of system cores can be utilized
+# (i.e. 0.5 - program will use 50% of total processors,
+#       0.1 - program will use 10% of total processors)
+SYSTEM_UTILIZATION = 0.5
+
+GTDB_RELEASE_VER = ['207', '202', '95', '89', '86', '83', '80']  # available GTDB release versions
```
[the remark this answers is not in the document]
MrCreosote: fixed :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: 👍

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: 👍

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: 👍

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: added `todo` comments. I think we will probably never use 86.

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: nope. removed.

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: 👍  good catch. updated the logic.

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -79,3 +222,70 @@ def download_genome_files(gene_ids: list[str], target_file_ext: list[str], resul
...
+     source,
+     threads,
+     overwrite) = (args.download_file_ext, args.release_ver, args.root_dir, args.source,
+                   args.threads, args.overwrite)
+
+    if source not in SOURCE:
+        raise ValueError(f'Unexpected source. Currently supported sources: {SOURCE}')
+
+    if release_ver not in gtdb_release_vers:
+        raise ValueError(f'Unsupported GTDB release version: {release_ver}')
```
MrCreosote, line 266: This is unreachable isn't it? Argparse will catch bad release vers

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
```diff
@@ -1,5 +1,103 @@
...
+@pytest.fixture(scope="module")
+def setup_and_teardown():
+    print('starting NCBI downloader tests')
+    tmp_dir = 'result_{}'.format(str(uuid.uuid4()))
+    os.makedirs(tmp_dir, exist_ok=True)
+
+    caller_filename_full = Path(__file__).resolve()
+    caller_file_dir = os.path.dirname(caller_filename_full)
+
+    project_dir = Path(caller_file_dir).resolve().parents[3]
```
MrCreosote, line 21: Ok, this is the 3rd time I think I've seen this pattern, can you do a separate PR to make a test helper method in `conftest.py`

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
```diff
@@ -1,5 +1,103 @@
...
+
+    shutil.rmtree(tmp_dir, ignore_errors=True)
+
+
+def test_make_work_dir(setup_and_teardown):
+    tmp_dir, script_file = setup_and_teardown
+
+    with pytest.raises(ValueError, match='Unexpected source:'):
+        fake_source = 'hello_fake'
+        _ = ncbi_downloader._make_work_dir(tmp_dir, fake_source, 'release_ver')
```
MrCreosote, line 34: why `=`

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
```diff
@@ -1,5 +1,103 @@
+import os
```
MrCreosote, line 1: The tests present look good, but given tests aren't required in prototype mode, I'm not commenting on uncovered code or missing tests

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: yup. removed.

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
Tianhao-Gu: removed. I did it sometimes just to remind me the method actually returns something.

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
Tianhao-Gu: 👍 yup. there are uncovered codes. I mainly do this to help me check some methods are working and returning properly.

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

MrCreosote (approved):
LGTM

PR #89, opened 2022-12-14
RE2022-64: Matching proposal rev -99

MrCreosote:


design/matching_lineage.md
```diff
@@ -0,0 +1,114 @@
...
+        via an API. TBD
+      * Matches must expire based on the *last usage* - e.g. if a match is retrieved, its lifetime
+        must be renewed.
+        * Perhaps up to some maximum lifetime, to avoid crafy people setting up cron jobs to
+          touch their matches.
+      * Will need some means of cleaning up the match and any supporting DP data produced (see
+        below), including ArangoDB records, cache service data, etc.
+  * Data visualization
+    * When accessing a DP (like `taxa_counts`) the user would provide the match ID
+    * The DP is responsible for any processing that is required to support the visualization,
```
scanon, line 95: I think the wording here is a bit confusing.  The data product processing can and should do any pre-processing that can help to accelerate these queries, but the collections service will still be responsible for how to use any of that including what to cache.

scanon (dismissed):
This seems good.  I had a few comments, but we can continue to refine and clarify things.

I feel like some of the abbreviations get in the way.  I had to keep looking back to see what they meant.

Also we may want to add "match" and "connectors" to the Nomenclature.

design/matching_lineage.md
```diff
@@ -0,0 +1,114 @@
...
+    * Matches are not expected to be searchable or otherwise accessible unless the user has
+      the match ID.
+  * Exact matches:
+    * The CS runs an `IN` AQL query on the collection attributes AC to find matching members of
+      the collection
+      * May need to be split into multiple queries if there are too many inputs
+    * This may return tens of thousands of matches per input G/A
+  * Rank matches:
+    * The user must also provide a rank level (e.g. phylum, domain, etc)
+    * The CS truncates the input matches' lineages to the provided rank
```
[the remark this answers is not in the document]
MrCreosote: How do you figure? It'd either be an equality search on a 7 member compound index or a range search on a single index, unless I'm not understanding your strategy. I don't think it'd make a huge difference, and the number of returned results should be the same

design/matching_lineage.md
MrCreosote: yeah, the DP I'm referring to here is a subsystem of the CS. I'll clarify that.

design/matching_lineage.md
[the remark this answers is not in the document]
MrCreosote: Well, a 7 member compound index might be overkill, there may be ways to reduce that depending on uniqueness of names in the tree. Still though, I don't think a range search on the full lineage is going to be much different

design/matching_lineage.md
MrCreosote: Fixed. Also changed "connectors" to "matchers" as that seems to explain what they do better

scanon (approved):

dauglyon (approved):

PR #93, opened 2022-12-22
RE2022-79: Add workspace to config / startup

MrCreosote:
Will be needed for matchers.

Tested startup code manually with good & bad workspace urls. The get_workspace method is not yet tested; it'll get tested when it's used.

The http_bearer changes were tested with by printing the user in one of the endpoints.


docker-compose.yaml
```diff
@@ -18,9 +18,10 @@ services:
       - KBCOLL_ARANGO_DB=collections_test
       - KBCOLL_ARANGO_USER=root
       - KBCOLL_ARANGO_PWD=foobar
-      - KBASE_AUTH_URL=https://ci.kbase.us/services/auth
+      - KBCOLL_KBASE_AUTH2_URL=https://ci.kbase.us/services/auth
```
MrCreosote, line 21: oops

src/service/app_state.py
```diff
@@ -58,6 +62,18 @@ async def _build_storage(cfg: CollectionsServiceConfig, data_products: list[Data
         raise
 
 
+async def _get_workspace_url(cfg: CollectionsServiceConfig) -> str:
+    if cfg.dont_connect_to_external_services:
+        return False
+    ws = Workspace(cfg.workspace_url)
```
Tianhao-Gu, line 68: OK. But just in case, I would move `ws = Workspace(cfg.workspace_url)` to the try-except clause below.

Tianhao-Gu (dismissed):
Overall LGTM. Just have one comment.

src/service/app_state.py
MrCreosote: The only error that could occur here is a url parse error, for which the catch / rethrow isn't really necessary as the error message should provide enough info, but it doesn't hurt either, so fair enough

MrCreosote:
The last commit was a trivial change requested by the reviewer and the rest of the code is approved, so I'm going to go ahead and merge. If changes are necessary we can do another PR

PR #98, opened 2023-01-05
RE2022-78: Add matchers endpoint

MrCreosote:
Returns information about matchers avialable in the service. Useful for people adding matchers to collections

### DIFF
--- a/src/service/matchers/common_models.py
+++ b/src/service/matchers/common_models.py
@@ -1,0 +1,45 @@
+"""
+Data structures common to all matchers
+"""
+
+from pydantic import BaseModel, Field, validator
+from typing import Any
+
+
+class Matcher(BaseModel):
+    id: str = Field(
+        example="gtdb_lineage",
+        description="Matches assemblies and genomes via the GTDB lineage.",
+        regex="^[a-z_]+$",
+    )
+    types: list[str] = Field(
+        # TODO MATCHERS validate types against workspace on startup
+        # It'd be nice to specify type versions as well, but KBase has essentially jettisoned
+        # that idea
+        example=["KBaseGenomes.Genome", "KBaseGenomeAnnotations.Assembly"],
+        description="The KBase types against which the matcher operates.",
+    )
+    description: str = Field(
+        example="Matches assemblies and genomes via the GTDB lineage",
+        description="A free text description of the matcher.",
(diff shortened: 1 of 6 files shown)


src/service/matchers/common_models.py
```diff
@@ -0,0 +1,45 @@
...
+        # It'd be nice to specify type versions as well, but KBase has essentially jettisoned
+        # that idea
+        example=["KBaseGenomes.Genome", "KBaseGenomeAnnotations.Assembly"],
+        description="The KBase types against which the matcher operates.",
+    )
+    description: str = Field(
+        example="Matches assemblies and genomes via the GTDB lineage",
+        description="A free text description of the matcher.",
+    )
+    user_parameters: dict[str, Any] | None = Field(
```
dauglyon, line 26: Having this be any json-schema basically rules out any sort of automatic UI for user params, because of complexity. I'm sort of against an automatic UI for that anyway (it'd basically be duplicating behavior from apps) so I'm not opposed but I thought I should point it out

dauglyon (approved):
This looks good to me, one comment about user params being jsonschema

src/service/matchers/common_models.py
MrCreosote: I don't see any need for an automated UI builder given how many matchers we're probably going to have. That being said I don't really understand why JSONSchema rules that out? What would be better?

src/service/matchers/common_models.py
```diff
@@ -0,0 +1,45 @@
...
+
+from pydantic import BaseModel, Field, validator
+from typing import Any
+
+
+class Matcher(BaseModel):
+    id: str = Field(
+        example="gtdb_lineage",
+        description="Matches assemblies and genomes via the GTDB lineage.",
+        regex="^[a-z_]+$",
```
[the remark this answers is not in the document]
MrCreosote: Nah, I doubt we'll have more than a dozen or two matchers at most, so no need for a large ID space

src/service/matchers/common_models.py
dauglyon: It would rule out generating UI _from_ the jsonschema, as it's super expressive, so would require eons of work to cover all the possible cases. Doesn't rule out some other strategy

PR #100, opened 2023-01-07
Generate a metadata file for genome identifiers

Tianhao-Gu:
produce a metadata file for genome identifiers created by tools and mock some tests to improve performance.

### DIFF
--- a/src/loaders/genome_attributes/compute_genome_attribs.py
+++ b/src/loaders/genome_attributes/compute_genome_attribs.py
@@ -207,17 +207,23 @@ def _map_tool_id_to_genome_id(tool_name, original_genome_id, genome_file_name):
         raise ValueError(f'the method for tool {tool_name} has not been implemented.')
 
 
-def _create_tool_genome_id_metadata_file(tool_genome_id_map, genome_count, batch_dir):
+def _create_tool_genome_id_metadata_file(tool_genome_id_map, source_genome_file_map, genome_count, batch_dir):
     # create tab separated metadata file with tool generated genome identifier and original genome id.
 
     if len(tool_genome_id_map) != genome_count:
         raise ValueError('Some genomes are absent from the genome metadata file')
 
-    meta_file_path = os.path.join(batch_dir, loader_common_names.META_FILE_NAME)
-
-    with open(meta_file_path, "w") as meta_file:
+    # create tool genome identifier metadata file
+    tool_genome_meta_file_path = os.path.join(batch_dir, loader_common_names.TOOL_GENOME_META_FILE_NAME)
+    with open(tool_genome_meta_file_path, "w") as meta_file:
         for tool_genome_identifier, genome_id in tool_genome_id_map.items():
             meta_file.write(f'{tool_genome_identifier}\t{genome_id}\n')
+
+    # create source genome file metadata file
+    source_genome_meta_file_path = os.path.join(batch_dir, loader_common_names.SOURCE_GENOME_META_FILE_NAME)
+    with open(source_genome_meta_file_path, "w") as meta_file:
+        for genome_id, source_genome_file in source_genome_file_map.items():
+            meta_file.write(f'{genome_id}\t{source_genome_file}\n')
 
 
 def gtdb_tk(genome_ids, work_dir, source_data_dir, debug, program_threads, batch_number,
@@ -238,7 +244,7 @@ def gtdb_tk(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
     # tab separated in 2 columns (FASTA file, genome ID)
     batch_file_name = f'genome.fasta.list'
     batch_file_path = os.path.join(batch_dir, batch_file_name)
-    tool_genome_id_map = dict()
+    tool_genome_id_map, source_genome_file_map = dict(), dict()
     with open(batch_file_path, "w") as batch_file:
         for genome_id in genome_ids:
             genome_file = _find_genome_file(genome_id, 'genomic.fna.gz', source_data_dir,
@@ -249,10 +255,11 @@ def gtdb_tk(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
                 tool_genome_id_map.update(_map_tool_id_to_genome_id('gtdb_tk',
                                                                     genome_id,
                                                                     os.path.basename(genome_file[0])))
+                source_genome_file_map[genome_id] = genome_file[0]
                 # According to GTDB, the batch file should be a two column file indicating the location of each genome
                 # and the desired genome identifier
                 batch_file.write(f'{genome_file[0]}\t{genome_id}\n')
-    _create_tool_genome_id_metadata_file(tool_genome_id_map, len(genome_ids), batch_dir)
+    _create_tool_genome_id_metadata_file(tool_genome_id_map, source_genome_file_map, len(genome_ids), batch_dir)
     if run_steps:
         _run_gtdb_tk_steps(batch_file_path, batch_dir, debug, genome_ids, program_threads)
     else:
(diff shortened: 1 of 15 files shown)


test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
```diff
@@ -43,16 +43,15 @@ def test_make_work_dir(setup_and_teardown):
...
 
 
-def test_fetch_gtdb_genome_ids(setup_and_teardown):
+def test_fetch_gtdb_genome_ids(setup_and_teardown, mocker):
     tmp_dir, script_file = setup_and_teardown
 
-    release_ver = '207'
-    genome_ids = ncbi_downloader._fetch_gtdb_genome_ids(release_ver, tmp_dir)
-    assert len(genome_ids) == 317542
+    mock_form_url = mocker.patch('src.loaders.ncbi_downloader.ncbi_downloader._form_gtdb_taxonomy_file_url')
```
MrCreosote, line 49: I don't really understand why this mock is necessary...?

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
```diff
@@ -43,16 +43,15 @@ def test_make_work_dir(setup_and_teardown):
     assert path.parents[0].name == source
 
 
-def test_fetch_gtdb_genome_ids(setup_and_teardown):
+def test_fetch_gtdb_genome_ids(setup_and_teardown, mocker):
     tmp_dir, script_file = setup_and_teardown
 
-    release_ver = '207'
```
MrCreosote, line 49: If we're going to only download from one release why 202 vs 207?

src/loaders/genome_attributes/compute_genome_attribs.py
```diff
@@ -349,7 +393,7 @@ def main():
         # place computed results to COLLECTION_DATA_DIR directory
         work_dir = os.path.join(root_dir, loader_common_names.COLLECTION_DATA_DIR, kbase_collection,
                                 load_ver, tool)
-        # TODO writing a file with genome ID -> file mappings in work_dir
```
MrCreosote, line 352: Huh, I'd forgotten we had a TODO for this

src/loaders/genome_attributes/compute_genome_attribs.py
```diff
@@ -185,6 +185,41 @@ def _run_gtdb_tk_classify_wf(batch_file_path, work_dir, debug, genome_ids, progr
         f'{len(genome_ids)} genomes')
 
 
+def _map_tool_id_to_genome_id(tool_name, original_genome_id, genome_file_name):
```
MrCreosote, line 188: I thought we were going to save the path to the input in this file.

src/loaders/genome_attributes/compute_genome_attribs.py
```diff
@@ -185,6 +185,41 @@ def _run_gtdb_tk_classify_wf(batch_file_path, work_dir, debug, genome_ids, progr
...
+        # CheckM2 uses the base name (without extension) of genome_file as the genome identifier
+        # We know for checkM2 extension 'genomic.fna.gz' is consistently used for genome files
+
+        base_name = genome_file_name.split('.fna.gz')[0]
+        return {base_name: original_genome_id}
+
+    elif tool_name == 'gtdb_tk':
+        # When creating the batch file for GTDB-TK, we use the genome ID as the identifier for each genome.
+        return {original_genome_id: original_genome_id}
+    else:
```
MrCreosote, line 206: Maybe not yet, but we might want to think about creating an interface that defines all the tool-specific methods like these and the run methods and implementing a class for each tool so that all the code for each tool is in one place and it's clear where code needs to be added to add a new tool

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
Tianhao-Gu: with this mock, it only downloads the much smaller `ar` file rather than downloading both `bac` and `ar`. I was hoping this could help to speed up the test overall.

src/loaders/genome_attributes/compute_genome_attribs.py
Tianhao-Gu: 👍 added todo item.

src/loaders/genome_attributes/compute_genome_attribs.py
Tianhao-Gu: yea. We were talking about doing `genome_id`: `source_genome_file`. But when I implemented it, I thought again it made no sense to do that because `genome_id`: `source_genome_file` information is already embedded into the file structure. But more importantly, the parser cannot use `genome_id`: `source_genome_file` information directly to identify the genome identifier for each tool. The parser will again parse the `source_genome_file` based on different tools, as it is currently doing.

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
Tianhao-Gu: hmmm I didn't get what you mean. I removed `207` hoping to speed up the test. Are you suggesting we should add it back?

MrCreosote: I was wondering why you picked to test 202 rather than testing 207.

That said, maybe we should just leave the tests as is. They only take 30s on GHA, and the last couple tests I've run locally have been <90s

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
MrCreosote: I see, ok

src/loaders/genome_attributes/compute_genome_attribs.py
MrCreosote: :+1:

src/loaders/genome_attributes/compute_genome_attribs.py
MrCreosote: is there a reason not to have both the file and the genome id?

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
Tianhao-Gu: 👍  revert changes to tests. It takes about 30s run locally.

src/loaders/genome_attributes/compute_genome_attribs.py
Tianhao-Gu: `both the file and genome id`? The metadata file created has `tool_genome_identifer: genome_id`. The source file and genome id mapping are embedded already into the file structure. The folder name is the `genome_id` and the files inside are the source files downloaded from NCBI. Does that answer your question? I might misunderstand your question.

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
MrCreosote: :+1:

src/loaders/genome_attributes/compute_genome_attribs.py
MrCreosote: I understand you're saying it's redundant - what I'm asking is if there's a reason not to put this redundant information into the csv file just to make it easy for someone reading the file to figure out where the source file is

MrCreosote: Why add another file rather than another column in the current TSV file?

Tianhao-Gu: 👍 added source genome file column.

MrCreosote: :+1:

MrCreosote (approved):

PR #103, opened 2023-01-12
RE2022-79: Add matchers endpoint for collections

MrCreosote:


src/service/data_product_specs.py
```diff
@@ -9,6 +9,9 @@
 Note that "builtin" is a reserved ID for data products.
 """
 
+# NOTE: Once a collection has been saved with a data product, the data product cannot be
```
dauglyon, line 12: This seems concerning....

dauglyon: Any way to prevent this?

dauglyon (approved):
Approved with one noted concern

src/service/data_product_specs.py
MrCreosote: I don't think so. If the collection says it included data product A but the code to support data product A has been removed from the service, unexpected things are going to happen.

What would have to happen is that we'd need an updater script to go through the collections and remove the data product in question (and any matchers that might depend on the data product)

PR #109, opened 2023-01-18
RE2022-83: save match to arango

MrCreosote:


MrCreosote:
@dauglyon feel free to remove if you want - this doesn't touch the API though so I figured you wouldn't care. Added you accidentally

src/service/storage_arango.py
```diff
@@ -327,3 +330,45 @@ async def get_collection_versions(
         cur = await self._db.aql.execute(aql, bind_vars=bind_vars)
         return [_doc_to_saved_coll(d) async for d in cur]
 
+    async def save_match(self, match: models.InternalMatch) -> models.Match:
```
Tianhao-Gu, line 333: should we provide an option to overwrite the existing match?

Tianhao-Gu (approved):
LGTM

src/service/storage_arango.py
MrCreosote: It'd essentially be a noop since the match ID is the MD5 of the contents of the match. Even if that weren't the case, we don't want to overwrite because matches need to be deleted very carefully since there will be data associated with the match in other parts of the DB

PR #112, opened 2023-01-21
RE2022-83: Add user permissions check to model

MrCreosote:
Will be needed for the workspace permissions check coming up next.


Tianhao-Gu (approved):
lgtm

PR #116, opened 2023-01-26
RE2022-83: Run match process in separate process

MrCreosote:
Pydantic models don't pickle in some cases - it appears to have something to do with the validation function. Required some refactoring.

### DIFF
--- a/src/service/match_processing.py
+++ b/src/service/match_processing.py
@@ -1,0 +1,33 @@
+"""
+Common code for processing matches.
+"""
+
+import multiprocessing
+
+from pydantic import BaseModel, Field
+from typing import Callable, Any
+from src.service.app_state import PickleableStorage
+from src.service.storage_arango import ArangoStorage
+
+
+class MatchProcess(BaseModel):
+    """
+    Defines a match process.
+    """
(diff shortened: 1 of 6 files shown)


MrCreosote:
Tests are failing due to codecov upload

src/service/match_processing.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def start(self, match_id: str, storage: PickleableStorage):
+        """
+        Start the match process in a forkserver.
+
+        match_id - the ID of the match.
+        storage - the storage system containing the match and the data to match against.
+        """
+        ctx = multiprocessing.get_context("forkserver")
+        ctx.Process(target=self.process, args=(match_id, storage, self.args)).start()
```
Tianhao-Gu, line 33: I started to get lost here. We are creating a new process for every match? What's the reason for doing that?

MrCreosote: Because matches can take a long time. For now I'm kicking off a process, but in the future we may need to kick off an ee2 job or something else

Tianhao-Gu (approved):
👍

PR #117, opened 2023-01-27
RE2022-83: Add get_match method to storage

MrCreosote:
Targeted toward the API layer which shouldn't get internal match data


Tianhao-Gu (approved):
👍 LGTM

PR #119, opened 2023-01-30
refactor taxa count script

Tianhao-Gu:


.gitignore
```diff
@@ -12,3 +12,4 @@ gtdb_attribs.jsonl
 .dockerignore
 *_pycache__
 gtdb_attribs2.jsonl
+.DS_Store
```
MrCreosote, line 15: I don't really understand the reason for moving the files below from `genome_attribs` to `genome_collection`. The former seems fine to me - what's the rationale?

src/loaders/genome_collection/compute_genome_taxa_count.py
```diff
@@ -0,0 +1,173 @@
...
+required named arguments:
+  load_files            Files containing genome taxonomy (e.g. ar53_taxonomy_r207.tsv, computed_genome_attribs.json)
+  --load_ver LOAD_VER   KBase load version (e.g. r207.kbase.1).
+
+optional arguments:
+  --kbase_collection KBASE_COLLECTION
+                        KBase collection identifier name (default: GTDB).
+  -o OUTPUT, --output OUTPUT
+                        Output JSON file path.
+  --input_source {GTDB,genome_attributes}
```
MrCreosote, line 35: Is there a reason to continue supporting GTDB direct input? You should be able to get the same results by running the script that takes the metadata file and produces what would normally be produced from the calculation parsing script, and then running this script on that, right?

src/loaders/genome_collection/compute_genome_taxa_count.py
```diff
@@ -0,0 +1,173 @@
...
+# Collection and load version information will be prepended to this file name.
+GENOME_TAXA_COUNT_FILE = "parsed_genome_taxa_counts.json"
+
+# The source of the input file containing genome taxonomy information to be parsed
+# GTDB - taxonomy file downloaded directly from the GTDB website, such as 'bac120_taxonomy_r207.tsv'
+# genome_attributes - genome attributes file created by running 'parse_computed_genome_attribs.py' script.
+VALID_SOURCE = ['GTDB', 'genome_attributes']
+
+# taxonomy attribute name from computed genome attributes
+TAXA_ATTRI_NAME = 'classification'
```
MrCreosote, line 53: Hold on - this should be `gtdb_taxonomy`. This makes me think we may have miscommunicated about what exactly this code is supposed to take as input, so I'm going to hold off on reviewing the rest of this script until we sort that out.

My understanding is that it should be working on the JSONL files produced by the parsing script, but I don't think that's what's happening here

.gitignore
Tianhao-Gu: I was thinking taxa_count is not exactly genome_attributes.

MrCreosote: Ok, I guess we could split that into 2 directories, but this is fine for now, at least.
:+1:

src/loaders/genome_collection/compute_genome_taxa_count.py
MrCreosote: It turns out that the code is reading from the right file but the gtdb lineage key is different, which is a problem, but we'll deal with that later. https://kbase-jira.atlassian.net/browse/RE2022-99

MrCreosote: Continuing on with review below...

MrCreosote: never mind, looks fine otherwise. It's mostly identical to the old script other than input.

However - the difference between the keys between the output of the GTDB metadata parser and the compute -> parser scripts means this can't run on the output of the GTDB metadata parser. If the keys are the same, than we can get rid of the direct support for the GTDB metadata file and just run it on the parser output

src/loaders/genome_collection/compute_genome_taxa_count.py
MrCreosote: See comment chain below - we can't do this until the GTDB metadata parser and the compute -> parse scripts use the same key for gtdb lineage. Made a ticket for this.

MrCreosote (approved):
LGTM with the caveats re key name changes in the comments above, which we'll figure out after this PR

PR #121, opened 2023-01-30
RE2022-83: Add set match state to arango storage

MrCreosote:
Also some minor prep for matching and refactoring of pickable dependencies


Tianhao-Gu (approved):
👍

PR #124, opened 2023-01-31
unify name for taxonomy field

Tianhao-Gu:


MrCreosote (commented):
My feeling is that we should do the mapping the other way around, and have the field produced by GTDB_tk be the canonical key rather than the key in the meta data file

Tianhao-Gu:
> My feeling is that we should do the mapping the other way around, and have the field produced by GTDB_tk be the canonical key rather than the key in the meta data file

👍

MrCreosote (approved):
LGTM

PR #138, opened 2023-02-06
RE2022-98: Add data product match structure & arango methods

MrCreosote:
Also make internal_match_id a standard field for taxa counts so the matches can be stored in the same collection

Also also add a bulk update helper method which will be needed when inserting taxa_count records


Tianhao-Gu (approved):
💯

PR #142, opened 2023-02-07
Small bugfixes & simplifications

MrCreosote:
1. Make the MatchProcess class handle the asyncio stuff rather than making the implementer of the matcher do it. Now they just have to implement an async function with the right args.
2. Simplify the API for running through a genome attributes match slightly.
3. Fix a potential, although unlikely, race condition when matching
4. Fix a bug when a `None` or empty fields list is supplied to `process_match_documents`


Tianhao-Gu (approved):
👍

PR #144, opened 2023-02-09
Refactor processing code for reuse

MrCreosote:
match specific to non-specific

### DIFF
--- a/src/service/match_retrieval.py
+++ b/src/service/match_retrieval.py
@@ -1,0 +1,127 @@
+"""
+Methods for retriving matches from a storage system, ensuring that the user has permissions
+to the match, the match is in the expected state, and access times are updated correctly.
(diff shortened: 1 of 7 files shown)


src/service/match_retrieval.py
```diff
@@ -0,0 +1,127 @@
...
+
+from src.service import errors
+from src.service import models
+from src.service.models import remove_non_model_fields
+from src.service.storage_arango import ArangoStorage
+from src.service.timestamp import now_epoch_millis
+from src.service.workspace_wrapper import WorkspaceWrapper
+
+
+# might want to make this configurable
```
MrCreosote, line 14: C&P from `processing.py` from here down

Tianhao-Gu (approved):
LGTM

PR #146, opened 2023-02-09
RE2022-89: Add a heartbeat to match processes

MrCreosote:
Tested by adding an asyncio.sleep(100) statement to the match process and watching the match record in arango

Next:
* When a standard match or data product match is pulled from the DB, check the last heart beat and restart the process if it's beyond the threshold

### DIFF
--- a/src/service/processing.py
+++ b/src/service/processing.py
@@ -4,11 +4,18 @@ both performing the initial match and calculating secondary data products.
 """
 
 import asyncio
+from apscheduler.schedulers.asyncio import AsyncIOScheduler
+from apscheduler.triggers.interval import IntervalTrigger
 import multiprocessing
 
 from pydantic import BaseModel, Field
 from typing import Callable, Any
 from src.service.app_state import PickleableDependencies
+from src.service.timestamp import now_epoch_millis
+
+
+HEARTBEAT_INTERVAL_SEC = 10  # make configurable?
+HEARTBEAT_RESTART_THRESHOLD_SEC = 60  # 1 minute without a heartbeat
 
 
 class CollectionProcess(BaseModel):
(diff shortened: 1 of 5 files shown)


src/service/processing.py
```diff
@@ -45,3 +52,48 @@ def _run_process(
     args: list[Any]
 ):
     asyncio.run(function(data_id, pstorage, args))
+
+
+class Heartbeat:
```
Tianhao-Gu, line 57: Do you think in the future maybe we need something like a `status` method to check the current status of the heartbeat, whether it's running or stopped

src/service/processing.py
```diff
@@ -45,3 +52,48 @@ def _run_process(
...
+        self._schd.start()
+
+    async def _heartbeat(self):
+        await self._hbf(now_epoch_millis())
+
+    def stop(self):
+        """
+        Stop the heartbeat.
+        """
+        if self._job_id:
```
Tianhao-Gu, line 96: Do we need to handle the case where the _job_id is None or the job has already been removed?

Tianhao-Gu (approved):
LGTM.

src/service/processing.py
MrCreosote: I'm not sure I understand - this does handle that case by not doing anything. I don't think there's anything to do - what were you thinking?

src/service/processing.py
MrCreosote: Based on the current usage there's no need for it, but it'd be easy to add if a use case turns up. Right now it's just start it, forget it, and then stop it when your process is complete

Tianhao-Gu: 👍

src/service/processing.py
Tianhao-Gu: yea. I agree. I was just thinking something like a print statement in an `else` clause. It probably won't make any difference I agree.

PR #150, opened 2023-02-17
Refactor dependencies into class

MrCreosote:
While working on the dead process restart feature, I ran into a couple problems:

* `matcher_registry` imports were winding up everywhere in order to get matchers to rerun match processes. This lead to circular imports all over as well.
* I was having to pass in storage, workspace, pickleable deps, and matcher dependencies to some of the new methods, which was making the method signatures huge.

To fix this I refactored dependency handling into a state class that can be passed around and used to get dependencies as needed. I added the matchers to that class so that the matcher registry only needs to be imported at the top level of the app and the matchers passed into the state builder, which breaks the import loops.

### DIFF
--- a/collections_config.toml.jinja
+++ b/collections_config.toml.jinja
@@ -33,16 +33,6 @@ root_path = "{{ KBCOLL_SERVICE_ROOT_PATH or "" }}"
 # their sharding preferences.
 create_db_on_startup = "{{ KBCOLL_CREATE_DB_ON_STARTUP or "false" }}"
 
-# Set to "true" to not connect to any external services except for authorization,
-# including the database. This will cause all calls to the service that require external
-# service access to fail but is useful for quickly checking OpenAPI documentation
-# or general methods that don't access external services.
-# The service can be run via:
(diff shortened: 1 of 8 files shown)


collections_config.toml.jinja
```diff
@@ -33,16 +33,6 @@ root_path = "{{ KBCOLL_SERVICE_ROOT_PATH or "" }}"
...
 
-# Set to "true" to not connect to any external services except for authorization,
-# including the database. This will cause all calls to the service that require external
-# service access to fail but is useful for quickly checking OpenAPI documentation
-# or general methods that don't access external services.
-# The service can be run via:
-# docker run --env-file <your .env file> --publish 5000:5000 <image hash>
-# Example:
-# docker run --env-file collections.env --publish 5000:5000 3b6523ebf6b8
-dont_connect_to_external_services = "{{ KBCOLL_DONT_CONNECT_TO_EXT_SERVICES or "false" }}"
```
MrCreosote, line 44: this was useful at the beginning of development but now I never use and docker-compose works better anyway.

Tianhao-Gu (approved):
LGTM

PR #153, opened 2023-02-25
Pre match deletion refactoring

MrCreosote:
A little bit of refactoring for testability and code reuse in match deletion.

* Refactoring the processing code slightly to reuse in deletions.
* Made timestamp generation more easily mockable.
  * Also laid out rules for displaying timestamps to API clients.


Tianhao-Gu (approved):
💯

PR #155, opened 2023-02-27
RE2022-85: Move matches to a deleted state

MrCreosote:
Next: actually delete all the match data

```python
In [1]: from src.service.match_deletion import MatchCleanup

In [2]: from src.service import app_state

In [3]: from src.service.config import CollectionsServiceConfig

In [4]: from src.service.data_product_specs import DATA_PRODUCTS

In [5]: with open('./collections_config.toml', 'rb') as f:
   ...:     cfg = CollectionsServiceConfig(f)
   ...:

In [6]: dps = {dp.data_product: dp.db_collections for dp in DATA_PRODUCTS.values
   ...: ()}

In [7]: def mc():
   ...:     pd = app_state.PickleableDependencies(cfg, dps)
   ...:     mc = MatchCleanup(pd, 10, 5, 24 * 60 * 60 * 1000)
   ...:     mc.start()
   ...:     return mc
   ...:

In [8]: # taken from a match in the DB:

In [9]: 1677458286548 - (24 * 60 * 60 * 1000)
Out[9]: 1677371886548

In [10]: # save this value back to the DB in last_access

In [11]: matchc = mc()

In [12]: INFO:src.service.match_deletion:Moving match ff1d36301008ace844e77b8a7698273c to deleted state
In [12]:

In [12]: matchc.stop()
```

### DIFF
--- a/src/service/match_deletion.py
+++ b/src/service/match_deletion.py
@@ -2,23 +2,43 @@
 Routines to safely delete matches and clean up match data in the collections system
 """
 
+import logging
+
 from apscheduler.schedulers.background import BackgroundScheduler
 
-from datetime import datetime
-
 from src.service import app_state
+from src.service import models
 from src.service import processing
+from src.service.storage_arango import ArangoStorage
+
+async def _move_match_to_deletion(
+        deps: app_state.PickleableDependencies,
+        storage: ArangoStorage,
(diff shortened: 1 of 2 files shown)


Tianhao-Gu (approved):
LGTM

PR #165, opened 2023-03-03
Refactor to break import loops

MrCreosote:
* Made the data product registry load data products on request rather than at import time to break import cycles. Now we can reference the data product registry in various places without import loops.
* Split up app_state to allow importing subparts of the state and avoiding import loops.

### DIFF
--- a/src/service/_app_state_build_storage.py
+++ b/src/service/_app_state_build_storage.py
@@ -1,0 +1,55 @@
+import aioarango
+import asyncio
+import sys
+
+from src.service.config import CollectionsServiceConfig
+from src.service.data_products.common_models import DBCollection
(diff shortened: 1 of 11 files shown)


src/service/_app_state_build_storage.py
```diff
@@ -0,0 +1,55 @@
+import aioarango
```
MrCreosote, line 1: C&P from `app_state`

src/service/app_state_data_structures.py
```diff
@@ -0,0 +1,125 @@
+"""
```
MrCreosote, line 1: C&P from `app_state`

Tianhao-Gu (approved):
💯

PR #166, opened 2023-03-03
RE2022-85: Integrate match deletion into service

MrCreosote:
Tested manually by creating a match and then altering its `last_access` field to be a week in the past.


Tianhao-Gu (approved):
👍

PR #167, opened 2023-03-06
run multi-task per node

Tianhao-Gu:
add the ability to run multiple tasks per node if needed.


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -216,17 +218,35 @@ def _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper
...
-def _create_batch_script(job_dir, task_list_file, n_jobs):
+def _cal_node_num(tool, n_jobs):
+    """
+    Calculate the number of nodes required for the task
+    """
+
+    tool_exe_time = TASK_META[tool]['exe_time']
+    jobs_per_node = NODE_TIME_LIMIT * 60 // tool_exe_time
+
+    num_nodes = min(math.ceil(n_jobs / jobs_per_node), MAX_NODE_NUM)
```
MrCreosote, line 229: What happens to the jobs that can't run because there aren't enough nodehours scheduled if `n_jobs/jobs_per_node > MAX_NODE_NUM`?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -216,17 +218,35 @@ def _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper
...
+
+    node_num = _cal_node_num(tool, n_jobs)
+
     batch_script = f'''#!/bin/sh
 
-#SBATCH -N {n_jobs + 1} -c 64
+#SBATCH -N {node_num + 1} -c 256
 #SBATCH -q regular
-#SBATCH --time=4:00:00
+#SBATCH --time={NODE_TIME_LIMIT}:00:00
```
MrCreosote, line 245: I'm not particularly familiar with slurm - once one of the tasks in a task list is completed on a node will it schedule another task on that node, and so on, until the node is out of time? I thought you might have to handle that part yourself in the task list, and code each task to run `jobs_per_node` jobs sequentially.

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -37,9 +38,11 @@
 
 TOOLS_AVAILABLE = ['gtdb_tk']  # TODO: fix checkm2 container bug
 
-# All nodes in Perlmutter have identical computational resources
-CHUNK_SIZE = {'checkm2': 5000,
-              'gtdb_tk': 1000}
+# estimated execution time (in minutes) for each tool to process a chunk of data
+TASK_META = {'checkm2': {'chunk_size': 5000, 'exe_time': 60},
+             'gtdb_tk': {'chunk_size': 1000, 'exe_time': 90}}
```
MrCreosote, line 43: I assume you'll check these? They probably need a bit of a fudge factor to be sure jobs don't fail

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote, line 245: If there are so few jobs you only need one node for less time it'd be nice to request that amount of time, but that can be done later IMO

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -216,17 +218,35 @@ def _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper
...
     """
     Create the batch script (submit_taskfarmer.sl) for job submission
     """
+
+    node_num = _cal_node_num(tool, n_jobs)
+
     batch_script = f'''#!/bin/sh
 
-#SBATCH -N {n_jobs + 1} -c 64
+#SBATCH -N {node_num + 1} -c 256
```
MrCreosote, line 243: Remind me what `-c` does?

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: Those nodes will hit wall-time and get killed. And re-submit is needed. Do you think we should remove `MAX_NODE_NUM`?

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: `taskfarmer` handles it. When there are more tasks than reserved nodes, tasks are distributed evenly across all nodes. And they are executed in sequence within a node.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: yea. I will test it for sure. According to my previous benchmarks, GTDB-TK takes about 30 mins for 1000 genomes with 128 cpus.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍 nice things to have.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: cpu num available for the workers.

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: My preference would be to not allow users to schedule a set of jobs where some are guaranteed to fail. Right now that means they're requesting 10000 hours of compute, which for our case, seems like plenty given running GTDB_TK on GTDB, which is the most expensive compute we currently support, is only ~500 hours.

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: Ok, that's what I assumed. I thought we said a ways back that there was diminishing returns to adding CPUs to GTDB (at least), and it might make sense to run multiple instances of the tool on the same node at the same time.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍  updated MAX_NODE_NUM = 100, and raise an error if the requested node num is greater than it, meaning they are requesting more than 1000 total node compute time.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: Do you still have questions on this? I am not sure what you are asking.

MrCreosote: I'm pretty sure I remember that the original benchmarking you did showed that there was a point where it wasn't very useful to throw more cores at GTDBTK and the time gains were minimal. IIRC we discussed that rather than assigning more cores per GTDBTK process, we'd get better throughput by running multiple GTDBTK processes on the same machine, splitting the cores between them.

This PR is quadrupling the number of cores for the GTDBTK process (unless I don't understand what's going here, which is def possible) which seems to contradict the above.

Tianhao-Gu: oh I see what you mean now. Each Perlmutter node has 256 cores. By doing `-c 256`,  I want to take all cores available. (I need to review this with Shane to be sure). Running multiple tasks on the same node is done via the arguments by `compute_genome_attri.py`. Right now, it was hardcoded as `threads=256, program_threads=256` in https://github.com/kbase/collections/blob/main/src/loaders/jobs/taskfarmer/task_generator.py#L179-L180
Meaning, we want to use 256 cores and each task also takes 256 cores (1 task per node). Once this is fully tested, I want to introduce `program_threads` argument, maybe some variable configured by tools. If we set it to 128, there will be 2 tasks running on the same node. But `-c` in the batch script doesn't affect how we want to schedule tasks per node.

MrCreosote: Ok, is there a ticket that covers changing the system to run > 1 task per node somewhere?

Tianhao-Gu: I really don't think we need a ticket for every single thing especially since we are developing something not well-defined. But https://kbase-jira.atlassian.net/browse/RE2022-113

MrCreosote: I don't want to lose track of this stuff, and unless we make a ticket we will *shrug*

MrCreosote (approved):
LGTM

PR #169, opened 2023-03-07
Ignore duplicate objects in match input

MrCreosote:
If there are multiple paths to the same object, drop all exept the shortest. If there's more than one with the shortest length, take the first.


Tianhao-Gu (approved):
👍

PR #170, opened 2023-03-07
RE2022-84: Support object sets in matches

MrCreosote:


src/service/workspace_wrapper.py
```diff
@@ -20,23 +23,57 @@ def __init__(self, workspace: Workspace):
...
+            set_upas = set()
+            for path, o in zip(paths, res):
+                if not o:
+                    # Race condition - the workspace allowed get_info3 but denied on get_objects2
+                    # Need to test with mocked workspace client
+                    raise errors.DataPermissionError(
+                        f"The workspace service disallowed acces to object {path}"
+                    )
+                set_upas.update([f'{path};{r}' for r in o['refs']])
+            if len(set_upas) > 10000:
```
Tianhao-Gu, line 68: using `match_retrieval.MAX_UPAS` here?

src/service/workspace_wrapper.py
```diff
@@ -47,16 +84,24 @@ def get_object_metadata(
...
+            if allowed_set_types and type_ in allowed_set_types:
+                set_objs.append(info[10])
+            elif allowed_types and type_ not in allowed_types:
                 raise errors.IllegalParameterError(
                     f"Workspace object {upa} is type {info[2]}, which is not one of "
                     + f"the allowed types: {allowed_types}"
                 )
-        return [info[10] for info in res['infos']]
+            else:
+                std_objs.append(info[10])
```
Tianhao-Gu, line 103: do you need if-elif-else.  Why not just do one if statement.  

`if allowed_types and type_ not in allowed_types: raise Error
  
std_objs.append(info[10])
`

src/service/workspace_wrapper.py
MrCreosote: This limit is slightly different than max upas as it's there because I'm lazy and didn't want to write a loop over get_object_info3 to do more than 10000 set objects. It might go away in the future, or the limit in match_retrieval might go away and this stays the same, but either way with the current implementation can't handle more than 10000 set UPAs even if we got rid of this check. I just put it here so we'd know if this happened, which could give us the impetus to write the loop or make a concrete decision to not support more than X UPAs in a match. As such my preference is to leave it alone for now

src/service/workspace_wrapper.py
MrCreosote: If I do
```python
            if allowed_set_types and type_ in allowed_set_types:
                set_objs.append(info[10])
            else allowed_types and type_ not in allowed_types:
                raise errors.IllegalParameterError(
                    f"Workspace object {upa} is type {info[2]}, which is not one of "
                    + f"the allowed types: {allowed_types}"
                )
            std_objs.append(info[10])
```
... which is what I think you're saying (?) then set objects will also get added to std objects

If I'm not understanding your suggestion LMK

Tianhao-Gu: Oh nv. I misread `set_objs` vs `std_objs`.

src/service/workspace_wrapper.py
Tianhao-Gu: 👍

Tianhao-Gu (approved):
LGTM

PR #174, opened 2023-03-08
RE2022-97: Send user params for matchers to matcher

MrCreosote:
Only real matcher is gtdb_lineage for now

Next - truncate lineages with rank parameter


Tianhao-Gu (approved):
👍

PR #175, opened 2023-03-08
add global task info file

Tianhao-Gu:
add global task info file so that we can look up job ids attached to each task. NERSC only caches job status for a certain period of time.

### DIFF
--- a/src/loaders/jobs/taskfarmer/task_generator.py
+++ b/src/loaders/jobs/taskfarmer/task_generator.py
@@ -2,7 +2,10 @@ import argparse
 import datetime
 import math
 import os
+import shutil
 import subprocess
+
+import jsonlines
 
 from src.loaders.common import loader_common_names
 
@@ -61,6 +64,10 @@ TOOL_VOLUME_MAP = {'checkm2': {CHECKM2_DB: '/CheckM2_database'},
 
 # Docker image tags for the tools
 DEFAULT_IMG_TAG = 'latest'
+
+# directory containing the TaskFarmer job scripts under the root directory
+TASKFARMER_JOB_DIR = 'task_farmer_jobs'
+TASK_INFO_FILE = 'task_info.jsonl'  # file containing the information of each task
 
 
 def _run_command(command, job_dir, log_file_prefix='', check_return_code=True):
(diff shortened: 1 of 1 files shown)


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -263,6 +271,37 @@ def _create_batch_script(job_dir, task_list_file, n_jobs, tool):
...
 
 
+def _create_job_dir(root_dir, tool, kbase_collection, load_ver, force=False):
+    """
+    Create the job directory
+
+    Job directory structure: <root_dir>/<TASKFARMER_JOB_DIR>/<kbase_collection>_<load_ver>_<tool>
+    """
+
+    job_dir = os.path.join(root_dir, TASKFARMER_JOB_DIR, f'{kbase_collection}_{load_ver}_{tool}')
```
MrCreosote, line 281: This doesn't take the source data dir into account, which maybe is ok, but the user probably needs to know that effectively the load version is bound to that source data dir unless they `--force` (which kind of makes sense).

Also I assume this will need adjusting when we start doing subsets of the source data dir, like bacdive or whatever?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -263,6 +271,37 @@ def _create_batch_script(job_dir, task_list_file, n_jobs, tool):
     return batch_script_file
 
 
+def _create_job_dir(root_dir, tool, kbase_collection, load_ver, force=False):
```
MrCreosote, line 274: So right now the script doesn't check the job dir to see what's already run, right, and it's just storing what was just run? That's coming later?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -263,6 +271,37 @@ def _create_batch_script(job_dir, task_list_file, n_jobs, tool):
...
+    """
+    Append the task information to the task info file
+    """
+
+    task_info_file = os.path.join(root_dir, TASKFARMER_JOB_DIR, TASK_INFO_FILE)
+
+    # Append the new record to the file
+    with jsonlines.open(task_info_file, mode='a') as writer:
+        if not os.path.exists(task_info_file):
+            writer.write([])
```
MrCreosote, line 301: I'm not understanding why this part is necessary

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍 added `source_data_dir`

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: you are right. I don't need this. removed.

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: Ok, so I see `source_data_dir` was added to the job info file, but the directory is still named solely based on the collection, tool, and load_ver. Is the strategy that on running again the code will check the `source_data_dir` in the job info file and warn or error if it's different than the `source_data_dir` in the program args?

MrCreosote: Also I still have the question about data subsets / bacdive etc. above

Tianhao-Gu: yes. following PRs will check if the current task has a conflict source_data_dir from the latest run and prompt users to use `-force` option. 

Regarding to subsets, I think each product, bacdive or gtdb, should have its own source data directory. And files within are just links to the NCBI directory.

MrCreosote: :+1:

MrCreosote (approved):

PR #177, opened 2023-03-09
add a task manager and check previous task status

Tianhao-Gu:


src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+            out put will be like below (with no job info just the header):
+            JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
+
+            * after the period of time, squeue command will return error code 1 and error message:
+            slurm_load_jobs error: Invalid job id specified
+        """
+
+        std_out_file, std_err_file, exit_code = tf_common.run_nersc_command(['squeue', '-j', str(job_id)],
+                                                                            self.job_dir, log_file_prefix='squeue',
+                                                                            check_return_code=False)
```
MrCreosote, line 134:
```suggestion
        std_out_file, std_err_file, exit_code = tf_common.run_nersc_command(
            ['squeue', '-j', str(job_id)], self.job_dir, log_file_prefix='squeue', check_return_code=False
        )
```

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
+import os
+import shutil
+from enum import Enum
+
+import pandas as pd
+
+import src.loaders.jobs.taskfarmer.taskfarmer_common as tf_common
+
+NERSC_SLURM_DONE_FILE = 'done.tasks.txt.tfin'
```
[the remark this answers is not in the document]
MrCreosote: Seems like the data is stored as `jsonl`?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def create_job_dir(self):
+        """
+        Create the job directory.
+        """
+        os.makedirs(self.job_dir, exist_ok=True)
+
+        return self.job_dir
+
+    def get_task_info_file(self):
```
MrCreosote, line 65: This method could also be private or inlined

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Retrieve the latest task status.
+
+        """
+        if not self.task_exists:
+            raise ValueError(f"Task does not exist for {self.kbase_collection}, {self.load_ver}, {self.tool}")
+
+        latest_task = self.get_latest_task()
+
+        done_file = os.path.join(self.job_dir, NERSC_SLURM_DONE_FILE)
```
MrCreosote, line 157: How does this file get created?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        The task info file is located under the TASKFARMER_JOB_DIR directory.
+        The file name is retrieved from TASK_INFO_FILE.
+        """
+        task_info_file = os.path.join(self.root_dir, tf_common.TASKFARMER_JOB_DIR, tf_common.TASK_INFO_FILE)
+
+        if not os.path.exists(task_info_file):
+            with open(task_info_file, 'w') as f:
+                pass  # creates an empty file
+
+        return task_info_file
```
MrCreosote, line 78: What actually writes to the task info file? I would suggest rather than having some other code write to it directly, have a method in this class that the code calls so that the file name / location / format can be changed in the future in this class without breaking outside code

MrCreosote (commented):
I've only reviewed the task manager so far but this review is getting pretty huge so I'll stop there for now

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
Tianhao-Gu: It is specific to the taskfarmer job output file.
https://docs.nersc.gov/jobs/workflow/taskfarmer/#output

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        self.kbase_collection = kbase_collection
+        self.load_ver = load_ver
+        self.tool = tool
+        self.root_dir = root_dir
+
+        self.job_dir = self.get_job_dir()
+        self.task_exists = os.path.isdir(self.job_dir)
+        self.tasks_df = self.retrieve_task_info()
+
+    def get_job_dir(self):
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        return self.job_dir
+
+    def get_task_info_file(self):
+        """
+        Get the task info file.
+
+        The task info file is located under the TASKFARMER_JOB_DIR directory.
+        The file name is retrieved from TASK_INFO_FILE.
+        """
+        task_info_file = os.path.join(self.root_dir, tf_common.TASKFARMER_JOB_DIR, tf_common.TASK_INFO_FILE)
```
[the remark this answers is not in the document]
Tianhao-Gu: at least before the RSV. I don't think we will have more than 100 tasks.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+
+class TFTaskManager:
+
+    def __init__(self, kbase_collection, load_ver, tool, root_dir):
+        self.kbase_collection = kbase_collection
+        self.load_ver = load_ver
+        self.tool = tool
+        self.root_dir = root_dir
+
+        self.job_dir = self.get_job_dir()
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        task_info_file = self.get_task_info_file()
+        df = pd.read_json(task_info_file, lines=True)
+
+        tasks_df = df[(df["kbase_collection"] == self.kbase_collection) &
+                      (df["load_ver"] == self.load_ver) &
+                      (df["tool"] == self.tool)]
+
+        tasks_df.sort_values(by="job_start_time", ascending=False, inplace=True)
+
+        return tasks_df
```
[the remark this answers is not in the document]
Tianhao-Gu: dataframe.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        Get the latest task information
+
+        Tasks are sorted by job_start_time in descending order. The latest task is the first row.
+        """
+        if not self.task_exists:
+            return {}
+
+        return self.tasks_df.iloc[0].to_dict()
+
+    def get_job_status_fm_nersc(self, job_id):
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+        When job is running, squeue command will return code 0 and out put will be like:
+        JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
+        5987779 regular_m submit_t      tgu  R       1:20      2 nid[005689-005690]
+
+        When job is finished/failed:
+            * job status is cached for a period of time (about 10 mins). squeue command will return code 0 and
+            out put will be like below (with no job info just the header):
+            JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
```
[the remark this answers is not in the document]
Tianhao-Gu: nope. If they keep job history, it will be a lot easier for us.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        if exit_code != 0:
+            # job (successfully) finished status is checked by checking the `done.tasks.txt.tfin` file in job directory
+            return JobStatus.FAILED
+
+        with open(std_out_file, "r") as std_out, open(std_err_file, "r") as std_err:
+            squeue_out = std_out.read().strip()
+            if str(job_id) in squeue_out:
+                return JobStatus.RUNNING
+            else:
+                return JobStatus.PENDING
```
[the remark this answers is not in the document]
Tianhao-Gu: complete is checked via the done file at the beginning.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+            * after the period of time, squeue command will return error code 1 and error message:
+            slurm_load_jobs error: Invalid job id specified
+        """
+
+        std_out_file, std_err_file, exit_code = tf_common.run_nersc_command(['squeue', '-j', str(job_id)],
+                                                                            self.job_dir, log_file_prefix='squeue',
+                                                                            check_return_code=False)
+
+        if exit_code != 0:
+            # job (successfully) finished status is checked by checking the `done.tasks.txt.tfin` file in job directory
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: once the task finishes, it will be generated in the job_dir by nersc slurm.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: Can you add some comments around this file; e.g. what it is, where it comes from, contents?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+    PENDING = 'PENDING'
+    RUNNING = 'RUNNING'
+    COMPLETED = 'COMPLETED'
+    FAILED = 'FAILED'
+
+
+class TFTaskManager:
+
+    def __init__(self, kbase_collection, load_ver, tool, root_dir):
+        self.kbase_collection = kbase_collection
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Same comment as https://github.com/kbase/collections/pull/177#discussion_r1134487789

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        task_info_file = os.path.join(self.root_dir, tf_common.TASKFARMER_JOB_DIR, tf_common.TASK_INFO_FILE)
+
+        if not os.path.exists(task_info_file):
+            with open(task_info_file, 'w') as f:
+                pass  # creates an empty file
+
+        return task_info_file
+
+    def retrieve_task_info(self):
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,164 @@
...
+        return job_dir
+
+    def recreate_job_dir(self):
+        """
+        Recreate the job directory.
+        """
+
+        print(f'recreating job dir {self.job_dir}')
+        if os.path.exists(self.job_dir):
+            shutil.rmtree(self.job_dir)
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: Add that link as well maybe

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: This is used internally only so I think that's fine, but can you prefix self.tasks_df with an underscore to mark it as such?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: Let me make sure I understand the possible states:

```
Done file  Exit Code   Job listed    Status
Y          Any         Any           COMPLETE
N          0           N             QUEUED or FAILED (if failed, exit code will flip to -1 within ~10m)
N          0           Y             RUNNING
N          -1          N/A           FAILED    
```

Is that right?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: Ok, IIUC, this is acting as a universal ledger for jobs and there are no deletions, only appends, right? If so, that seems reasonable for now, although we might want to switch to something like SQLite in the future so we can filter out older jobs.

The one thing I would suggest is locking the file around the write in case multiple jobs are started at the same time, see
https://stackoverflow.com/a/11853621/643675

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,179 @@
...
+        self.load_ver = load_ver
+        self.tool = tool
+        self.root_dir = root_dir
+
+        # job directory is named as <kbase_collection>_<load_ver>_<tool>
+        self.job_dir = os.path.join(self.root_dir, tf_common.TASKFARMER_JOB_DIR,
+                                    f'{self.kbase_collection}_{self.load_ver}_{self.tool}')
+        self.task_exists = os.path.isdir(self.job_dir) and os.listdir(self.job_dir)
+        self._create_job_dir(destroy_old_job_dir=destroy_old_job_dir)
+        self.tasks_df = self._retrieve_all_tasks()
```
MrCreosote: What's confusing to me here is that a client can call the `append_task_info` method, essentially creating a running task, IIUC, and get `task_exists` will remain false. I feel like I'm missing something here.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
Tianhao-Gu: Yes, currently only appending jobs. It makes sense to move to a proper job database when we have job runs periodically.

👍 Added file locking.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
Tianhao-Gu: Yea, there is a period of time when job is failed but this function returns `pending` status. But it only lasts for a short period of time. 

Added this matrix to the doc as a reference.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,197 @@
...
+        NOTE: If job fails and squeue command hasn't flipped the exit code to -1, this function will return 'PENDING'
+        as the job status. After about 10 mins, the exit code will be flipped to -1 and this function will return
+        'FAILED' status as the job status.
+        """
+
+        # Once all tasks are completed, NERSC generates this file.
+        # (ref: https://docs.nersc.gov/jobs/workflow/taskfarmer/#output)
+        done_file = os.path.join(self.job_dir, NERSC_SLURM_DONE_FILE)
+        if os.path.isfile(done_file):
+            return JobStatus.COMPLETED
```
MrCreosote, line 119: I'm confused here as well. What if you start a job, it finishes, and then you start a new job (say with a different source directory) with a new job_id. This function will still report the job as done, won't it?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,197 @@
...
+        self._tasks_df = self._retrieve_all_tasks()
+
+    def get_latest_task(self):
+        """
+        Get the latest task information
+
+        Tasks are sorted by job_submit_time in descending order. The latest task is the first row.
+        """
+
+        return self._tasks_df.iloc[0].to_dict() if self.task_exists else {}
```
MrCreosote, line 168: There's no use case for getting any task other than the latest?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: They don't make it easy to figure out what happened to your job, geez

Anyway :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,179 @@
...
+from enum import Enum
+
+import jsonlines
+import pandas as pd
+
+import src.loaders.jobs.taskfarmer.taskfarmer_common as tf_common
+from src.loaders.common import loader_common_names
+
+NERSC_SLURM_DONE_FILE = 'done.tasks.txt.tfin'
+REQUIRED_TASK_INFO_KEYS = ['kbase_collection', 'load_ver', 'tool', 'job_id', 'job_submit_time', 'source_data_dir']
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍 removed those 3 fields.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍  you are right. added updating task status after creating the task record.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: you can only start a new job by calling `task_generator.py` script which initializes TaskManager. At this point, task returned by the TaskManager will be the latest job created.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: not at this moment. But I think we will use tasks_df at some later point when we decide what tools has/hasn't run for a specific collection.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: when `_submit_job` is called, all `.tfin` files will be removed first. that including the done file. If that's what you are asking.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Maybe _submit_job should be a task manager method so that all the done file handling is encapsulated in the task manager?

Tianhao-Gu: 👍

MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_common.py
```diff
@@ -0,0 +1,37 @@
+import datetime
+import os
+import subprocess
+
+# directory containing the TaskFarmer job scripts under the root directory
+TASKFARMER_JOB_DIR = 'task_farmer_jobs'
+# file containing the information of each task
+TASK_INFO_FILE = 'task_info.jsonl'
```
MrCreosote, line 8: Seems like these two constants no longer need to be in common, they can just live in the manager

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _append_task_info(self, task_info):
+        """
+        Append the task information to the task info file
+        """
+
+        task_info.update({'kbase_collection': self.kbase_collection, 'load_ver': self.load_ver, 'tool': self.tool})
+
+        if not all(key in task_info for key in REQUIRED_TASK_INFO_KEYS):
+            raise ValueError(f"task_info must contain all keys: {REQUIRED_TASK_INFO_KEYS}")
```
MrCreosote, line 145: I would remove the 3 guaranteed keys from REQUIRED_TASK... otherwise this error message is going to be confusing

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,231 @@
+import datetime
```
MrCreosote, line 1: A couple behavior points I'm not sure about:

* the task manager is only going to know about the most recent job that ran for the coll/load_ver/tool combo when it started, or the most recent job that was run via `submit_job`. 
* If submit_job is called again for the same coll/load_ver/tool combo while jobs are running, things could get messy

Are those concerns we need to be worried about?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # If the .tfin files exist (these files are created by the previous run),
+        # delete them before starting a new run to avoid any issues caused by the previous run's files.
+        # TODO: investigate whether it is necessary to delete the .tfin files if a job fails due to walltime.
+        for filename in os.listdir(self.job_dir):
+            if filename.endswith(".tfin"):
+                os.remove(os.path.join(self.job_dir, filename))
+
+        current_datetime = datetime.datetime.now()
+        std_out_file, std_err_file, exit_code = tf_common.run_nersc_command(
```
MrCreosote, line 216: What if the taskfarmer script doesn't exist in `job_dir`?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,231 @@
...
+            raise ValueError(f"Task does not exist for {self.kbase_collection}, {self.load_ver}, {self.tool}")
+
+        latest_task = self.get_latest_task()
+        job_status = self._get_job_status_from_nersc(latest_task["job_id"])
+
+        return job_status
+
+    def submit_job(self, source_data_dir):
+        """
+        Submit the job to slurm
```
MrCreosote, line 204: Should doc the requirements for the job to work, e.g. the taskfarmer script named X needs to be in the job dir

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,231 @@
...
+        current_datetime = datetime.datetime.now()
+        std_out_file, std_err_file, exit_code = tf_common.run_nersc_command(
+            ['sbatch', os.path.join(self.job_dir, 'submit_taskfarmer.sl')],
+            self.job_dir, log_file_prefix='sbatch_submit')
+        with open(std_out_file, "r") as f:
+            sbatch_out = f.read().strip()
+            job_id = sbatch_out.split(' ')[-1]
+            print(f'Job submitted to slurm.\n{sbatch_out}')
+
+        task_info = {'job_id': job_id, 'job_submit_time': current_datetime.strftime("%Y-%m-%d %H:%M:%S"),
```
MrCreosote, line 224: should the job dir go in here as well?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,231 @@
...
+
+        self.kbase_collection = kbase_collection
+        self.load_ver = load_ver
+        self.tool = tool
+        self.root_dir = root_dir
+
+        # job directory is named as <kbase_collection>_<load_ver>_<tool>
+        self.job_dir = os.path.join(self.root_dir, tf_common.TASKFARMER_JOB_DIR,
+                                    f'{self.kbase_collection}_{self.load_ver}_{self.tool}')
+        self._task_exists = os.path.isdir(self.job_dir) and os.listdir(self.job_dir)
```
MrCreosote, line 176: I'm wondering if task exists should also check that there's a task in the task registry for the job dir

MrCreosote (commented):
Sorry, this is a bit of a peeling the onion review - as I get better understanding of the code and the code becomes clearer I keep finding new questions

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: never mind, you can figure out the job dir from the 3 i`__init__` args

src/loaders/jobs/taskfarmer/taskfarmer_common.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: if previous job is still running/pending, it will be caught by `_check_preconditions`.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: added file check

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

Tianhao-Gu:
previously, I didn't get the pending job status right. I did more experiments on NERSC with `squeue` and updated the `_get_job_status_from_nersc` method. 

Also, added `canceled` job status.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Maybe the manager should do the precondition check in submit_job?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_common.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,316 @@
...
+            * after cancellation, job is cached for a period of time (about 10 mins).
+            squeue command will return code 0 and out put will be like below (with no job info just the header):
+            JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
+            * after the period of time, squeue command will return error code -1 and error message:
+            slurm_load_jobs error: Invalid job id specified
+
+        Matrix of job status:
+        Done file  Exit Code   Job listed    Job Status Code  Returned Status
+        Y          Any         N/A           N/A              COMPLETED
+        N          0           N             PD               PENDING
```
MrCreosote, line 127:
```suggestion
        N          0           Y             PD               PENDING
```
Pretty sure that's right, let me know if I'm off base here

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,316 @@
...
+        if not all(key in task_info for key in REQUIRED_TASK_INFO_KEYS):
+            raise ValueError(f"task_info must contain all keys: {REQUIRED_TASK_INFO_KEYS}")
+
+        task_info.update({'kbase_collection': self.kbase_collection, 'load_ver': self.load_ver, 'tool': self.tool})
+
+        # Append the new record to the file
+        task_info_file = self._get_task_info_file()
+        with open(task_info_file, 'a') as writer:
+            fcntl.flock(writer.fileno(), fcntl.LOCK_EX)
+            json.dump(task_info, writer)
```
MrCreosote, line 178: out of curiosity why the switch from jsonlines?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,316 @@
...
+        df = pd.read_json(task_info_file, lines=True)
+
+        if df.empty:
+            return False
+
+        tasks_df = df[(df["kbase_collection"] == self.kbase_collection) &
+                      (df["load_ver"] == self.load_ver) &
+                      (df["tool"] == self.tool)]
+
+        return job_dir_exists or not tasks_df.empty
```
MrCreosote, line 201: Shouldn't this be an `and`?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,316 @@
...
+        Cancel a job on NERSC.
+        """
+
+        job_status, retry = self._get_job_status_from_nersc(job_id), 0
+
+        if job_status in [JobStatus.COMPLETED, JobStatus.FAILED, JobStatus.CANCELLED]:
+            return True
+
+        while job_status in [JobStatus.RUNNING, JobStatus.PENDING] and retry < 3:
+            time.sleep(2)
```
MrCreosote, line 214: I would think the sleep should go between the cancel call and the status call

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        Submit the job to slurm
+
+        Follow the steps in https://docs.nersc.gov/jobs/workflow/taskfarmer/#taskfarmer and generate all the necessary
+        files for the job submission (e.g. wrapper script, task list and batch script, etc.) before calling this function.
+        """
+
+        job_status, job_id = self.retrieve_latest_task_status()
+        if job_status in [JobStatus.RUNNING, JobStatus.PENDING]:
+            self._cancel_job(job_id)
```
MrCreosote, line 284: Hmm, I wouldn't do this unless a `force` param or something was sent to the method, rather throw an error by default

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,316 @@
...
+        # TODO: investigate whether it is necessary to delete the .tfin files if a job fails due to walltime.
+        for filename in os.listdir(self.job_dir):
+            if filename.endswith(".tfin"):
+                os.remove(os.path.join(self.job_dir, filename))
+
+        # check if all the required files exist
+        required_files = [tf_common.WRAPPER_FILE, tf_common.TASK_FILE, tf_common.BATCH_SCRIPT]
+        for filename in required_files:
+            if not os.path.isfile(os.path.join(self.job_dir, filename)):
+                raise ValueError(f"{filename} does not exist in {self.job_dir}")
```
MrCreosote, line 299: I would do these checks first thing, before canceling the job and removing done files

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: `writer.fileno()` doesn't work for` josnlines.open`

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍 I coded pretty late last night and my brain stopped working for a while.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,362 @@
...
+# file containing the information of each task
+TASK_INFO_FILE = 'task_info.jsonl'
+
+
+class TaskError(Exception):
+    def __init__(self, message):
+        self.message = message
+
+    def __str__(self):
+        return f"{self.message}"
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,362 @@
...
+            fcntl.flock(writer.fileno(), fcntl.LOCK_UN)
+
+    def _task_exists(self):
+        """
+        Check if the task for kbase_collection, load_ver, tool has been submitted.
+        """
+
+        if self.force_run:
+            # check if the job directory exists (directory maybe recreated during initialization)
+            job_dir_exists = os.path.isdir(self.job_dir)
```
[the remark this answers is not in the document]
Tianhao-Gu: updated `job_dir_exists = True` when self.force_run is True. I just need to ignore `os.listdir(self.job_dir)` when force_run is True.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,362 @@
...
+
+        if self.force_run:
+            # cancel the previous job if it is running or pending
+            if latest_task_status in [JobStatus.RUNNING, JobStatus.PENDING]:
+                self._cancel_job(job_id)
+            return True
+
+        if latest_task['source_data_dir'] != self.source_data_dir:
+            raise PreconditionError(
+                f'There is a previous run of the same tool and load version with a different source data directory.')
```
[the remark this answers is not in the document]
Tianhao-Gu: * job was cancelled only when force_run is true. This condition shouldn't happen before `if self.force_run` block. Unless I misunderstood you. 
* it is ignored when force_run is true.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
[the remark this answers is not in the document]
MrCreosote: > it is ignored when force_run is true.

I'm missing something because it looks to me like that check is run whether force_run is True or not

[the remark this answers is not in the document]
Tianhao-Gu: `if self.force_run:` returns True.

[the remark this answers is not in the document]
MrCreosote: Ugh, missed that.

:+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,357 @@
...
+        else:
+            print(f"Job {job_id} is canceled.")
+
+    def _get_latest_task(self):
+        """
+        Get the latest task information
+
+        Tasks are sorted by job_submit_time in descending order. The latest task is the first row.
+        """
+        return self._tasks_df.iloc[0].to_dict() if self._task_exists() else {}
```
MrCreosote, line 246: This does mean that if another run of the submission script occurs while this manager exists `_get_latest_task` will be out of date, but I'm not sure if that's a huge issue right now... thoughts?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,357 @@
...
+            time.sleep(2)
+            # Get the job status to confirm the job is canceled
+            job_status = self._get_job_status_from_nersc(job_id)
+
+        if job_status in [JobStatus.RUNNING, JobStatus.PENDING]:
+            raise ValueError(f"Failed to cancel job {job_id}.")
+        else:
+            print(f"Job {job_id} is canceled.")
+
+    def _get_latest_task(self):
```
MrCreosote, line 240: I think you could merge this method into `_retrieve_latest_task_status` as the latter is always called directly after this method. The return could be the task info with the status added into a new dict key

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,357 @@
...
+        self.tool = tool
+        self.source_data_dir = source_data_dir
+        self.root_dir = root_dir
+        self.force_run = force_run
+
+        # job directory is named as <kbase_collection>_<load_ver>_<tool>
+        self.job_dir = os.path.join(self.root_dir, TASKFARMER_JOB_DIR,
+                                    f'{self.kbase_collection}_{self.load_ver}_{self.tool}')
+
+        self._create_job_dir(destroy_old_job_dir=self.force_run)
```
MrCreosote, line 315: This is another item where I'm sort of thinking ahead a bit, but theoretically there could be some time between creation of the manager and a submit_job call. That means there could be a case where there is a currently running job and the manager is created and deletes the job directory, which could cause lots of weird failures.

Maybe the force_run argument should be provided to submit_job and directory deleted after check preconditions has run, so the prior job is cancelled first

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,357 @@
...
+        files for the job submission (e.g. wrapper script, task list and batch script, etc.) before calling this function.
+        """
+
+        # check if all the required files exist
+        required_files = [tf_common.WRAPPER_FILE, tf_common.TASK_FILE, tf_common.BATCH_SCRIPT]
+        for filename in required_files:
+            if not os.path.isfile(os.path.join(self.job_dir, filename)):
+                raise ValueError(f"{filename} does not exist in {self.job_dir}")
+
+        self._check_preconditions()
```
MrCreosote, line 332: Hmm, it's theoretically possible that > 1 instances of the job submission script could be run at the same time for the same job dir and so > 1 sets of jobs kicked off... not sure if that's worth worrying about right now.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,357 @@
+import datetime
```
MrCreosote, line 1: I'm guessing the next step after this PR is figuring out what's already been run and resubmitting just failures / incompletes?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        Check if the task for kbase_collection, load_ver, tool has been submitted.
+        """
+
+        if self.force_run:
+            # The job directory is recreated as part of the initialization process.
+            job_dir_exists = True
+        else:
+            # check if the job directory exists and is not empty
+            job_dir_exists = os.path.isdir(self.job_dir) and os.listdir(self.job_dir)
```
MrCreosote, line 201: I'm still having trouble with the _task_exists logic. I'm not sure I understand why the directory is taken into account. If the job exists, shouldn't that be the gold standard? 

For example, if someone manually deletes the job dir and then runs the task submission script again, the new jobs will be scheduled while the old jobs are still running, IIUC

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: well if there are multiple instances of manager and each of them submits a job, one of the `_get_latest_task` will be out of date. But I think it's unlikely to happen, for most of the time I assume we will use this module together with task_generator.py.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: I am not sure. There are a lot of edge cases that could happen during a job run. Should we worry about them at this stage? I would say no.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: yup. that sounds like a good next step.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍 yea. you are probably right. removed job_dir check.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: I found a bug after making submit_job take force_run and recreate job_dir at the check precondition step. We actually want to recreate the folder before all scripts are generated which is done by task_generator.py.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: ok. updated the method so that it's retrieving the task list from the task file every time it's called.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Thinking about this a bit more, the task manager isn't responsible for setting up the job dir - that's the responsibility of the calling script. Maybe the task manager shouldn't delete the job dir at all and force (or maybe a better named argument) just kills any currently running job

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Fair enough :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        Submit the job to slurm
+
+        Follow the steps in https://docs.nersc.gov/jobs/workflow/taskfarmer/#taskfarmer and generate all the necessary
+        files for the job submission (e.g. wrapper script, task list and batch script, etc.) before calling this function.
+
+        :param restart_on_demand: if True, killed any running/pending jobs and restart running the tool again.
+        """
+
+        # check if all the required files exist
```
MrCreosote, line 300: This method used to delete *.tfin files before it started. If it doesn't do that, won't that mess up the job status checking method, which IIUC interprets the presence of tfin files as meaning the job is successful?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Append the new record to the file
+        task_info_file = self._get_task_info_file()
+        with open(task_info_file, 'a') as writer:
+            fcntl.flock(writer.fileno(), fcntl.LOCK_EX)
+            json.dump(task_info, writer)
+            writer.write('\n')
+            fcntl.flock(writer.fileno(), fcntl.LOCK_UN)
+
+    def _task_exists(self):
```
MrCreosote, line 177: `_task_exists` and `_retrieve_all_tasks` are almost identical in function - I think they could be merged into a single function that returns None or tasks_df, with None implying no tasks exists. 

I probably should've noticed this earlier, sorry

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # check if all the required files exist
+        required_files = [tf_common.WRAPPER_FILE, tf_common.TASK_FILE, tf_common.BATCH_SCRIPT]
+        for filename in required_files:
+            if not os.path.isfile(os.path.join(self.job_dir, filename)):
+                raise ValueError(f"{filename} does not exist in {self.job_dir}")
+
+        self._check_preconditions(restart_on_demand)
+
+        os.chdir(self.job_dir)
```
MrCreosote, line 308: Why is this necessary? The job dir is provided to nersc running function

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -66,44 +64,16 @@
...
     """
     Pulls the specified Shifter image from the specified registry.
     """
 
     print(f"Fetching Shifter image {image_str}...")
-    sp_std_out_file, sp_std_err_file, sp_exit_code = _run_command(["shifterimg", "pull", image_str], job_dir,
-                                                                  log_file_prefix='shifterimg_pull')
+    sp_std_out_file, sp_std_err_file, sp_exit_code = tf_common.run_nersc_command(["shifterimg", "pull", image_str],
+                                                                                 job_dir,
+                                                                                 log_file_prefix='shifterimg_pull')
```
MrCreosote, line 76:
```suggestion
    sp_std_out_file, sp_std_err_file, sp_exit_code = tf_common.run_nersc_command(
        ["shifterimg", "pull", image_str], job_dir, log_file_prefix='shifterimg_pull')
```

MrCreosote:
It's been a long road but almost there I think

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: no it won't mess up job status checking. NERSC keeps track of tasks using those files. I was only testing 1 job before and thought those are not needed. But when I tested multiple jobs, those files are very helpful. Say 1 one 5 jobs failed. When we resubmit, it will once run the failed 1.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: this is to make sure slurm logs will be created in side the job dir. Otherwise, it will be created where you run task_generator script.

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Why not just pass the job dir in the `cwd` argument to subprocess? https://docs.python.org/3/library/subprocess.html#subprocess.run

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: But if you run a new job and the tfin files are still there, the task manager will think the current job is done, right, and so would allow starting another job even if `force` is false?

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: There's a few other places in the file that have the squished up command lines, can you fix those?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: force will recreate the job_dir.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: I guess they are the same thing.

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Except that if you do it in the function it's less likely to cause bugs, because ppl calling the function may not realize they need to chdir themselves. Makes the function self contained and easier to use

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: yea. in case of failed/canceled/time-out.

MrCreosote: Hold on, my logic is bad. What about this scenario:

* Run job, job finishes successfully, leaving `tfin` files.
* Run another job for the same workdir. This is allowed since the job is done.
* `tfin files are still there so the task manager will think the just submitted job is done
* Run another job for the same workdir. Since the task manager thinks the 2 previous jobs are both done, it'll allow the 3rd job to start even without `force`, which isn't supposed to happen

Is that right?

Tianhao-Gu: `Run another job for the same workdir. This is allowed since the job is done.` When there is a previous done job, task manager will not allow another run without --force.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Ok, you're right. In that case, should `--force` cause the task manager to error if there are tfin files, or maybe just delete them?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Why `shell` vs `cwd`? `shell` has security issues and so should be avoided if possible
https://docs.python.org/3/library/subprocess.html#security-considerations

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: force will recreate the job_dir meaning we no longer care about previous runs. tfin files from previous run are not needed any more.

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Also the current code still required the caller of the run_nersc_command option to run a cd, which is what I wanted to avoid

MrCreosote: IOW
```
p = subprocess.Popen(command, stdout=std_out, stderr=std_err, text=True, cwd=job_dir)```

Tianhao-Gu: avoid 'cd' completely?

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: But there's no guarantee in the task manager that happens. The task manager should check that conditions are right to start the job and not rely on the calling function

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: > any strong reasons, we want to use subprocess? It will create a log file if I use the helper run_command function. Is that worth it for a cd command?

Sorry I missed this comment earlier. I don't want to call the run nersc function to do the cd, I want, effectively, the cd to happen in the run nersc function like above

MrCreosote: Oh, you did this in the latest commit. Lots of race conditions here

MrCreosote: :+1:

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: hmmm now you completely lost me. when you call `submit_job`, it does `_check_preconditions`. that function checks all conditions a job need to run.

Tianhao-Gu: 👍 removed all tfin files when restart_on_demand is checked.

MrCreosote: :+1:

MrCreosote (approved):
LGTM :tada:

PR #179, opened 2023-03-10
Fix lineage parsing

MrCreosote:
GTDB_tk always outputs full ranks; a missing name indicates an unresolved rank.

Also move the parsing logic to the class constructor.


Tianhao-Gu (approved):
👍

PR #183, opened 2023-03-15
RE2022-115: Create selection

MrCreosote:
Still need to fire off the process that actually marks things as selected


Tianhao-Gu (approved):
👍

PR #184, opened 2023-03-15
RE2022-115: Add get selection endpoint

MrCreosote:


src/service/routes_collections.py
```diff
@@ -405,6 +431,41 @@ async def get_match(
...
+    KBASE_COLLECTIONS_SELECTION: str = Header(
+        description="The selection ID / token returned when creating a selection."
+    ),
+    verbose: bool = _QUERY_SELECTION_VERBOSE,
+) -> Selection:
+    appstate = app_state.get_app_state(r)
+    hashed_token = _hash_token(KBASE_COLLECTIONS_SELECTION)
+    active_sel = await appstate.arangostorage.get_selection_active(hashed_token)
+    # could save bandwidth by passing verbose to DB layer and not pulling IDs
+    internal_sel = await appstate.arangostorage.get_selection_internal(
```
Tianhao-Gu, line 453: Would it be beneficial to include an option in get_selection_internal that allows the caller to exclude selection_ids from the return?

Tianhao-Gu (approved):

src/service/routes_collections.py
MrCreosote: Sure, but I was only planning to do that if it became an issue. Since the changes are all internal to the service it wouldn't break backwards compatibility

I could possibly be convinced otherwise, but if so let's do it in another PR

PR #187, opened 2023-03-17
avoid overwritting files

Tianhao-Gu:
there shouldn't be any case where we want to overwrite batch scripts if they exist, though I think if they are identical it won't affect anything.

### DIFF
--- a/src/loaders/jobs/taskfarmer/task_generator.py
+++ b/src/loaders/jobs/taskfarmer/task_generator.py
@@ -316,7 +316,11 @@ def main():
     batch_script = _create_batch_script(job_dir, task_list_file, n_jobs, tool)
 
     if args.submit_job:
-        task_mgr.submit_job()
+        try:
+            task_mgr.submit_job()
+        except PreconditionError as e:
+            raise ValueError(f'Error submitting job:\n{e}\n'
+                             f'Please use the --force flag to overwrite the previous run.') from e
     else:
         print(f'Please go to Job Directory: {job_dir} and submit the batch script: {batch_script} to the scheduler.')
 
--- a/src/loaders/jobs/taskfarmer/task_generator.py
+++ b/src/loaders/jobs/taskfarmer/task_generator.py
@@ -87,20 +87,18 @@ def _pull_image(image_str, job_dir):
                              f"Standard error: {std_err.read()}")
 
 
-def _write_to_file(file_path, content, mode="x", ignore_existing=False):
+def _write_to_file(file_path, content, overwrite=False):
     """
     Writes the specified content to the specified file.
 
-    If the file already exists, an exception is raised unless ignore_existing is set to True.
+    If the file already exists, an exception is raised unless overwrite is set to True.
     """
+    mode = 'w' if overwrite else 'x'
     try:
-        with open(file_path, mode) as f:
-            f.write(content)
-    except FileExistsError as e:
-        if ignore_existing:
-            print(f"File {file_path} already exists.")
-        else:
-            raise ValueError(f"File {file_path} already exists.") from e
+        with open(file_path, mode) as file:
+            file.write(content)
+    except FileExistsError:
+        print(f"File '{file_path}' already exists. Set '--force' flag to True to overwrite the file.")
 
 
 def _fetch_image(registry, image_name, job_dir, tag='latest', force_pull=True):
@@ -134,7 +132,7 @@ def _fetch_image(registry, image_name, job_dir, tag='latest', force_pull=True):
     return image_str
 
 
-def _create_shifter_wrapper(job_dir, image_str):
+def _create_shifter_wrapper(job_dir, image_str, overwrite=False):
     """
     Creates the Shifter wrapper script.
     """
@@ -157,24 +155,24 @@ cd {job_dir}
 shifter --image=$image $command'''
 
     wrapper_file = os.path.join(job_dir, tf_common.WRAPPER_FILE)
-    _write_to_file(wrapper_file, shifter_wrapper, ignore_existing=True)
+    _write_to_file(wrapper_file, shifter_wrapper, overwrite=overwrite)
 
     os.chmod(wrapper_file, 0o777)
 
     return wrapper_file
 
 
-def _create_genome_id_file(genome_ids, genome_id_file):
+def _create_genome_id_file(genome_ids, genome_id_file, overwrite=False):
     """
     Create a tab-separated values (TSV) file with a list of genome IDs.
     """
 
     content = "genome_id\n" + "\n".join(genome_ids)
-    _write_to_file(genome_id_file, content, ignore_existing=True)
+    _write_to_file(genome_id_file, content, overwrite=overwrite)
 
 
 def _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper_file, job_dir, root_dir,
-                      threads=256, program_threads=256, source_file_ext='genomic.fna.gz'):
+                      overwrite=False, threads=256, program_threads=256, source_file_ext='genomic.fna.gz'):
     """
     Create task list file (tasks.txt)
     """
@@ -191,7 +189,7 @@ def _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper
         task_list += wrapper_file
 
         genome_id_file = os.path.join(job_dir, f'genome_id_{idx}.tsv')
-        _create_genome_id_file(genome_ids_chunk, genome_id_file)
+        _create_genome_id_file(genome_ids_chunk, genome_id_file, overwrite=overwrite)
 
         env_vars = {'TOOLS': tool, 'LOAD_VER': load_ver, 'SOURCE_DATA_DIR': source_data_dir,
                     'KBASE_COLLECTION': kbase_collection, 'ROOT_DIR': root_dir,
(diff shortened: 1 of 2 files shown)


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -83,6 +87,17 @@ def _pull_image(image_str, job_dir):
...
 
+def _write_to_file(file_path, content, mode="x"):
+    """
+    Writes the specified content to the specified file.
+    """
+    try:
+        with open(file_path, mode) as f:
+            f.write(content)
+    except FileExistsError:
+        print(f"File {file_path} already exists.")
```
MrCreosote, line 98: This makes me very nervous as it could cause difficult to find bugs if files are overwritten but people aren't carefully checking the logs. I'd prefer to throw the error by default and have a flag to ignore file collisions

MrCreosote (changes requested):

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -83,6 +87,22 @@ def _pull_image(image_str, job_dir):
...
+    If the file already exists, an exception is raised unless ignore_existing is set to True.
+    """
+    try:
+        with open(file_path, mode) as f:
+            f.write(content)
+    except FileExistsError:
+        if ignore_existing:
+            print(f"File {file_path} already exists.")
+        else:
+            raise FileExistsError(f"File {file_path} already exists.")
```
MrCreosote, line 103:
```suggestion
            raise
```
any reason not to use the same exception?

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: I meant there should be a cli flag to overwrite existing files

Tianhao-Gu: that's done by giving the 'w' mode.

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: That's not a CLI flag?

MrCreosote: Thinking about this further, IIUC if any of the submission scripts / text files exist, they won't be overwritten by this logic. I guess I don't understand why you wouldn't just generate all new scripts (maybe backing up the old ones if necessary)

Tianhao-Gu: My aim was to avoid overwriting existing files, particularly in situations where a user runs the script multiple times. For instance, if a user executes the script again without specifying the --force option, it will overwrite all existing files. This may not be problematic if all the arguments remain the same. However, if the user runs the script with a different source_dir argument and without --force, the situation becomes complicated.

MrCreosote: I feel like I'm missing something here - IIUC if I run the script twice in a row with different arguments, the script will ignore the 2nd set of arguments because it doesn't overwrite the files. That seems like the exact opposite of the behavior anyone would expect

MrCreosote: Maybe we should zoom on Monday and talk through this, this thread is getting long

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -237,10 +264,6 @@ def _check_preconditions(self, restart_on_demand):
 
             return True
 
-        if latest_task['source_data_dir'] != self.source_data_dir:
```
Tianhao-Gu, line 240: This seems to be irrelevant at this point. If the previous job failed/canceled/timeout, it's okay to restart the job with a different argument. If the previous job is completed/running/pending, the job won't run unless `force`. If the user force runs the job, it also becomes irrelevant.

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -292,6 +306,7 @@ def main():
     root_dir = args.root_dir
 
     task_mgr = TFTaskManager(kbase_collection, load_ver, tool, source_data_dir, root_dir=root_dir)
+    task_mgr.check_preconditions(args.force)
```
Tianhao-Gu, line 309: I think we should just do the precondition check before writing batch scripts. When the previous job is in complete/running/pending status, we shouldn't update batch scripts. However, when '--force' or the previous job is in failed/canceled/timeout status, we should update the batch script and allow users to resubmit jobs with the same/different arguments.

Tianhao-Gu: after precondition checking is passed, (the previous job is in failed/canceled/timeout status or --force is True), it's safe to overwrite existing batch files.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍 made a few changes. But sure let's sync up on Monday.

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -83,6 +87,20 @@ def _pull_image(image_str, job_dir):
                              f"Standard error: {std_err.read()}")
 
 
+def _write_to_file(file_path, content, overwrite=True):
```
MrCreosote, line 90: Looks like none of the calls specified overwrite=False, so I assume that arg can be removed

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -277,11 +302,7 @@ def _check_preconditions(self, restart_on_demand):
...
-            raise PreconditionError(
-                f'There is a previous run of the same tool and load version with a different source data directory.')
-
-        if latest_task_status in [JobStatus.FAILED, JobStatus.CANCELLED, JobStatus.TIMEOUT]:
+        if latest_task_status in [JobStatus.FAILED, JobStatus.CANCELLED]:
```
MrCreosote, line 305: Was timeout deleted intentionally?

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote, line 309: If the preconditions check needs to happen every time, just call it from the class constructor

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: Hmm, what about the case where the job failed on 1 genome but all the others are ok?

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: we do have a ticket for that situation. But I haven't thought through how we can do this. 
https://kbase-jira.atlassian.net/browse/RE2022-96
I feel this still has nothing to do with `source_data_dir`.

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -305,7 +313,7 @@ def main():
 
     if args.submit_job:
         try:
-            task_mgr.submit_job(restart_on_demand=args.force)
+            task_mgr.submit_job()
         except PreconditionError as e:
```
MrCreosote, line 317: so submit_job can't throw a precondition error as is, but maybe both the constructor and the submit call should check preconditions in case there's a change in between. Either way the I assume the precondition error needs to be caught when calling the constructor

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: Files are always overwritten if the script proceeds at this point so moot

:+1:

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: added back `source_data_dir` check.

MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py

MrCreosote (approved):
LGTM

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍 updated to catch preconditionError at both places.

PR #188, opened 2023-03-17
add timeout status

Tianhao-Gu:
not sure if it's essential to add a timeout status. That means the task hit wall time. We could just treat the timeout task the same as a failed task. But I think maybe down the road we want to do something differently.

### DIFF
--- a/src/common/storage/collection_and_field_names.py
+++ b/src/common/storage/collection_and_field_names.py
@@ -64,6 +64,9 @@ COLL_SRV_DATA_PRODUCT_PROCESSES = _SRV_PREFIX + "data_prod_proc"
 COLL_SRV_SELECTIONS = _SRV_PREFIX + "selections"
 """ A collection holding selections for Collections. """
 
+COLL_SRV_SELECTIONS_DELETED = COLL_SRV_SELECTIONS + "_deleted"
+""" A collection holding selections in the deleted state. """
+
 ## Data product collections
 
 ### Taxa counts
--- a/src/service/app_state.py
+++ b/src/service/app_state.py
@@ -64,9 +64,9 @@ async def destroy_app_state(app: FastAPI):
     """
     Destroy the application state, shutting down services and releasing resources.
     """
-    state = _get_app_state_from_app(app)
-    state._match_deletion.stop()
-    await state._colstate.destroy()
+    app.state._match_deletion.stop()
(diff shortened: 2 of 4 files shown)


src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -50,6 +53,33 @@ class TFTaskManager:
     The task manager is responsible for managing the tasks of TaskFarmer jobs.
     """
 
+    @staticmethod
+    def _read_last_n_line(file_path, n=5):
```
MrCreosote, line 57: I'd use a package for this rather than rolling your own, https://pypi.org/project/file-read-backwards/ might be reasonable

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
```diff
@@ -36,8 +38,9 @@ class JobStatus(Enum):
     PENDING = 'PENDING'
     RUNNING = 'RUNNING'
     COMPLETED = 'COMPLETED'
-    FAILED = 'FAILED'
+    FAILED = 'FAILED/CANCELLED'  # in many situations, we cannot identify whether a job has failed or cancelled
```
MrCreosote, line 41: Even with the slurm log it's impossible to tell if the job failed vs cancelled? There's no exit code? it seems surprising that slurm doesn't log that the job was cancelled somewhere

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/taskfarmer_task_mgr.py
Tianhao-Gu: yea. they both just showed 'KILLED' in the slurm log. Maybe as I do more testing, I will come up with some ways to identify those. But I think we could treat canceled as failed or vice visa at this time just fine.

MrCreosote: Weird that cancellation due to timeout is logged clearly but regular cancellation is indistinguishable from failure

MrCreosote (approved):

Issue #163, opened 2023-03-02
Figure out what do to about abandonware async arango clients


MrCreosote:
We're currently using https://github.com/mirrorrim/aioarango for ArangoDB comms, but that repo hasn't been updated in almost 2 years. Questions about that are going unanswered: https://github.com/mirrorrim/aioarango/issues/3 https://github.com/mirrorrim/aioarango/issues/5 
The main client repo has no plans to support async/await: https://github.com/ArangoDB-Community/python-arango/issues/95
The only other async lib I can find also appears to be abandonware: https://github.com/getflaps/aioarangodb

The whole service is structured around async/await and undoing that would be painful and also probably lead to a performance hit on longer queries.

Currently it's not causing any problems, but some library dependencies are locked to specific major versions and so we won't be able to update those packages once they do a major release:
```
aioarango==1.0.0
  - httpx [required: >=0.18.2,<0.19.0, installed: 0.18.2]
    - certifi [required: Any, installed: 2022.12.7]
    - httpcore [required: >=0.13.3,<0.14.0, installed: 0.13.7]
      - anyio [required: ==3.*, installed: 3.6.2]
        - idna [required: >=2.8, installed: 3.4]
        - sniffio [required: >=1.1, installed: 1.3.0]
      - h11 [required: >=0.11,<0.13, installed: 0.12.0]
      - sniffio [required: ==1.*, installed: 1.3.0]
    - rfc3986 [required: >=1.3,<2, installed: 1.5.0]
    - sniffio [required: Any, installed: 1.3.0]
  - PyJWT [required: >=2.1.0,<3.0.0, installed: 2.6.0]
  - requests-toolbelt [required: >=0.9.1,<0.10.0, installed: 0.9.1]
    - requests [required: >=2.0.1,<3.0.0, installed: 2.28.2]
      - certifi [required: >=2017.4.17, installed: 2022.12.7]
      - charset-normalizer [required: >=2,<4, installed: 2.1.1]
      - idna [required: >=2.5,<4, installed: 3.4]
      - urllib3 [required: >=1.21.1,<1.27, installed: 1.26.14]
```

For example, httpx is now on version `0.23.3` (https://github.com/mirrorrim/aioarango/issues/7)

MrCreosote:
I suppose on the dependency front we could go back to `pip`... meh

MrCreosote:
[...]

PR #190, opened 2023-03-20
RE2022-116: Add selection state update method

MrCreosote:
Also DRY up the 3 state update methods.


MrCreosote:
Ugh, one of those trivy alerts is for httpx, which is pinned to < 19.0 by aioarango, which https://github.com/kbase/collections/issues/163

dauglyon (approved):
This all LGTM... but I can't say I am confident in my ability to be confident about AQL. That said, LGTM

PR #191, opened 2023-03-20
RE2022-116: Apply selection to genome attribs data

MrCreosote:


src/service/data_products/genome_attributes.py
```diff
@@ -552,3 +562,47 @@ async def delete_match(storage: ArangoStorage, internal_match_id: str):
...
+            OPTIONS {{exclusive: true}}
+            LET updated = NEW
+            RETURN KEEP(updated, "{names.FLD_GENOME_ATTRIBS_KBASE_GENOME_ID}")
+        """
+    bind_vars = {
+        f"@{_FLD_COL_NAME}": names.COLL_GENOME_ATTRIBS,
+        _FLD_COL_ID: coll.id,
+        _FLD_COL_LV: load_ver,
+        "genome_ids": sel.selection_ids,
+        "internal_match_id": internal_selection_id,
```
MrCreosote, line 597: This should be `"internal_selection_id"`. It's just a name and localized to this fn so works fine, but I have it fixed locally and the fix will show up in a future PR.

src/common/storage/collection_and_field_names.py
```diff
@@ -102,3 +102,7 @@
 # Used for marking matches when returning data to the user and they select match marking vs.
 # filtering on a match.
 FLD_GENOME_ATTRIBS_MATCHED = "__match__"
+
+# Used for marking selections in the Arango collection; contains a list of internal selection IDs.
+# Underscore to separate from "real" attribs
+FLD_GENOME_ATTRIBS_SELECTIONS = "_selections"
```
dauglyon, line 108: Why the different key format than `__match__`?

dauglyon: would prefer the same for clarity unless I'm missing something

dauglyon (approved):

src/common/storage/collection_and_field_names.py
MrCreosote: This is an internal only key, it's not exposed by the API. The selection mark key will be `__sel__`

PR #192, opened 2023-03-21
RE2022-116: refactor selection handling into module

MrCreosote:
... in preparation to add process creation

### DIFF
--- a/src/service/models.py
+++ b/src/service/models.py
@@ -11,7 +11,7 @@ may still pass. In this case, a translation from the older data is required.
 """
 
 from enum import Enum
-from pydantic import BaseModel, Field, validator, HttpUrl
+from pydantic import BaseModel, Field, validator, HttpUrl, root_validator
 from typing import Any, Optional
 
 from src.service.arg_checkers import contains_control_characters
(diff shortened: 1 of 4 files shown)


src/service/selection_processing.py
MrCreosote, line 1: This is basically just C&P from `routes_collections.py`

src/service/tokens.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def create_token(prefix=None):
+    """
+    Create a URL safe token using the default python bit size for secrets.
+
+    prefix - an optional prefix to apply to the token string.
+    """
+    prefix = prefix if prefix else ""
+    return "coll-selection-" + secrets.token_urlsafe()  # 256 bits by default
```
MrCreosote, line 16: So this is a bug. That being said, with the md5ification of the selection IDs I'm working on right now, this file gets entirely deleted, so it's moot

src/service/models.py
```diff
@@ -173,6 +176,14 @@ class Collection(BaseModel):
     matchers: list[Matcher] = Field(
         description="The matchers associated with the collection"
     )
+    default_select: str | None = Field(  # might need to make this a list in future...? not sure
+        **DATA_PRODUCT_ID_FIELD_PROPS | {
+        "description":
+            "The ID of the data product to which non-data product specific selections "
```
dauglyon, line 182: I'm not sure I understand what this means

dauglyon: Does this essentially mean selections are "part" of the genome attribs data product for now?

MrCreosote: It tells the system which data product is the "primary"  or "default" in terms of where to initially apply the selection. For now, the only choice we have is genome attribs, but this allows setting a different data product in the future.

So when saving a collection, you'd specify which data product the selection should initially be applied to here, and then other data products would calculate their selections from that primary selection.

MrCreosote: On the other hand, a selection against taxa counts couldn't work that way so it would deal with figuring out the selection itself

dauglyon (approved):

PR #201, opened 2023-03-23
move match deletion to deletion

MrCreosote:
in preparation for resusing the scheduler for selection deletion


dauglyon (approved):

dauglyon (approved):
approved but logging off for the evening

PR #202, opened 2023-03-23
move match deletion test to deletion test

MrCreosote:


dauglyon (approved):
the triviometer has ascended into another plane. lgtm

PR #203, opened 2023-03-23
RE2022-117: Simplify method sigs for data product processes

MrCreosote:


src/service/models.py
```diff
@@ -432,7 +432,27 @@ class ProcessType(str, Enum):
     """ A seletion process. """
 
 
-class DataProductProcess(ProcessAttributes):
+class DataProductProcessIdentifier(BaseModel):
```
dauglyon, line 435: Idk if I love this verbiage, but I get it, once I understood it.

dauglyon: just a little odd to have the type/data product ID also in there.

DataProductProccessLocator? Idk.

dauglyon (approved):
approved regardless

src/service/models.py
MrCreosote: yeah, not a huge fan either but it describes what it does and i couldn't think of anything better. If you have better ideas I'm happy to consider it

PR #207, opened 2023-03-23
RE2022-116: Switch selections from token to md5 based ID

MrCreosote:


dauglyon (approved):
As far as I can tell everything looks good. It may be worth getting another set of eyes if you want extra confidence, but things made sense to me

PR #208, opened 2023-03-23
RE2022-130: Don't alter counts if match/selections are just marks

MrCreosote:
Only alter the count if they're set up to filter the main results


dauglyon (approved):
Excellent!

PR #220, opened 2023-03-30
make taskfarmer handles parallelization

Tianhao-Gu:


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -42,13 +42,12 @@
 
 # estimated execution time (in minutes) for each tool to process a chunk of data
 TASK_META = {'checkm2': {'chunk_size': 5000, 'exe_time': 60},
-             'gtdb_tk': {'chunk_size': 1000, 'exe_time': 90}}
-NODE_TIME_LIMIT = 10  # hours
+             'gtdb_tk': {'chunk_size': 1000, 'exe_time': 65}}
+NODE_TIME_LIMIT = 5  # hours  # TODO: automatically calculate this based on tool execution time and NODE_THREADS
```
Tianhao-Gu, line 46: per benchmark, each task will take ~50 minutes. Manually set exe_time and NODE_TIME_LIMIT so that we reserve just enough nodes with 4 tasks per node. More configurable settings will be done at a later date.

MrCreosote: 5 * 300 / 65 = 4.6 batches of 4 1000 genome gtdb_tk runs. Why 4.6?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -166,9 +165,14 @@ def _create_genome_id_file(genome_ids, genome_id_file):
 
 
 def _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper_file, job_dir, root_dir,
-                      threads=256, program_threads=256, source_file_ext='genomic.fna.gz'):
+                      threads=32, program_threads=32, source_file_ext='genomic.fna.gz'):
```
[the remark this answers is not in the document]
MrCreosote: Just to be sure I understand, setting theads & program threads the same means that the calculation script will only run 1 instance of the program, and that instance will use `threads` threads? If I'm right could you document that super clearly?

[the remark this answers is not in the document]
MrCreosote: Also can you document why the defaults are 32

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -42,13 +42,12 @@
...
-NODE_TIME_LIMIT = 10  # hours
+             'gtdb_tk': {'chunk_size': 1000, 'exe_time': 65}}
+NODE_TIME_LIMIT = 5  # hours  # TODO: automatically calculate this based on tool execution time and NODE_THREADS
 MAX_NODE_NUM = 100  # maximum number of nodes to use
 # The THREADS variable controls the number of parallel tasks per node
-# we want to set it to 1 (execute tasks one by one) because batch parallelization is handled by the
-# compute_genome_attribs.py script
-NODE_THREADS = 1
+# TODO: make this configurable based on tool used
+NODE_THREADS = 4
```
MrCreosote, line 50: Can you document the reason for this value of NODE_THREADS

src/loaders/jobs/taskfarmer/task_generator.py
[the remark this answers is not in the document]
Tianhao-Gu: yes. parallelization with a task is controlled by those two arguments. Within a task, `batch_num = threads// program_threads`.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: The logic to compute task per node is` 5 * 60 // 65 = 4`. I did 5 hrs to give it a little bit more buffer.

MrCreosote: no idea where that 300 came from, 4.6 is 5 * 60 / 65...

Ok, so you're assuming 4 batches per node, so 4 * 65 = 260mins, so a 40min buffer. Can you document that?

MrCreosote: Just somewhere in here document the math, e.g. we're using 128/256 cpus as that gave us the best run time w/ 4 simultaneous instances of 1000 genomes running for 65 mins (w/ buffer), we're assuming 4 sequential runs, etc. When the script is changed to specify all this stuff on a tool by tool basis we can move the docs to the specific tool settings

Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/task_generator.py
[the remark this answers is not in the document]
MrCreosote: Can you also mention the 256 vs. 128 core thing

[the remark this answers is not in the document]
MrCreosote: Other than that looks good enough for now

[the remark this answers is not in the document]
Tianhao-Gu: 👍

MrCreosote (commented):
good enough until you refactor to support per tool configs with all the params

MrCreosote (approved):

PR #221, opened 2023-03-30
RE2022-137: Add get selection export types endpoint

MrCreosote:


Tianhao-Gu (approved):
sry about the delay. But it LGTM.

PR #222, opened 2023-03-31
RE2022-137: Add method to save sets to workspace to workspace wrapper

MrCreosote:


src/service/workspace_wrapper.py
```diff
@@ -124,9 +152,58 @@ async def check_workspace_permissions(self, workspace_ids: set[int]) -> None:
...
+        if not sets:
+            raise ValueError("no sets")
+        objs = []
+        for s in sets:
+            if not s.upas:
+                raise ValueError("all sets must have at least one UPA")
+            setinfo = _TYPE_TO_SET_INFO.get(s.upa_type)
+            if not setinfo:
+                raise errors.IllegalParameterError(f"Unsupported workspace type: {s.upa_type}")
+            wsset = {"description": s.description} | setinfo["items"](s.upas)
```
Tianhao-Gu, line 194: just want to make sure when s.description is none, ws.save also works for both sets type right? `description` is optional for both sets type?

Tianhao-Gu (approved):
👍

src/service/workspace_wrapper.py
MrCreosote: It's not optional, but the workspace allows `null` values for any primitive field.

PR #237, opened 2023-04-05
RE2022-138: Return column info for microtraits

MrCreosote:


src/service/data_products/heatmap.py
```diff
@@ -93,9 +100,17 @@ async def _delete_selection(self, storage: ArangoStorage, internal_selection_id:
...
+        collection_id: str = PATH_VALIDATOR_COLLECTION_ID,
+        load_ver_override: str | None = QUERY_VALIDATOR_LOAD_VERSION_OVERRIDE,
+        user: kb_auth.KBaseUser = Depends(_OPT_AUTH)
+    ) -> heatmap_models.Columns:
+        appstate = app_state.get_app_state(r)
+        storage = appstate.arangostorage
+        load_ver = await get_load_version(
+            storage, collection_id, self._id, load_ver_override, user)
+        doc = await get_collection_singleton_from_db(
+            storage, self._colname_columns, collection_id, load_ver, bool(load_ver_override))
```
Tianhao-Gu, line 115: It's unlikely we need to worry about load_ver_override being the same as the current load_ver and it raises `NoDataFoundError` right?

Tianhao-Gu (approved):
👍

src/service/data_products/heatmap.py
MrCreosote: If `load_ver_override` is provided, the user must be an admin so I'm not super worried about whether they get a 400 instead of a 500 if there's no current data loaded. That situation shouldn't happen (we shouldn't be activating collections pointing at data that doesn't exit) anyway so this is a really edgy case. Unless you feel strongly about it I think it's ok as it.

PR #249, opened 2023-04-11
add microtrait in compute genome attri script

Tianhao-Gu:


src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -310,6 +321,42 @@ def checkm2(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
...
+                                                                os.path.basename(genome_file[0]),
+                                                                source_file_ext))
+            source_genome_file_map[genome_id] = genome_file[0]
+            fna_files.append(str(genome_file[0]))
+
+    start = time.time()
+
+    # RUN MicroTrait in serial
+    args_list = [(*genome_file, debug) for genome_file in list(source_genome_file_map.items())]
+    results = [_run_microtrait(*args) for args in args_list]
```
Tianhao-Gu, line 349: I was attempting to implement parallelization but found a few issues. 

- I wasn't able to use threads due to rpy2 threads are blocking each other. In my case it's even worse. My second thread won't be able to import `microtrait` due to GIL. (https://stackoverflow.com/questions/48952411/how-can-rpy2-r-code-from-python-be-run-in-a-thread-without-blocking-other-thre)
- Using process here is also problematic. In `main`, we already split each tool operation call into multiple process. We need to turn daemon off in order to support children processes within main parent processes. I think it's a bad idea to turn daemon off and since python 3.11, turning off daemon is no longer an option.   

I think it leaves us a few options, 

- Try to implement microtrait child processes with python 3.10
- Run microtrait in serial in compute_attri script. In the taskfarmer script, we made batch size for microtrait small (100 per batch) and run more taskfarmer tasks. 
- Try to implement parallelization with R. This option is the most efficient one in term of performance. But may add maintenance work in the future.

MrCreosote: My first thought is to put microtrait in `SERIES_TOOLS` and then run one `subprocess` calling the r script per genome id (possibly with a queue if necessary). Then the r script is isolated and you don't have to worry about r2py.

MrCreosote: I def don't want to be stuck on py3.10 forever

Pipfile
```diff
@@ -28,6 +28,7 @@ coverage = "==7.2.2"
 pytest-cov = "==4.0.0"
 ipython = "==8.12.0"
 pytest-asyncio = "==0.21.0"
+rpy2 = "==3.5.10"
```
MrCreosote, line 31: It sounds like this is going away in a later PR - can we get rid of it here so folks using main don't have to install it and then almost immediately uninstall it?

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -310,6 +321,49 @@ def checkm2(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
     return failed_ids
 
 
+def microtrait(genome_ids, work_dir, source_data_dir, debug, program_threads, node_id, source_file_ext):
+    failed_ids, size = list(), len(genome_ids)
+    print(f'Start executing MicroTrait for {len(genome_ids)} genomes')
```
MrCreosote, line 326:
```suggestion
    print(f'Start executing MicroTrait for {size} genomes')
```

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -310,6 +321,49 @@ def checkm2(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
...
+    batch_dir = _create_batch_dir(work_dir, 'series', size, node_id)
+    tool_genome_id_map, source_genome_file_map = dict(), dict()
+    # retrieve genomic.fna.gz files
+    fna_files = list()
+    for genome_id in genome_ids:
+        genome_file = _find_genome_file(genome_id, source_file_ext, source_data_dir,
+                                        exclude_file_name_substr=['cds_from', 'rna_from', 'ERR'],
+                                        expected_file_count=1)
+
+        if genome_file:
```
MrCreosote, line 337: Should an error be thrown if there's no genome file?

Pipfile
Tianhao-Gu: right. that's why I added to `dev-packages` so that tests can pass. Do you prefer to completely remove it and the import from comput_genome_attris script?

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -310,6 +321,49 @@ def checkm2(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
...
+    failed_ids, size = list(), len(genome_ids)
+    print(f'Start executing MicroTrait for {len(genome_ids)} genomes')
+
+    batch_dir = _create_batch_dir(work_dir, 'series', size, node_id)
+    tool_genome_id_map, source_genome_file_map = dict(), dict()
+    # retrieve genomic.fna.gz files
+    fna_files = list()
+    for genome_id in genome_ids:
+        genome_file = _find_genome_file(genome_id, source_file_ext, source_data_dir,
+                                        exclude_file_name_substr=['cds_from', 'rna_from', 'ERR'],
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: there are a lot of GTDB genomes that doesn't have genome files in NCBI. Do we really want to raise? In that case, none of GTDB collection will run.

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -310,6 +321,49 @@ def checkm2(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
...
+                                        expected_file_count=1)
+
+        if genome_file:
+            tool_genome_id_map.update(_map_tool_id_to_genome_id('microtrait',
+                                                                genome_id,
+                                                                os.path.basename(genome_file[0]),
+                                                                source_file_ext))
+            source_genome_file_map[genome_id] = genome_file[0]
+            fna_files.append(str(genome_file[0]))
```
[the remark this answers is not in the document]
Tianhao-Gu: They are some duplicate codes, but I really don't see how we can future abstract. Yes, all those 3 tools are fetching genome files, but gtdb-tk needs a master file that has all genome and genome files info, checkm2 needs to pass that file list to the commend line. Microtrait and checkm2 share some similarities. But it might get changed down the road. I can probably do a function that takes genome_ids and return a dict with genome_id and file_path. But then all those tool functions also need to go over the whole dict and process genome files to suit their needs.  Do you think that worth the refactor?

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: I was thinking the parser script should warm or generate a file for failed genomes. I put a todo item in each tool functions regarding `failed_ids`.

Pipfile
MrCreosote: yeah let's get rid of it unless there's a really compelling reason to keep it

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: I'm talking specifically about lines 332-343 above - I'm thinking you're talking about the whole function? AFAICT those lines are almost identical

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Maybe a CLI param then? For PMI etc it can fail, but for GTDB it just warns

MrCreosote: Or maybe just in the tool config at the start of the script

Pipfile
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍 refactored

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Thinking about this further - the list of IDs to process come from the directory names in the `<collection>/<load_ver>` directory or a genome IDs file which specifies a subset of those directory names. Why are we creating directories that are (I assume) empty, since there's no assembly data? Shouldn't we just not make those directories?

Pipfile
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: :+1: I'll add any comments about the refactor to a new review

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: Well maybe. But if we don't create directories for every genome. 1. There will be a lack of synchronization between the genomes in our system and the genomes defined in GTDB metadata. 2. Some genomes may have files for tool A, but not for tool B.

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -303,9 +342,36 @@ def checkm2(genome_ids, work_dir, source_data_dir, debug, program_threads, batch
...
+
+    # TODO process results from _run_microtrait
+
+    end_time = time.time()
+    print(
+        f'Used {round((end_time - start) / 60, 2)} minutes to execute MicroTrait for {size} genomes')
 
-    _create_genome_metadata_file(tool_genome_id_map, source_genome_file_map, len(genome_ids), batch_dir)
+    _create_genome_metadata_file(genomes_meta, batch_dir)
     # TODO: inspect stdout for failed ids or do it in the parser program
```
MrCreosote, line 375: To my mind this should be done here so the job is registered as a fail and can be restarted. Is this something that'll show up in a a PR or two? If not I'm thinking we should make a ticket, I don't want to lose track of this

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -69,15 +69,32 @@
...
+SERIES_TOOLS = ['microtrait']  # Tools cannot be executed in parallel
 
 # Fraction amount of system cores can be utilized
 # (i.e. 0.5 - program will use 50% of total processors,
 #       0.1 - program will use 10% of total processors)
 SYSTEM_UTILIZATION = 0.5
 
+# source genome files can be missing for those collections
+# genomes with missing files will be skipped rather than raising an error
+# TODO parser script should catch those missing genomes and report them or writes to a file for later use
```
MrCreosote, line 81: is this something that'll happen soon are you thinking?

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: For 1, there will be a lack of sync anyway because there's nothing to run the tools on anyway, right? So there will be nothing in the system for those genomes (unless we plan to merge data from the metadata file and the tool output, which right now the calc/parser scripts won't handle unless I'm mistaken).

For 2, are you thinking that we'll want to run some tools starting from data other than an assembly file?

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -187,42 +208,83 @@ def _run_gtdb_tk_classify_wf(batch_file_path, work_dir, debug, genome_ids, progr
         f'{len(genome_ids)} genomes')
 
 
-def _map_tool_id_to_genome_id(tool_name, original_genome_id, genome_file_name, source_file_ext):
+def _retrieve_tool_genome_identifier(tool_name, original_genome_id, genome_file_name=None, source_file_ext=None):
     """
-    Construct a dictionary that associates tool-generated genome IDs with the original genome IDs used to create
-    the _key field of a document.
+    Generate the unique genome identifier linked to the tool, which corresponds to the original genome ID.
```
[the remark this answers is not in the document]
Tianhao-Gu: GTDB-TK and microTrait (hopefully) provide us with the ability to assign a name to a genome identifier. However, Checkm2 lacks this functionality as it follows its own logic, which involves trimming the extension of the fasta file.

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -89,9 +106,13 @@ def _find_genome_file(genome_id, file_ext, source_data_dir, exclude_file_name_su
         genome_files = [f for f in genome_files if
                         all(name_substr not in f for name_substr in exclude_file_name_substr)]
```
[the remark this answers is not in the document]
Tianhao-Gu: sure. I think it will make things easier.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: we do have multiple tickets in the backlog regarding rerun incomplete genomes. I remember during one of the meetings, we agreed that it's not a pressing concern at the moment, as it would actually be an improvement since we can simply restart the entire task.
https://kbase-jira.atlassian.net/browse/RE2022-133
https://kbase-jira.atlassian.net/browse/RE2022-96

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: Based on the answer below will that not mess up checkm, which gets the ID from the file name?

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Right, but this is a little different - it's causing the job to fail so that it can be rerun. Those tickets are about rerunning failed jobs, this TODO is about making the job fail when it should

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: It should be addressed with the rerun incomplete genomes ticket. Definitely not before the microtrait work. But we can bump the priority if you think it's more important.

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
Tianhao-Gu: just `assembly.fa`? this might. I was thinking {genome_id}.fa.

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: What is this TODO for then?

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: Where `{genome_id}` is the name of the folder, e.g. `GB_GCA_016176815.1` or `3541_2_6`?

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: just to remind me that I need to fix those failed ids.

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
Tianhao-Gu: yea. that will work.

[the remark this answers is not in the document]
MrCreosote: Would that mean the genome_id and checkm genome_id would be the same, so we wouldn't need the tool -> genome_id conversion stuff?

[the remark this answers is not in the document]
Tianhao-Gu: yes. I think so. but only for those 3 tools that we currently have. Cannot say sure for future tools.

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: On phone call we decided to just not have empty folders. @Tianhao-Gu will make a ticket

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Per phone convo, these TODOs are no longer relevant, will be removed

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: If we remove the empty directories this will be moot, so can be removed

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: That sounds good to me

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: If we change the filename as above, this mapping will no longer be required
:+1:

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍 [src/loaders/genome_collection/compute_genome_attribs.py](https://kbase-jira.atlassian.net/browse/RE2022-147)

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
Tianhao-Gu: yup. mentioned this in one of ACs. https://kbase-jira.atlassian.net/browse/RE2022-147

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍 updated todo.

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -187,108 +207,124 @@ def _run_gtdb_tk_classify_wf(batch_file_path, work_dir, debug, genome_ids, progr
...
+            node_id, source_file_ext, kbase_collection, run_steps=False):
     # NOTE: Require defining the 'GTDBTK_DATA_PATH' environment variable
     #       e.g. export GTDBTK_DATA_PATH=/global/cfs/cdirs/kbase/collections/libraries/gtdb_tk/release207_v2
     #       https://ecogenomics.github.io/GTDBTk/installing/index.html#gtdb-tk-reference-data
     #
     #       Ensure that Third-party software are on the system path.
     #       https://ecogenomics.github.io/GTDBTk/installing/index.html#installing-third-party-software
 
-    failed_ids, size = list(), len(genome_ids)
+    size = list()
```
MrCreosote, line 294: this looks wrong

MrCreosote (changes requested):

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: There's another one like this below

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1:

MrCreosote: To explain for posterity, the failed_ids here referred to genome_ids without fasta files (e.g. empty directories). If a file fails to process otherwise, the job will fail per @Tianhao-Gu

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍 ah oh.

MrCreosote (approved):

PR #251, opened 2023-04-11
RE2022-143: Refactor genome attribs selection for general use

MrCreosote:
Can now be used by the heatmap code for selection and match marking

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ gtdb_attribs2.jsonl
 .DS_Store
 .vscode/settings.json
 collections_config.toml
+gtdb_attribs3.jsonl
+gtdb_taxa_count2_rank.jsonl
+gtdb_taxa_count2.jsonl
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -15,3 +15,6 @@ gtdb_attribs2.jsonl
 .DS_Store
 .vscode/settings.json
 collections_config.toml
+gtdb_attribs3.jsonl
+gtdb_taxa_count2_rank.jsonl
+gtdb_taxa_count2.jsonl
```
Tianhao-Gu, line 20: I am not aware any tests create those files. Just personal files for testing?

Tianhao-Gu (approved):
LGTM

.gitignore
MrCreosote: yeah. Usually I don't check in .gitignore for this reason but it looks like it got checked in at some point

PR #255, opened 2023-04-12
RE2022-143: Use internal IDs for managing match / sel processes

MrCreosote:
Makes it easier to reuse code between primary matches / selections & secondary m/ss, since secondary m/ss only keep the internal ID in the DB record (as they should)

Also infinitesimally safer since the process will only write to the m/s they were started against and will fail if that m/s is deleted rather than writing to a new m/s (which should never happen, but...)

Also also makes the `CollectionProcess` class a little saner since the internal ID can be provided to gtdb_lineage matcher right away rather than having to wait for the matcher to calculate it and provide it to the process `start() ` method later

Finally, move the match API parameter specifications into a shared location for reuse with heatmaps


Tianhao-Gu (approved):
LGTM

PR #259, opened 2023-04-14
RE2022-143: Add matching & selection to heatmaps

MrCreosote:


Tianhao-Gu (approved):
sry it got slipped away from my todo list. But it looks good.

PR #261, opened 2023-04-16
Add `status_only` flag to taxa_counts

### LINKED ISSUE #212: Add a `poll_only` option to the `taxa_count` data product
The `taxa_count` DP always returns partial data when polling for match or selection state, which is a (fairly small) waste of bandwidth. Add an option to only return DP process state with no partial data.

MrCreosote:
Save a little bandwidth

Closes #212


Tianhao-Gu (approved):
👍

PR #262, opened 2023-04-17
Start switch to `Annotated` param definitions

MrCreosote:
Just checking that it now works after the fastapi 0.95.1 release. From now on use the `Annotated` stycle and covert old style params over when touching them,


Tianhao-Gu (approved):
LGTM

PR #263, opened 2023-04-17
RE2022-148: Store missing IDs rather than fail on heatmap match / select

MrCreosote:
It's to be expected that not all data sources will have microtrait data, or the data required to calculated it, available.

Next step is to allow the client to retrive the missing IDs.


Tianhao-Gu (approved):
👍

PR #267, opened 2023-04-20
Refactor compute script to isolate dependencies pt1

MrCreosote:
Put any tool specific code, along with dependencies, in separate modules that are only run in the tool's own dockerfile. This means tool dependencies are isolated from each other and the server.

TODO:
* microtrait
* gtdb_tk
* remove old compute script and dockerfiles
* remove rpy2 from pipenv


src/loaders/compute_tools/tool_common.py
```diff
@@ -0,0 +1,458 @@
...
+                        f"Please ensure {DATA_ID_COLUMN_HEADER} column exists in the "
+                        + "data id file.")
+
+            if not set(all_data_ids) >= set(data_ids):
+                raise ValueError(
+                    f'Data files for {set(data_ids) - set(all_data_ids)} are not downloaded yet')
+        else:
+            # executing all data IDs
+            data_ids = all_data_ids
+        return data_ids
```
Tianhao-Gu, line 214: maybe we should return `set(data_ids)`? It might be possible if the data_id_file contains duplicate data_ids.

src/loaders/compute_tools/tool_common.py
```diff
@@ -0,0 +1,458 @@
...
+        pool.starmap(tool_callable, args)
+        pool.close()
+        pool.join()
+        prefix = "In total used" if total else "Used"
+        print(f"{prefix} {round((time.time() - start) / 60, 2)} minutes to "
+            + f"execute {self._tool} for {len(self._data_ids)} data units"
+        )
+
+
+def unpack_gz_file(gz_file: Path):
```
Tianhao-Gu, line 302: should this be moved to micrtrait.py or maybe some common module?

Tianhao-Gu (approved):
overall looks good. But have some questions.

src/loaders/compute_tools/tool_common.py
MrCreosote: It's in `tool_common`...?

I don't think it should go in microtrait, as it may be used elsewhere in the future. That being said we can move it to `loader_helper.py` or something if and when it's needed elsewhere

src/loaders/compute_tools/tool_common.py
MrCreosote: Sure, although `task_generator.py` is creating all the lists at the moment so it seems unlikely. I'll do it in a future PR.

PR #268, opened 2023-04-20
add data retrieval from microtrait

Tianhao-Gu:


src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -309,38 +330,70 @@ def _run_microtrait(genome_id, fna_file, batch_dir):
...
+    cell_info_df = cell_info_df.rename(columns={'microtrait_trait-displaynameshort': 'col_name',
+                                                'microtrait_trait-displaynamelong': 'col_description',
+                                                'microtrait_trait-value': 'cell_val'})
+
+    categories, cells, celid, min_value, max_value = list(), list(), 0, float('inf'), float('-inf')
+    for category_name, category_df in cell_info_df.groupby('category'):
+        columns = list()
+        for i, row in category_df.iterrows():
+            columns.append(ColumnInformation(id=str(i), name=row['col_name'], description=row['col_description']))
+            cell_val = float(row['cell_val'])
```
Tianhao-Gu, line 383: I am not sure what to do if `microtrait_trait-value` cannot be converted to float.

MrCreosote: Have you found any cases like this?

MrCreosote: Or do you have a reason to expect it'll happen?

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -70,6 +71,8 @@
 from rpy2 import robjects
 
 from src.loaders.common import loader_common_names
+from src.service.data_products.heatmap_common_models import (Cell, ColumnInformation, ColumnCategory, HeatMapMeta,
```
MrCreosote, line 74: If these are going to be shared (which I think it right) they should be moved form service to common

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -281,6 +284,24 @@ def _get_r_list_element(r_list, element_name):
     return r_list[pos]
 
 
+def _r_table_to_df(r_table):
```
MrCreosote, line 287: I'm not well versed enough in R or rpy2 to effectively review this, so I'll just have to assume it works

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -309,38 +330,70 @@ def _run_microtrait(genome_id, fna_file, batch_dir):
...
+        """
+        r_func = robjects.r(r_script)
+        r_result = r_func(fna_file, genome_dir)
+
+        microtrait_result = _get_r_list_element(r_result, 'microtrait_result')
+
+        trait_counts_atgranularity3 = _get_r_list_element(microtrait_result, 'trait_counts_atgranularity3')
+        trait_counts_atgranularity3_df = _r_table_to_df(trait_counts_atgranularity3)
+
+        trait_counts_atgranularity3_df.to_csv(os.path.join(genome_dir, 'trait_counts_atgranularity3.csv'), index=False)
```
MrCreosote, line 356: Can we make this so it's easy to switch granularity and the parser script Just Works regardless of the granularity? Something like

```suggestion
        trait_counts = _get_r_list_element(microtrait_result, 'trait_counts_atgranularity3')  # make this name a constant at top of file
        trait_counts = _r_table_to_df(trait_counts)

        trait_counts.to_csv(os.path.join(genome_dir, 'trait_counts.csv'), index=False)
```

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -309,38 +330,70 @@ def _run_microtrait(genome_id, fna_file, batch_dir):
...
+        categories.append(ColumnCategory(category=category_name, columns=columns))
+
+    meta = HeatMapMeta(categories=categories, min_value=min_value, max_value=max_value)
+    row = HeatMapRow(kbase_id=genome_id, cells=cells)
+
+    with open(os.path.join(genome_dir, 'heatmap_meta.json'), "w") as heatmap_meta:
+        json.dump(dict(meta), heatmap_meta)
+
+    with open(os.path.join(genome_dir, 'heatmap_row.json'), "w") as heatmap_row:
+        json.dump(dict(row), heatmap_row)
```
MrCreosote, line 396: Could you share sample contents for the meta and row files?

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -309,38 +330,70 @@ def _run_microtrait(genome_id, fna_file, batch_dir):
...
+            cells.append(Cell(celid=str(celid), colid=str(i), val=cell_val))
+            min_value, max_value = min(min_value, cell_val), max(max_value, cell_val)
+            celid += 1
+        categories.append(ColumnCategory(category=category_name, columns=columns))
+
+    meta = HeatMapMeta(categories=categories, min_value=min_value, max_value=max_value)
+    row = HeatMapRow(kbase_id=genome_id, cells=cells)
+
+    with open(os.path.join(genome_dir, 'heatmap_meta.json'), "w") as heatmap_meta:
+        json.dump(dict(meta), heatmap_meta)
```
MrCreosote, line 393: is it possible that the heat map meta might difffer from microtrait run to microtrait run (other than the min and max values)? If so, how will you deal with that?

src/loaders/genome_collection/compute_genome_attribs.py
```diff
@@ -309,38 +330,70 @@ def _run_microtrait(genome_id, fna_file, batch_dir):
...
+                                                'microtrait_trait-displaynamelong': 'col_description',
+                                                'microtrait_trait-value': 'cell_val'})
+
+    categories, cells, celid, min_value, max_value = list(), list(), 0, float('inf'), float('-inf')
+    for category_name, category_df in cell_info_df.groupby('category'):
+        columns = list()
+        for i, row in category_df.iterrows():
+            columns.append(ColumnInformation(id=str(i), name=row['col_name'], description=row['col_description']))
+            cell_val = float(row['cell_val'])
+            cells.append(Cell(celid=str(celid), colid=str(i), val=cell_val))
```
MrCreosote, line 384: Every cell in the entire heatmap has to have a unique ID. These IDs are only unique per genome

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Although maybe they should be used in the parser rather than here

MrCreosote (commented):
I'm not super well versed in pandas but I think I know what's going on here

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: I didn't notice any bad cases. I think it should be fine.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍 will move the parsing logic to the parsing script in next PR.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: will do that as a comment in the next PR.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: they will be different? within the same load_ver?

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍 I will handle this in the parser script.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: for `trait_counts_atgranularity` it seems to work fine.

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: I wouldn't sweat it then if there's no reason to expect it'll happen

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: What exactly do you mean? I think the vast majority of the logic should stay here, but maybe it just outputs csv or json without going through the models?

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Why not this PR? It's producing the output

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: I don't know, that's why I'm asking

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Ok, I see you already made a commit, if that stuff moves to the parser script that seems fine, as long as we're done with the R stuff. Easier to deal with the unique cell IDs if you handle it in the parser script as well

MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: So different request - can you share a csv output example from this code?

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: :+1: moot now with the removal of this code

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: I really don't think we need to worry about this.

MrCreosote: What I'm wondering about is whether if there's no trait found for a particular genome that column will be missing, that kind o fthing

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Sorry, I didn't mean for it to be a code comment, I just thought you'd post it here. Anyway fine as is.

MrCreosote: One question: what's the difference between `microtrait_trait-value` and `microtrait_trait-value1`?

MrCreosote:
> port over changes to new microtrait.py file

Thanks for doing that

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: I didn't even notice `microtrait_trait-value1`. Your guess is just as good as mine.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: hmmmm. the genome is missing? I think what microtrait does is on the gene level. Do you mean the complete `row` or `meta` is empty? Even that in the result json will still be something like {kbase_id: 'foo', load_ver: 'bar', ... , categories: [], cells:[]}.

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Should we remove it to avoid confusion / bugs?

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: No, a trait is missing. So different microtrait run outputs might have different or out of order trait lists

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Actually, is that column in the original output from microtrait? Maybe when merging the tables together in pandas if there's duplicate column names one of them is suffixed with `1`?

Tianhao-Gu: it comes with the original output. We should probably keep it. At least for this step, we are trying to reserve the output from the tool.

src/loaders/genome_collection/compute_genome_attribs.py
Tianhao-Gu: I know see what you are asking. I will handle it the parsing script and keep a global variable to check it.

MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_attribs.py
MrCreosote: Which table is it in? is it in the same table as `microtrait_trait-value`

Tianhao-Gu: yes. they both in the `trait_counts_atgranularity3` table.

MrCreosote: Weird, added it to doc containing questions for @ukaraoz

MrCreosote: :+1:

MrCreosote (approved):

PR #270, opened 2023-04-21
RE2022-152: Add gtdb_tk to ToolRunner setup and fix ID issue

MrCreosote:


src/loaders/compute_tools/tool_common.py
```diff
@@ -407,7 +424,9 @@ def _retrieve_data_file(
     #                  worry about multiple files matching the extension
     if data_file:
         data_file = Path(data_file)
-        if data_id_from_filename:
+        if random_ids:
+            tool_identifier = str(uuid.uuid4())
```
Tianhao-Gu, line 428: This is fine. But I was going to do something like `f'{data_id}_kbase'` in case we want to look at the result file manually.

Tianhao-Gu (approved):

src/loaders/compute_tools/tool_common.py
MrCreosote: done

Tianhao-Gu (approved):
👍

PR #272, opened 2023-04-21
RE2022-153: Add type information to heatmap columns

MrCreosote:


src/service/data_products/heatmap_common_models.py
```diff
@@ -72,10 +90,12 @@ class Cell(BaseModel):
         example="8",
         description="The ID of the column in which this cell is located."
     )
-    val: float = Field(
+    val: float | bool = Field(
         example=4.2,
         description="The value of the heatmap at this cell."
     )
+    class Config:
+        smart_union=True
```
Tianhao-Gu, line 98: I am not exactly sure what does that do.

src/service/data_products/heatmap_common_models.py
```diff
@@ -72,10 +90,12 @@ class Cell(BaseModel):
         example="8",
         description="The ID of the column in which this cell is located."
     )
-    val: float = Field(
+    val: float | bool = Field(
```
Tianhao-Gu, line 93: we probably also need `int` type here. Otherwise, the column type might be `int` and all cell values are automatically converted to float int.

src/service/data_products/heatmap_common_models.py
```diff
@@ -12,6 +13,19 @@
...
+class ColumnType(str, Enum):
+    """
+    The type of a column's values.
+    """
+    FLOAT = "float",
+    """ A float. """
+    INT = "int"
+    """ An integer. """
+    COUNT = "count"
+    """ Similar to an integer, but represents a numeric count. """
```
Tianhao-Gu, line 25: primary types make sense to me. But I am not sure about `count`. Do we need to build extra layers to check data quality? Because a `count` type column can have cells with float or boolean values.

Tianhao-Gu (dismissed):
looks good. But have some follow up questions.

src/service/data_products/heatmap_common_models.py
MrCreosote: It makes the union types (like `float | bool`) work as you expect. Otherwise they always get coerced to the first type in the union. I'm guessing it's a backwards compatibility thing

src/service/data_products/heatmap_common_models.py
MrCreosote: Count here is really just an int, but there's some descriptive info about the semantics of the int for the user. The microtrait type is `count` so I wanted to preserve the nomenclature for people familiar with microtrait.

How could you have a float when you're counting things? You don't count half a thing

src/service/data_products/heatmap_common_models.py
MrCreosote: done

src/service/data_products/heatmap_common_models.py
Tianhao-Gu: yea exactly, that's my question. We need to build additional checks to ensure `count` column only has integers. Currently, the cell value accepts bool, float and int.

src/service/data_products/heatmap_common_models.py
Tianhao-Gu: 👍

src/service/data_products/heatmap_common_models.py
Tianhao-Gu: 👍

src/service/data_products/heatmap_common_models.py
MrCreosote: Oh, you're talking about error checking. For `count` I'd just make sure it's an int > 0 in the loader, which it always was for the microtrait output I looked at

MrCreosote: The HeatMapRow model represents what the service can send to the user, so it has to represent all possible heatmap values, but the loader knows what the types are supposed to be so the more specific check should go there

Tianhao-Gu: 👍 got it.

Tianhao-Gu (approved):

PR #273, opened 2023-04-21
Simplify dependency pickling process

MrCreosote:
Now there's no reason not to import the spec registry anywhere needed since import loops are broken


Tianhao-Gu (approved):
👍

PR #275, opened 2023-04-21
Hide the arango aql object

MrCreosote:
Makes things easier to mock when we get around to writing tests, and makes the interface a bit more sane.


Tianhao-Gu (approved):
LGTM

PR #278, opened 2023-04-29
RE2022-158: Speed up heatmap endpoint by >10x

MrCreosote:
Not quite sure why returning the model in this case was killing performance, but doing the serialization manually is a lot faster

For url:
```
time curl http://localhost:5000/collections/PMI/data_products/microtrait/ > /dev/null
```

Before changes
```
real	0m2.908s
user	0m0.006s
sys	0m0.007s

```
After changes:
```
real	0m0.216s
user	0m0.009s
sys	0m0.024s
```


Tianhao-Gu (approved):
🥇

PR #279, opened 2023-05-01
RE2022-157: Add the source dir and metadata file name to the output metadata CSV (PR1)

Xiangs18:
PR1 for ticket https://kbase-jira.atlassian.net/jira/software/projects/RE2022/boards/78?selectedIssue=RE2022-157


src/loaders/compute_tools/tool_common.py
```diff
@@ -357,12 +358,15 @@ def _create_metadata_file(
...
+    genome_meta_file = loader_common_names.GENOME_METADATA_FILE
+    genome_meta_file_path = os.path.join(batch_dir, genome_meta_file)
     with open(genome_meta_file_path, "w") as meta_file:
         for genome_id, genome_meta_info in meta.items():
             meta_file.write(
                 f'{genome_meta_info[_META_TOOL_IDENTIFIER]}\t{genome_id}\t'
-                + f'{genome_meta_info[_META_SOURCE_FILE]}\n'
+                + f'{genome_meta_info[_META_SOURCE_FILE]}\t'
+                + f'{genome_meta_info[_META_SOURCE_DIR]}\t'
+                + f'{genome_meta_file}\n'
```
Tianhao-Gu, line 369: I think you misunderstood the task. It should include the metadata file created when assembly files were downloaded. i.e.  Acidovorax_sp_CF301.meta for Acidovorax_sp_CF301.fa. Then during the parsing step, this meta data can be used to retrieve the UPA.

Tianhao-Gu (changes requested):

src/loaders/compute_tools/tool_common.py
Xiangs18: Yeah, now it makes more sense after [RE2022-160](https://kbase-jira.atlassian.net/jira/software/projects/RE2022/boards/78?selectedIssue=RE2022-160)

Updated!!

Xiangs18:
Screen Shot 2023-05-10 at 4 43 07 PM

src/loaders/compute_tools/tool_common.py
```diff
@@ -134,7 +136,7 @@ def __init__(
         self._data_ids = self._get_data_ids()
 
         if not self._threads:
-            threads = max(int(multiprocessing.cpu_count() * min(_SYSTEM_UTILIZATION, 1)), 1)
+            self._threads = max(int(multiprocessing.cpu_count() * min(_SYSTEM_UTILIZATION, 1)), 1)
```
Tianhao-Gu, line 139: nice catch.

src/loaders/compute_tools/tool_common.py
```diff
@@ -362,7 +364,9 @@ def _create_metadata_file(
         for genome_id, genome_meta_info in meta.items():
```
Tianhao-Gu, line 364: we should probably consider adding a header row as more columns are added. But that can be future PRs when you refactor the parser script since adding the header will also require some changes for the parser script.

src/loaders/compute_tools/tool_common.py
```diff
@@ -399,7 +403,14 @@ def _prepare_tool(
...
         if data_file:
-            meta[data_id] = {_META_TOOL_IDENTIFIER: tool_identifier, _META_SOURCE_FILE: data_file}
+
+            mata_filename_path = os.path.join(data_file.parent, data_file.parent.name + ".meta")
+            meta_filename = os.path.basename(mata_filename_path) if os.path.exists(mata_filename_path) else ""
+
+            meta[data_id] = {_META_TOOL_IDENTIFIER: tool_identifier, 
+                             _META_SOURCE_FILE: data_file,
+                             _META_SOURCE_DIR: data_file.parent,
+                             _META_SOURCE_FILE_NAME: meta_filename}
```
Tianhao-Gu, line 413: so we are using full path for the source assembly file, but only the file name for the meta file. We should keep them consistent.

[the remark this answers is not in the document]
Xiangs18: switched to path for the meta datafile.

src/loaders/compute_tools/tool_common.py
Xiangs18: Good advice!! Header concern will be addressed in PR3 as described in [RE2022-157](https://kbase-jira.atlassian.net/jira/software/projects/RE2022/boards/78?selectedIssue=RE2022-157)

Tianhao-Gu (approved):
👍

src/loaders/compute_tools/tool_common.py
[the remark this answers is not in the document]
MrCreosote: In the ticket we said we were going to add the directory and change the file paths to just file names, but I guess just having the full path for every file is less bug-prone, so I've updated the ticket to say so

MrCreosote (approved):
LGTM

PR #290, opened 2023-05-03
RE2022-164: Add matcher set design doc

MrCreosote:


MrCreosote:
Closing this for now as it sounds like we won't need it for quite a while

PR #295, opened 2023-05-06
RE2022-160: Add collection parameters to workspace downloader

Xiangs18:
JIRA ticket: https://kbase-jira.atlassian.net/jira/software/projects/RE2022/boards/78?selectedIssue=RE2022-160

`PYTHONPATH=. python src/loaders/workspace_downloader/workspace_downloader.py --workspace_id 39795 --collection PMI --source_version 2023.1 --kb_base_url https://ci.kbase.us/services/ --keep_job_dir --token_filepath ~/.kbase_ci`

Screen Shot 2023-05-03 at 5 12 54 PM


src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -10,6 +10,10 @@
 required named arguments:
   --workspace_id WORKSPACE_ID
                         Workspace addressed by the permanent ID
+  --collection COLLECTION
+                        Create a collection and link in data to that colleciton from the overall workspace source data dir
+  --source_version SOURCE_VERSION
+                        Create a source version and link in data to that colleciton from the overall workspace source data dir
```
MrCreosote, line 16: These can be optional - if there's no collection given, don't make a collection. If one is required, both are required though

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -291,6 +350,9 @@ def main():
     output_dir = _make_output_dir(
         root_dir, loader_common_names.SOURCE_DATA_DIR, SOURCE, workspace_id
     )
+    collection_source_dir = _make_collection_source_dir(
+        root_dir, loader_common_names.SOURCE_DATA_DIR, collection, source_version
```
MrCreosote, line 354: If we want separate directories for "raw" data (`WS`, `NCBI`) and derived collections (`GTDB`, `PMI`) now is a good time to do it. `collections` makes the most sense to me, although that collides with the git repo, but maybe we could move that into a `git` directory.

@Tianhao-Gu what do you think?

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -275,8 +323,19 @@ def main():
...
+    (
+        workspace_id,
+        collection,
+        source_version,
+        root_dir,
+        kb_base_url,
+        workers,
+        token_filepath,
+        keep_job_dir,
+    ) = (
```
MrCreosote, line 335: It's be easier to read and less lines if it was just
```
workspace_id = args.workspace_id
collection = args.collection
...
```

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -238,6 +274,18 @@ def main():
         type=int,
         help="Workspace addressed by the permanent ID",
     )
+    required.add_argument(
+        "--collection",
+        required=True,
+        type=str,
+        help="Create a collection and link in data to that colleciton from the overall workspace source data dir",
```
MrCreosote, line 281:
```suggestion
        help="Create a collection and link in data to that collection from the overall workspace source data dir",
```

MrCreosote: same below

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -219,6 +244,17 @@ def process_input(conf):
...
+        # remove pre-exsit soft links and reset
+        csd_upa_dir = os.path.join(conf.csd, upa)
+        if os.path.isdir(csd_upa_dir):
+            shutil.rmtree(csd_upa_dir)
+        os.makedirs(csd_upa_dir)
+
+        # soft link .fa file and meta file
+        # from sourcedata/WS/<wsid>/<upa> to sourcedata/collection/soure_version/<upa>
+        os.symlink(dst, os.path.join(csd_upa_dir, f"{upa}.fa"))
+        os.symlink(metafile, os.path.join(csd_upa_dir, f"{upa}.meta"))
```
MrCreosote, line 256: Just softlink `csd_upa_dir` to `dstd`, so any other files we add to the source dir will automatically show up

MrCreosote (changes requested):

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: I think that will work. Given it's just softlink, we can probably use `collectionsdata/collection/load_ver/sourcedata`.

MrCreosote: So it's in the same directory as the batch directories?

Tianhao-Gu: yea. together with the tools directories. `sourcedata`, `checkm2`, `microtrait`, etc.

MrCreosote: Oh, I forgot about the tool dirs. Ok, that seems reasonable

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -219,6 +244,17 @@ def process_input(conf):
         with open(metafile, "w", encoding="utf8") as json_file:
             json.dump(_process_object_info(obj_info), json_file, indent=2)
 
+        # remove pre-exsit soft links and reset
+        csd_upa_dir = os.path.join(conf.csd, upa)
+        if os.path.isdir(csd_upa_dir):
+            shutil.rmtree(csd_upa_dir)
+        os.makedirs(csd_upa_dir)
```
Tianhao-Gu, line 251: I am not sure why we want to completely rebuild the csd_upa_dir if they already exist. Shouldn't we check existing link, if not matching, raise error?

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -302,7 +364,14 @@ def main():
...
         # set up conf and start callback server
-        conf = Conf(job_dir, output_dir, workers, kb_base_url, token_filepath)
+        conf = Conf(
+            job_dir,
+            output_dir,
+            collection_source_dir,
+            workers,
+            kb_base_url,
+            token_filepath,
+        )
```
MrCreosote, line 374: I forgot one thing - currently in the object loop below any files that have already been downloaded will be skipped. However, the downloader still needs to ensure they're linked from the collection

MrCreosote (changes requested):

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: Right

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -1,6 +1,6 @@
 """
```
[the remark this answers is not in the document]
MrCreosote: It already does that I think, if the meta file is there and has all the expected keys the script skips it

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: It does do that I think: https://github.com/kbase/collections/pull/295/files#diff-1ce3e6c7567701005d836ab7d2569c5938318aa586369a8c3224eb495406546aR378-R381

Although if the meta data file is missing or incorrect it just nukes the dir, which to me seems ok

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Based on https://github.com/kbase/collections/pull/295#discussion_r1187870747, it seems to me part 2 of downloader task is only optional. If that's the case, ensuring they're linked from the collection shouldn't be a mandatory check.

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -1,6 +1,6 @@
...
-usage: workspace_downloader.py [-h] --workspace_id WORKSPACE_ID [--root_dir ROOT_DIR] [--kb_base_url KB_BASE_URL]
-                               [--workers WORKERS] [--token_filepath TOKEN_FILEPATH] [--keep_job_dir]
+usage: workspace_downloader.py [-h] --workspace_id WORKSPACE_ID --collection COLLECTION --source_version SOURCE_VERSION [--root_dir ROOT_DIR]
```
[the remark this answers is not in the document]
Xiangs18: Yeap, we can have a common name for those vars across all scripts, but I am open to discussion on --load_ver. Should we use `source_ver` or `load_ver`? @MrCreosote @Tianhao-Gu

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Updated!

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Updated!

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Updated

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Updated

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Any ideas how to correctly soft link two dirs? I somehow cannot `cd` to dst dir after soft link. 

```
import os
src = "/Users/sean/Desktop/ouput_dir"
dst = "/Users/sean/Desktop/col_dir"
os.symlink(src, dst, target_is_directory=True)
```

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -219,6 +244,17 @@ def process_input(conf):
         with open(metafile, "w", encoding="utf8") as json_file:
             json.dump(_process_object_info(obj_info), json_file, indent=2)
 
+        # remove pre-exsit soft links and reset
```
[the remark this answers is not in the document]
Xiangs18: Updated

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -219,6 +244,17 @@ def process_input(conf):
...
+        csd_upa_dir = os.path.join(conf.csd, upa)
```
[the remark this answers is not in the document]
Xiangs18: Updated. conf.csd can be None.

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: It seems above code works in Linux env 👍

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
Xiangs18: I changed to load_ver for now. Please let me know if we need to switch.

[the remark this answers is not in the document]
Tianhao-Gu: per conversation at standup, we should use 'source_ver'. Sorry about that.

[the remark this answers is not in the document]
Xiangs18: I already switched back!!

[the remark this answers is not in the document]
MrCreosote: We should use the argument name from `loader_common_names` like the other scripts do thoug

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -270,19 +318,26 @@ def main():
     )
 
     args = parser.parse_args()
+
+    if (args.kbase_collection and not args.source_verion) or (
+        args.source_verion and not args.kbase_collection
+    ):
```
MrCreosote, line 324:
```suggestion
    if bool(args.kbase_collection) ^ bool(args.source_version):  # xor
```

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -291,6 +346,17 @@ def main():
...
+        _make_collection_source_dir(
+            root_dir,
+            loader_common_names.COLLECTION_DATA_DIR,
+            kbase_collection,
+            source_verion,
+            loader_common_names.SOURCE_DATA_DIR,
+        )
+        if (kbase_collection and source_verion)
+        else None
+    )
```
MrCreosote, line 359:
```suggestion
    collection_source_dir = None
    if kbase_collection:
        collection_source_dir = _make_collection_source_dir(
            root_dir,
            loader_common_names.COLLECTION_DATA_DIR,
            kbase_collection,
            source_verion,
            loader_common_names.SOURCE_DATA_DIR
        )
```

MrCreosote (changes requested):

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: I missed that this was the linked dir, not the source dir. You're right AFAICT, we shouldn't be calling rmdir on the link

Xiangs18: need to remove old csd_upa_dir to softlink csd_upa_dir and dstd, otherwise get file exist error.

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Add check_linking_status_for_existing_object logic!

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -22,7 +27,7 @@
 
             
 e.g.
-PYTHONPATH=. python src/loaders/workspace_downloader/workspace_downloader.py --workspace_id 39795 --kb_base_url https://ci.kbase.us/services/  --keep_job_dir
+PYTHONPATH=. python src/loaders/workspace_downloader/workspace_downloader.py --workspace_id 39795 --collection PMI --source_version 2023.1 --kb_base_url https://ci.kbase.us/services/ --keep_job_dir
```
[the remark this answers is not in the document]
Xiangs18: Good catch! Updated already

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -240,6 +278,16 @@ def main():
     )
 
     # Optional argument
+    optional.add_argument(
+        "--kbase_collection",
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Updated!

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Updated!

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: I would restructure your code a bit to make it more understandable / simple. I'd do things in 2 phases:

1) Download all the objects that need to be downloaded. This is a subset of the `list_objects` results
2) Link everything in the list_objects results if they don't already have a link. This happens after the queues have been `join()`ed.

MrCreosote: For example, currently if the file has been downloaded but isn't already linked, it gets downloaded again, which is not the behavior we want

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
Xiangs18: removed conf.csd from conf and add softlink check at end of script.

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: logic is updated. see below.

src/loaders/workspace_downloader/workspace_downloader.py
Xiangs18: Now do things in 2 phrases!

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -219,6 +247,16 @@ def process_input(conf):
         with open(metafile, "w", encoding="utf8") as json_file:
             json.dump(_process_object_info(obj_info), json_file, indent=2)
 
+        if conf.csd:
+            # remove pre-existing soft links and reset
+            csd_upa_dir = os.path.join(conf.csd, upa)
+            if os.path.isdir(csd_upa_dir):
+                shutil.rmtree(csd_upa_dir)
```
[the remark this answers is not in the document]
Xiangs18: logic is updated. see above!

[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

MrCreosote (changes requested):

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -322,6 +380,14 @@ def main():
         conf.pools.close()
         conf.pools.join()
 
+        # create a softlink from the relevant directory under collectionsdata
+        if collection_source_dir:
+            for upa in upas:
+                upa_dir = os.path.join(output_dir, upa)
+                csd_upa_dir = os.path.join(collection_source_dir, upa)
+                if not os.path.islink(csd_upa_dir):
```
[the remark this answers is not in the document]
Xiangs18: My current logic is still not correct.. Need to put into a function that return a bool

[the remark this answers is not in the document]
Xiangs18: Updated!

MrCreosote:
@Tianhao-Gu @Xiangs18 I think we need to change the directory structure again. Right now the hierarchy looks like this:

* `collectionsdata`
  * `<collection ID>`
    * `<source version OR load version>`
      *   `<tool dirs for load vesion OR sourcedata for source version>`

IOW we're mixing source version and load version in the same context and they're already east to confuse, as shown by all 3 of us confusing them on multiple occasions.

I think we need to do some thing else to clearly separate source versions and load versions - maybe put the source collections in `collectionssource` vs. `collectionsdata`? So 
```
.../kbase/collections/collectionssource/<collection ID>/<source version>/<links to UPA directories>
```

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
MrCreosote: Hmm, the current logic of deleting anything in the `collectionssource` directory if it's not the correct link makes me a bit nervous. This is not something that should normally happen since only this script should be making changes to the directory hierarchy, so I'm thinking if that there's something blocking the link operation the script should just throw an error and let the user resolve it.

Since the script now has "pick up where you left off" mechanics and, generally speaking, this shouldn't happen often if ever, that seems safer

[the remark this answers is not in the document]
MrCreosote: Also you might was well move the actual symlink operation into _create_softlink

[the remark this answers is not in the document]
Xiangs18: Updated. I think I am making it more complicated that it should be. May just below logic should be enough?

```
try:
    os.symlink(upa_dir, csd_upa_dir, target_is_directory=True)
except error as e:
    raise ValueError("two directories are not successfully linked") from e
```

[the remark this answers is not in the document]
MrCreosote: the current implementation still misses the case where csd_upa_dir is a file. I'd do something like this (psuedocode):

```
if src.exists():
    if src.is_link() and src.link() == target:
        return
    raise ValueError(f"{src} already exists and does not link to {target} as expected")
target.link(src)  # this is a race condition, but if src is created after the exists() call it'll just throw an error
```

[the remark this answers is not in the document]
MrCreosote: :+1:

MrCreosote (commented):
This LGTM. If @Tianhao-Gu approves so do I.

Tianhao-Gu (approved):
👍 let's go!

MrCreosote (approved):

PR #304, opened 2023-05-18
add check for directory count and object account

Xiangs18:


MrCreosote (approved):
This LGTM but @Tianhao-Gu should also give it a look

Tianhao-Gu (approved):

PR #305, opened 2023-05-19
RE2022-179: Move match / selection deletion under `admin`

MrCreosote:
Silly to have a different admin endpoint for both. Also lines up with the regular endpoint.


Tianhao-Gu (approved):
👍

PR #308, opened 2023-05-24
RE2022-179 Add endpoint to remove matches & selections

MrCreosote:
... for a collection version. This saves time when reloading data, as reloading leaves existing matches and selections in an inconsistent state. Putting them in a deleted state will allow the clean up routine to remove dead match / selection data eventually


Tianhao-Gu (approved):
👍

PR #314, opened 2023-06-02
mash parser

Tianhao-Gu:


src/loaders/compute_tools/mash/mash.py
```diff
@@ -17,21 +19,37 @@ def _run_mash_single(
...
-    # Save run info to a metadata file in the output directory for parsing later
-    metadata_file = output_dir / f'{genome_id}' / f'mash_run_metadata.json'
-    metadata = {'source_file': str(source_file), 'kmer_size': kmer_size, 'sketch_size': sketch_size}
-    with open(metadata_file, 'w') as f:
-        json.dump(metadata, f, indent=4)
+
+    remove_gz_file = False
+    if source_file.suffix == '.gz':
+        remove_gz_file = True
+        source_file = unpack_gz_file(source_file)
```
MrCreosote, line 28: I think we should build this stuff into the `ToolRunner.parallel_*` methods as it seems like a pretty common use case. Could just add an `unzip=False` default param to the methods

MrCreosote: Maybe do that in a different PR? If you want to do it here that's also fine

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -305,6 +308,38 @@ def _create_import_files(root_dir: str, file_name: str, docs: list[dict[str, Any
...
+    sketch_files = list()
+    for batch_dir in batch_dirs:
+        data_ids = [item for item in os.listdir(os.path.join(result_dir, batch_dir)) if
+                    os.path.isdir(os.path.join(result_dir, batch_dir, item))]
+        for data_id in data_ids:
+            data_dir = Path(result_dir, batch_dir, data_id)
+            with open(data_dir / loader_common_names.MASH_METADATA, 'r') as file:
+                metadata = json.load(file)
+            # Append '.msh' to the source file name to generate the sketch file name (default by Mash sketch)
+            sketch_file = metadata['source_file'] + '.msh'
```
MrCreosote, line 328: I'd say just add the sketch file location to the metadata file

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -305,6 +308,38 @@ def _create_import_files(root_dir: str, file_name: str, docs: list[dict[str, Any
...
+            if not os.path.exists(sketch_file):
+                raise ValueError(f'Unable to locate the sketch file: {sketch_file} for genome: {data_id}')
+            sketch_files.append(sketch_file)
+
+    import_dir = Path(root_dir, IMPORT_DIR)
+    os.makedirs(import_dir, exist_ok=True)
+    mash_output_prefix = import_dir / f'{kbase_collection}_{load_ver}_merged_sketch'
+
+    # run mash paste
+    command = ['mash', 'paste', mash_output_prefix] + sketch_files
```
MrCreosote, line 338: If there are 300k input files is this going to work? Does mash paste take a list of files like sketch does?

src/loaders/compute_tools/mash/mash.py
Tianhao-Gu: updated runner methods.

src/loaders/genome_collection/parse_tool_results.py

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: 👍  switch to use `-l file`.

src/loaders/compute_tools/mash/mash.py
MrCreosote: IIUC, the meta file will now contain the unzipped file as the source, but that unzipped file gets deleted? That seems wrong if so

src/loaders/genome_collection/parse_tool_results.py

src/loaders/genome_collection/parse_tool_results.py

src/loaders/compute_tools/mash/mash.py
Tianhao-Gu: right. But that is the true source file used to execute the tool. And if we are going to use source file as some identifier later, we need to know the file that the tool consumes. i.e. checkm2 uses source file name without extension as genome identifier.

MrCreosote: Ugh. I can see this causing issues in the future. Maybe we need `source` and `uncompressed_source` with the understanding that if `uncompressed_source` exists that was provided to the tool rather than `source`?

Tianhao-Gu: 👍

MrCreosote (approved):
LGTM once tests pass

PR #319, opened 2023-06-05
create seq meta

Tianhao-Gu:


src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -327,23 +347,32 @@ def _process_mash_tool(root_dir: str,
...
 
-        # run mash paste
-        command = ['mash', 'paste', mash_output_prefix, '-l', temp_file_path]
-        print(f'Running mash paste: {command}')
-        run_command(command)
+    # run mash paste
+    command = ['mash', 'paste', str(mash_output_prefix), '-l', temp_file_path]
+    print(f'Running mash paste: {" ".join(command)}')
+    run_command(command)
+    os.remove(temp_file_path)
```
MrCreosote, line 375: I'm confused here - the temp file should exist while mash is running with the old implementation, right? The original implementation looks right

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -127,6 +127,9 @@
 # Default directory name for the parsed JSONL files for arango import
 IMPORT_DIR = 'import_files'
 
+# The sequence metadata file name for Assembly Homology service (https://github.com/jgi-kbase/AssemblyHomologyService#sequence-metadata-file)
```
MrCreosote, line 130:
```suggestion
# The suffix for the sequence metadata file name for Assembly Homology service (https://github.com/jgi-kbase/AssemblyHomologyService#sequence-metadata-file)
```

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -327,23 +347,32 @@ def _process_mash_tool(root_dir: str,
                 metadata = json.load(file)
 
             sketch_file = metadata['sketch_file']
+            sketch_data = _read_sketch(sketch_file)
+            sketches = sketch_data['sketches']
+            if len(sketches) != 1:
+                raise ValueError(f'Expected only one sketch in the mash info output for genome: {data_id}')
+            sketch_id = sketches[0]['name']
+            seq_meta.append({'sourceid': data_id, 'id': sketch_id})
```
MrCreosote, line 355: is there a reason  you chose to get the file name from the sketch rather than the metdata file? I'm not saying this is the wrong way to go, just curious.

I suppose if we really wanted to be safe we could check that that sketch and metadata file names are the same

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -309,6 +312,23 @@ def _create_import_files(root_dir: str, file_name: str, docs: list[dict[str, Any
         convert_to_json(docs, f)
 
 
+def _read_sketch(sketch_file: Path) -> dict:
+    # run mash info on the sketch file and return the JSON output
+
+    temp_dir = Path(tempfile.mkdtemp())
```
MrCreosote, line 318: Will this get deleted?

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: I think before the file is closed, the file content is empty. Thus mash paste creates empty msh file.

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: 👍

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: added the check. yea I was thinking reading from the sketch file is safe. I also don't know if the tool is always going to use file name as ID. But we added the check if it errors we will know.

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: forget about it. added remove.

src/loaders/genome_collection/parse_tool_results.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

MrCreosote: But why is the file empty?

src/loaders/genome_collection/parse_tool_results.py

src/loaders/genome_collection/parse_tool_results.py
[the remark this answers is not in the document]
MrCreosote: I think you can move the try up a couple lines, even if the entire mash paste section needs to be inside the tempfile context block

src/loaders/genome_collection/parse_tool_results.py
MrCreosote: :+1:

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: `In [5]:         sketch_files = ['a', 'b', 'c']
   ...:         with tempfile.NamedTemporaryFile(mode='w', delete=False) as temp_file:
   ...:             temp_file.write('\n'.join(sketch_files))
   ...:             temp_file_path = temp_file.name
   ...: 
   ...:             with open(temp_file_path, 'r') as file:
   ...:                 content = file.read()
   ...:                 print(f'file content: {content}')
   ...: 
file content: 

`

Tianhao-Gu: Screenshot 2023-06-06 at 8 29 30 AM

MrCreosote: Maybe add `temp_file.flush()`?

MrCreosote: Surprised that's necessary but at least it works and is a lot simpler than the other solutions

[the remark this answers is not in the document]
MrCreosote: Moot now so :+1:

MrCreosote (approved):

PR #323, opened 2023-06-07
Fix engima

Xiangs18:
This PR address [this comment](https://kbase.slack.com/archives/C03FK9RKSBX/p1686075502548529) in Slack

### DIFF
--- a/src/loaders/common/loader_common_names.py
+++ b/src/loaders/common/loader_common_names.py
@@ -82,5 +82,5 @@ OBJECTS_NAME_GENOME = "KBaseGenomes.Genome"
 # The metadata file name created during the Mash run
 MASH_METADATA = 'mash_run_metadata.json'
 
-# The trait counts error file created if trait_counts.csv is not sucessfully generated
-TRAIT_COUNTS_ERROR = "trait_counts_error.txt"
+# The fatal error file created if trait_counts.csv is not sucessfully generated
+FATAL_ERROR_FILE = "fatal_error.json"
--- a/src/loaders/common/loader_common_names.py
+++ b/src/loaders/common/loader_common_names.py
@@ -81,3 +81,6 @@ OBJECTS_NAME_GENOME = "KBaseGenomes.Genome"
 
 # The metadata file name created during the Mash run
 MASH_METADATA = 'mash_run_metadata.json'
+
+# The trait counts error file created if trait_counts.csv is not sucessfully generated
+TRAIT_COUNTS_ERROR = "trait_counts_error.txt"
--- a/src/loaders/common/loader_common_names.py
+++ b/src/loaders/common/loader_common_names.py
@@ -82,5 +82,5 @@ OBJECTS_NAME_GENOME = "KBaseGenomes.Genome"
 # The metadata file name created during the Mash run
 MASH_METADATA = 'mash_run_metadata.json'
 
-# The fatal error file created if trait_counts.csv is not sucessfully generated
+# The fatal error file created if a data file cannot be successfully processed
 FATAL_ERROR_FILE = "fatal_error.json"
--- a/src/loaders/compute_tools/microtrait/microtrait.py
+++ b/src/loaders/compute_tools/microtrait/microtrait.py
@@ -19,8 +19,11 @@ _GENE_SCORE_COL = 'gene_score'  # column name from the genes_detected_table file
 
 def _get_r_list_element(r_list, element_name):
     # retrieve the element from the R list
+    exist = True
+    if element_name not in r_list.names:
+        return None, False
     pos = r_list.names.index(element_name)
-    return r_list[pos]
+    return r_list[pos], exist
 
 
 def _r_table_to_df(r_table):
@@ -65,13 +68,13 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
     r_func = robjects.r(r_script)
     r_result = r_func(str(fna_file), str(genome_dir))
 
-    microtrait_result = _get_r_list_element(r_result, 'microtrait_result')
+    microtrait_result, exist = _get_r_list_element(r_result, 'microtrait_result')
 
-    try:
-        trait_counts = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
-    except ValueError as e:
-        with open(os.path.join(genome_dir, loader_common_names.TRAIT_COUNTS_ERROR), 'w') as f:
-            f.write(e)
+    trait_counts, exist = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
+    if not exist:
+        fatal_dict = {fna_file: f"ValueError: Error running microtrait on {fna_file}"}
+        with open(os.path.join(genome_dir, loader_common_names.FATAL_ERROR_FILE), 'w') as outfile:
+            json.dump(fatal_dict, outfile)
         return 
     # example trait_counts_df from trait_counts_atgranularity3
     # microtrait_trait-name,microtrait_trait-value,microtrait_trait-displaynameshort,microtrait_trait-displaynamelong,microtrait_trait-strategy,microtrait_trait-type,microtrait_trait-granularity,microtrait_trait-version,microtrait_trait-displayorder,microtrait_trait-value1
@@ -88,7 +91,7 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
                                                 how='left')
         trait_counts_df.drop(columns=[loader_common_names.SYS_TRAIT_ID], inplace=True)
 
-        genes_detected_table = _get_r_list_element(microtrait_result, 'genes_detected_table')
+        genes_detected_table, exist = _get_r_list_element(microtrait_result, 'genes_detected_table')
         genes_detected_df = _r_table_to_df(genes_detected_table)
         detected_genes_score = dict(zip(genes_detected_df[_GENE_NAME_COL], genes_detected_df[_GENE_SCORE_COL]))
 
--- a/src/loaders/compute_tools/microtrait/microtrait.py
+++ b/src/loaders/compute_tools/microtrait/microtrait.py
@@ -69,8 +69,10 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
 
     try:
         trait_counts = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
-    except Exception as e:
-        return
+    except ValueError as e:
+        with open(os.path.join(genome_dir, loader_common_names.TRAIT_COUNTS_ERROR), 'w') as f:
+            f.write(e)
+        return 
     # example trait_counts_df from trait_counts_atgranularity3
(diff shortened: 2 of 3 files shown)


src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -67,7 +67,10 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
 
     microtrait_result = _get_r_list_element(r_result, 'microtrait_result')
 
-    trait_counts = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
+    try:
+        trait_counts = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
+    except Exception as e:
```
MrCreosote, line 72: This is too broad - it'll catch exceptions that should be thrown. What should happen is that there should be a test that checks whether the key exists or not, and if not, adds some sort of information to the output file (or an alternate file if that'll screw up the output file) that tells the parser (and anyone that reads the output file) that microtrait didn't work. The parser should then check that file and log that it's skipping that result.

IOW, there needs to be a positive identification that a known error mode occurred that the parser can key off of, and not just missing data.

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -406,6 +409,10 @@ def _process_heatmap_tools(heatmap_tools: set[str],
         _create_import_files(root_dir, rows_output, heatmap_rows_list)
         _create_import_files(root_dir, cell_details_output, heatmap_cell_details_list)
 
+        result_dir = _locate_dir(root_dir, kbase_collection, load_ver, tool='microtrait')
```
MrCreosote, line 412: The tool isn't necessarily microtrait, we could add more heatmap tools in the future

MrCreosote (changes requested):

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: Updated!

MrCreosote: A `ValueError` is still too broad - lots of things throw that. Instead of catching an error, test to see if the key is in the pandas columns set (maybe `index` returns -1 if it isn't? not sure) and don't catch errors at all - or throw a custom error from `_get_r_list_element` that is only thrown in this context and catch that

src/loaders/genome_collection/parse_tool_results.py

Tianhao-Gu (commented):
We will know the bad ids only after we run the parser. I would create the bad_id file within `ToolRunner.parallel_single_execution`, so that once the tool is finished we would know there are bad genomes.

MrCreosote:
> We will know the bad ids only after we run the parser. I would create the bad_id file within ToolRunner.parallel_single_execution, so that once the tool is finished we would know there are bad genomes.

Not understanding here - the microtrait specific code understands whether the ID is bad or not, so doesn't it have to write the file?

Tianhao-Gu:
> 

No it doesn't have to. I am saying, as current code stands, we only know the bad ids once we run the parser (tool runner just skips the bad genome). But tool runner can collect bad genomes and write to a file so that we know it immediately after execution.

MrCreosote:
I must be missing something - the tool runner writes out an error file for the ID here: https://github.com/kbase/collections/pull/323/files#diff-93fe7edc56038d1dac9d90291799622f2dd22ba247b451665c83c97e99c5da63R73-R74

So we do know which IDs are bad without having to run the parser right? Or are you asking for something else?

Xiangs18:
Two different things. One is for positive identification for each bad id, Tian is saying gathering all bad ids and write to a file.

MrCreosote:
How would that work given there could me many simultaneous `ToolRunner` instances running on different NERSC nodes and there's only one batch directory for all of them?

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -406,6 +409,10 @@ def _process_heatmap_tools(heatmap_tools: set[str],
         _create_import_files(root_dir, rows_output, heatmap_rows_list)
         _create_import_files(root_dir, cell_details_output, heatmap_cell_details_list)
 
+        result_dir = _locate_dir(root_dir, kbase_collection, load_ver, tool=tool)
+        with open(os.path.join(result_dir, BAD_DATA_IDS), "w") as outfile:
+            json.dump(bad_data_id, outfile)
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: return exist boolean and write to fatal_error.json if exist boolean is False

src/loaders/common/loader_common_names.py
```diff
@@ -80,4 +80,7 @@
 OBJECTS_NAME_GENOME = "KBaseGenomes.Genome"
 
 # The metadata file name created during the Mash run
-MASH_METADATA = 'mash_run_metadata.json'
\ No newline at end of file
+MASH_METADATA = 'mash_run_metadata.json'
+
+# The fatal error file created if trait_counts.csv is not sucessfully generated
```
MrCreosote, line 85:
```suggestion
# The fatal error file created if a data file cannot be not successfully processed
```

We should use this generally in other tools as well.

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -19,8 +19,11 @@
 
 def _get_r_list_element(r_list, element_name):
     # retrieve the element from the R list
+    exist = True
```
MrCreosote, line 22:
```suggestion
```

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -19,8 +19,11 @@
...
+    if element_name not in r_list.names:
+        return None, False
     pos = r_list.names.index(element_name)
-    return r_list[pos]
+    return r_list[pos], exist
```
MrCreosote, line 26:
```suggestion
    return r_list[pos], True
```

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -65,9 +68,14 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
     r_func = robjects.r(r_script)
     r_result = r_func(str(fna_file), str(genome_dir))
 
-    microtrait_result = _get_r_list_element(r_result, 'microtrait_result')
+    microtrait_result, exist = _get_r_list_element(r_result, 'microtrait_result')
```
MrCreosote, line 71:
```suggestion
    microtrait_result, _ = _get_r_list_element(r_result, 'microtrait_result')
```

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -83,7 +91,7 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
                                                 how='left')
         trait_counts_df.drop(columns=[loader_common_names.SYS_TRAIT_ID], inplace=True)
 
-        genes_detected_table = _get_r_list_element(microtrait_result, 'genes_detected_table')
+        genes_detected_table, exist = _get_r_list_element(microtrait_result, 'genes_detected_table')
```
MrCreosote, line 94:
```suggestion
        genes_detected_table, _ = _get_r_list_element(microtrait_result, 'genes_detected_table')
```

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -65,9 +68,14 @@ def _run_microtrait(genome_id: str, fna_file: Path, genome_dir: Path, debug: boo
     r_func = robjects.r(r_script)
     r_result = r_func(str(fna_file), str(genome_dir))
 
-    microtrait_result = _get_r_list_element(r_result, 'microtrait_result')
+    microtrait_result, exist = _get_r_list_element(r_result, 'microtrait_result')
 
-    trait_counts = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
+    trait_counts, exist = _get_r_list_element(microtrait_result, TRAIT_COUNTS_ATGRANULARITY)
+    if not exist:
+        fatal_dict = {"ValueError": f"ValueError: Error running microtrait on {fna_file}"}
```
MrCreosote, line 75:
```suggestion
        fatal_dict = {"error": "Microtrait output no data", "input_file": str(fna_file), "stacktrace": None}
```

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -395,7 +398,7 @@ def _process_heatmap_tools(heatmap_tools: set[str],
         except AttributeError as e:
             raise ValueError(f'Please implement parsing method for: [{tool}]') from e
 
-        heatmap_meta_dict, heatmap_rows_list, heatmap_cell_details_list = parse_ops(
+        heatmap_meta_dict, heatmap_rows_list, heatmap_cell_details_list, bad_data_id = parse_ops(
```
MrCreosote, line 401: I would change `bad_data_id` to `bad_data_ids` throughout. Seems minor but accurate naming can help with comprehension

MrCreosote (changes requested):

src/loaders/compute_tools/microtrait/microtrait.py
MrCreosote: :+1:

src/loaders/genome_collection/parse_tool_results.py
Xiangs18: Updated all bad_data_id to bad_data_ids!

src/loaders/common/loader_common_names.py
Xiangs18: Updated!

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: Updated!

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: Updated!

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: Updated!

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: Updated!

src/loaders/compute_tools/microtrait/microtrait.py
Xiangs18: Updated!

src/loaders/common/loader_common_names.py

src/loaders/compute_tools/microtrait/microtrait.py
MrCreosote: :+1:

src/loaders/compute_tools/microtrait/microtrait.py
MrCreosote: :+1:

src/loaders/genome_collection/parse_tool_results.py

MrCreosote (approved):

Tianhao-Gu (approved):
LGTM

PR #327, opened 2023-06-12
RE2022-195 - add retrieve samples logic

Tianhao-Gu:


src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -445,15 +522,8 @@ def main():
...
-            upa_dir = os.path.join(output_dir, upa)
-            if os.path.isdir(upa_dir) and loader_helper.is_upa_info_complete(upa_dir):
-                continue
-
-            # remove legacy upa_dir to avoid FileExistsError in hard link
-            if os.path.isdir(upa_dir):
-                shutil.rmtree(upa_dir)
             genome_upa = assembly_genome_map[upa.replace("_", "/")]
-            conf.queue.put([upa, obj_info, genome_upa])
+            conf.queue.put([upa, obj_info, genome_upa, retrieve_sample])
```
MrCreosote, line 526: Why not add retrieve_sample to `Conf` since it doesn't change per task?

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -278,31 +281,99 @@ def process_input(conf):
...
+            print("Completed %s" % (upa))
+        else:
+            print(f"Skip downloading {upa} as it already exists")
 
-        # cfn points to the assembly file outside of the container
-        # get_assembly_as_fasta writes the file to /kb/module/workdir/tmp/<filename> inside the container.
-        # workdir is shared between the container and the external file system
-        # Any file path get_assembly_as_fasta returns will be relative to inside the container, and so is not useful for this script
+        if retrieve_sample:
+            _download_sample_data(conf, upa)
```
MrCreosote, line 319: I'm thinking we should add a key to the meta file with the sample file as the value, and skip the sample value if it exists. Is there a reason not to do this?

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -278,31 +281,99 @@ def process_input(conf):
...
+    # there should only be one data link for each upa
+    if len(data_links) != 1:
+        raise ValueError(f"Expected 1 data link for {upa}, got {len(data_links)}")
+
+    # retrieve sample data and save to file
+    sample_id = data_links[0]['id']
+    sample_ret = _post_sample_service(conf.token,
+                                      conf.sample_url,
+                                      "get_sample",
+                                      {"id": sample_id})
```
MrCreosote, line 352:
```suggestion
                                      "get_sample_via_data",
                                      {"upa": upa, "id": sample_id, "version": data_links[0]["version"]})
```

No guarantee we'll have direct access to the sample

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -278,31 +281,99 @@ def process_input(conf):
...
+
+    with open(sample_file, "w", encoding="utf8") as json_file:
+        json.dump(sample_ret, json_file, indent=2)
```
MrCreosote, line 355: Before writing the sample needs to be flattened into a single level key-value map. We should error out if there's a key other than "value" in the sample field map (except for latitude and longitude, which also have "units") and put the latitude and longitude keys into the common names file for use in indexing and aggregations in the service.

We may need to switch lat/long to a single field like `"location": [<lat>, <long>]` but I'm not sure about that yet.

What I don't want to have happen is that samples for other collections have different fields and we silently ignore them, hence the errors

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -278,31 +281,99 @@ def process_input(conf):
...
 
-        print("Completed %s" % (upa))
+    # retrieve data links associated with upa
+    links_ret = _post_sample_service(conf.token,
+                                     conf.sample_url,
+                                     "get_data_links_from_data",
+                                     {"upa": upa.replace("_", "/")})
+    data_links = links_ret['links']
+    if not data_links:
+        print(f"No sample data links found for {upa}")
```
MrCreosote, line 340: Could you add a cli flag so an error is thrown here if there's no sample rather than ignored

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -278,31 +281,99 @@ def process_input(conf):
...
+    sample_ret = _post_sample_service(conf.token,
+                                      conf.sample_url,
+                                      "get_sample",
+                                      {"id": sample_id})
+
+    with open(sample_file, "w", encoding="utf8") as json_file:
+        json.dump(sample_ret, json_file, indent=2)
+
+
+def _post_sample_service(token, sample_url, method, params):
```
MrCreosote, line 358: Def use the sample service client

MrCreosote (changes requested):

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -52,6 +53,7 @@
 from multiprocessing import Pool, Queue, cpu_count
 
 import docker
+import requests
```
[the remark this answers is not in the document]
Tianhao-Gu: I firstly tried to use the client. But got error that I cannot figure out. 

`
self.ss = SampleService(self.sample_url, token=self.token)
links_ret = conf.ss.get_data_links_from_data({"upa": upa.replace("_", "/")})
`

`Traceback (most recent call last):
  File "/global/homes/t/tgu/miniconda3/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/global/homes/t/tgu/miniconda3/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/global/homes/t/tgu/miniconda3/lib/python3.11/multiprocessing/pool.py", line 109, in worker
    initializer(*initargs)
  File "/global/u2/t/tgu/collections/src/loaders/workspace_downloader/workspace_downloader.py", line 328, in process_input
    _download_sample_data(conf, upa, metafile)
  File "/global/u2/t/tgu/collections/src/loaders/workspace_downloader/workspace_downloader.py", line 350, in _download_sample_data
    links_ret = conf.ss.get_data_links_from_data({"upa": upa.replace("_", "/")})
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/u2/t/tgu/collections/src/clients/SampleServiceClient.py", line 815, in get_data_links_from_data
    return self._client.call_method('SampleService.get_data_links_from_data',
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/u2/t/tgu/collections/src/clients/baseclient.py", line 280, in call_method
    url = self._get_service_url(service_method, service_ver)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/u2/t/tgu/collections/src/clients/baseclient.py", line 207, in _get_service_url
    service_status_ret = self._call(
                         ^^^^^^^^^^^
  File "/global/u2/t/tgu/collections/src/clients/baseclient.py", line 187, in _call
    raise ServerError(**err['error'])
src.clients.baseclient.ServerError: Method not found: -32601. `

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: Do it here? I was thinking about do it in the parser.

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: I actually don't see any benefit using the client. We are calling the service on NERSC.

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
MrCreosote: use the client from this PR: https://github.com/kbase/sample_service/pull/476

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: To my mind, the downloaders should do any special data munging that depends on the source, and put it into a common format for the parser to handle. Keep the parser as simple as possible to my mind

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: I'm not sure I understand - the client has the same benefits wherever the calling code lives. The benefits here are that the client already has all the calling code written and we don't have to have duplicate code

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: The sample key constant probably needs to be in a common area so the parser can use it

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: Also means that we fail early if there's an unexpected sample format

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: okay. that's fine. Do you know if we already have this logic implemented somewhere?

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: I also have no idea how samples data will be used. Just parse something like below from `meta_controlled` and `meta_user` and save the dict to the sample file? If we change parsing logic down the road, we will just redownload samples from sample service and re-parse? 
`
{
'depth_bgs': 299,
'country': 'USA',
'sesar:size': 1
}
`

MrCreosote: yeah, that's what I was thinking. The sample data will be used in the exact same way the genome attribs data is used, except a different table, at least for the first cut.

MrCreosote: Let's not bother with `meta_user` though

MrCreosote: > If we change parsing logic down the road, we will need to redownload all samples from sample service and re-parse?

Good point - let's save the original and the mutated version. They're tiny so NDB

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍 added logic to retrieve node data.

src/loaders/common/loader_common_names.py
```diff
@@ -83,4 +83,11 @@
...
-FATAL_ERROR_FILE = "fatal_error.json"
\ No newline at end of file
+FATAL_ERROR_FILE = "fatal_error.json"
+
+# key name for sample file in the metadata file for downloaded workspace objects
+SAMPLE_FILE_KEY = "sample_file"
+
+# extension for source sample data file and sample node data file for downloaded workspace objects
+SAMPLE_FILE_EXT = "sample"
+SAMPLE_NODE_FILE_EXT = "node_tree.sample"
```
MrCreosote, line 93:
```suggestion
SAMPLE_PREPARED_EXT = "prepared.sample"
```

Or something like that, not really all that happy with "prepared". Just something that ways this version has be munged and is ready for the parser

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,119 @@ def process_input(conf):
...
+
+            print("Completed %s" % (upa))
+        else:
+            print(f"Skip downloading {upa} as it already exists")
+
+        if conf.retrieve_sample:
+            _download_sample_data(conf, upa, metafile)
+
+
+def _download_sample_data(conf: Conf, upa: str, metafile: str) -> None:
```
MrCreosote, line 333: This method is ~50 lines now, can it be split up?

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,119 @@ def process_input(conf):
...
+    sample_id = data_links[0]['id']
+    sample_ret = conf.ss.get_sample_via_data({"upa": upa.replace("_", "/"),
+                                              "id": sample_id,
+                                              "version": data_links[0]["version"]})
 
-        print("Completed %s" % (upa))
+    with open(sample_file, "w", encoding="utf8") as json_file:
+        json.dump(sample_ret, json_file, indent=2)
+
+    meta[loader_common_names.SAMPLE_FILE_KEY] = sample_file_name
```
MrCreosote, line 371: Can we add the munged sample file here as well? And probably best to write the metadata after both files are saved.

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,119 @@ def process_input(conf):
...
+        raise BadNodeTreeError(f"Expected 1 node in node tree, got {len(node_tree)}")
+
+    sample_node = node_tree[0]
+
+    meta_controlled = sample_node['meta_controlled']
+    for key, meta_value in meta_controlled.items():
+
+        if 'value' not in meta_value:
+            raise BadNodeTreeError(f"Expected 'value' key in meta_value, got {meta_value}")
+
```
MrCreosote, line 407: A couple other things:
* We should throw an error if there's anything other than "value" (or also units for the lat and long keys, in which case we should check the units are degrees). Otherwise we may be silently ignoring data.
* For the lat and long keys:
    * We need to ensure that all samples that have lat and long data have that data in the same keys
        * Non-GROW samples may have different key names for lat and long, we may have to expand this code later to change the key names
    * The samples data product will need to add indexes on those keys so the key names should be in the common names file

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: Actually, given :https://www.arangodb.com/docs/stable/indexing-geo.html

We should probably put the lat and long into an array.

MrCreosote: Oy... for now let's do both. We can hide the array from users but we'll have it for geo queries if needed

src/loaders/common/loader_common_names.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/common/loader_common_names.py
```diff
@@ -83,4 +83,18 @@
...
\ No newline at end of file
+FATAL_ERROR_FILE = "fatal_error.json"
+
+# key name for sample file and prepared sample file in the metadata file for downloaded workspace objects
+SAMPLE_FILE_KEY = "sample_file"
+SAMPLE_PREPARED_KEY = "sample_prepared_file"
+
+# extension for source sample data file and prepared sample node data file for downloaded workspace objects
+SAMPLE_FILE_EXT = "sample"
+SAMPLE_PREPARED_EXT = "prepared.sample"
```
MrCreosote, line 94: Can you briefly mention in these comments that the standard key is the sample as retrieved and prepared is formatted for ingestion by the parser?

src/loaders/common/loader_common_names.py
```diff
@@ -83,4 +83,18 @@
...
+
+# key name for latitude and longitude data from sample's meta_controlled
+SAMPLE_LATITUDE = "latitude"
+SAMPLE_LONGITUDE = "longitude"
+# key name for created sample geo-spatial data in format of [longitude, latitude]
+SAMPLE_GEO = 'geo_spatial'
```
MrCreosote, line 100:
```suggestion
SAMPLE_GEO = '_geo_spatial'
```

To indicate that this field shouldn't be returned from the API

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,173 @@ def process_input(conf):
...
+            # save meta file with relevant object_info
+            with open(metafile, "w", encoding="utf8") as json_file:
+                json.dump(_process_object_info(obj_info, genome_upa), json_file, indent=2)
+
+            print("Completed %s" % (upa))
+        else:
+            print(f"Skip downloading {upa} as it already exists")
+
+        if conf.retrieve_sample:
+            _download_sample_data(conf, genome_upa.replace("/", "_"), metafile)
```
MrCreosote, line 330: I would do the replace in the called method; make the arguments what people would naturally expect when they see `upa` in the args list

MrCreosote, line 330: Ugh, I realize there's no guarantee the samples will be connected to the genome. I think what needs to be done is 
* check both the assembly and genome for sample links
* if they both have links throw an error
* proceed with the one with links if any

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,173 @@ def process_input(conf):
...
+    if len(data_links) != 1:
+        raise ValueError(f"Expected 1 data link for {upa}, got {len(data_links)}")
+
+    # retrieve sample data and save to file
+    sample_id = data_links[0]['id']
+    sample_ret = conf.ss.get_sample_via_data({"upa": upa.replace("_", "/"),
+                                              "id": sample_id,
+                                              "version": data_links[0]["version"]})
+    if not sample_ret:
+        raise ValueError(f"Retrieved empty sample data for {upa}")
```
MrCreosote, line 409: This should be impossible unless something is very very wrong with the sample service

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,173 @@ def process_input(conf):
...
+        return
+
+    with open(sample_file, "w", encoding="utf8") as json_file:
+        json.dump(sample_ret, json_file, indent=2)
+
+    sample_prepared_name = f"{upa}.{loader_common_names.SAMPLE_PREPARED_EXT}"
+    sample_prepared_file = os.path.join(upa_dir, sample_prepared_name)
+    if _check_file_exists(loader_common_names.SAMPLE_PREPARED_KEY, meta, sample_prepared_file):
+        print(f"Skip generating sample node tree for {upa} as it already exists")
+        return
```
MrCreosote, line 359: Hmm, this means that somehow there was no sample file but there was a prepared file, which should be impossible and indicates something weird has happened. I think what I would do here is log the anomaly and overwrite the existing prepared file

MrCreosote: or throw an error, that'd also work

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: Hmm, it's still 40 lines, but it's reasonably readable... up to you if you want to try and further split things up

src/loaders/common/loader_common_names.py

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +295,173 @@ def process_input(conf):
...
+    sample_ret = _retrieve_sample(conf, upa)
+    if not sample_ret:
+        return
+
+    with open(sample_file, "w", encoding="utf8") as json_file:
+        json.dump(sample_ret, json_file, indent=2)
+
+    sample_prepared_name = f"{upa}.{loader_common_names.SAMPLE_PREPARED_EXT}"
+    sample_prepared_file = os.path.join(upa_dir, sample_prepared_name)
+    if _check_file_exists(loader_common_names.SAMPLE_PREPARED_KEY, meta, sample_prepared_file):
```
MrCreosote, line 357: There's another case that I missed - if the sample file is written but the script fails before the prepared file is written, then the prepared file will never be generated even if the tool is run again.

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: I didn't get what you mean. `upa` is in format of `a_b_c` and `genome_up` is in format of `a/b/c`.

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: hmmm did you mis-read `loader_common_names.SAMPLE_PREPARED_KEY`? `SAMPLE_FILE_KEY` <--> `SAMPLE_PREPARED_KEY`.

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: I'm saying just pass `genome_upa` as is (e.g. `a/b/c`) and do the `replace` where needed in `_download_sample_data`. 

Also I see that in a sub method of `_download_sample_data` the opposite `replace` occurs, so those could be eliminated.

`a/b/c` is the standard form of an UPA and so we should pass that when it's reasonable to do so IMO

Tianhao-Gu: right. I don't recall the history of `upa` and `genome_up`. But I think when we create file we want `a_b_c.xxx`. But when we write upa to a metadata file we want `a/b/c`. Thus there are a lot of `replace` here.

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: I don't think so. UUIC the code
* checks to see if there's a sample file. If there is, it `returns`
* creates the sample file
* checks to see if there's a prepared file
     * if there is, it returns, but that means that
         * the "raw" sample file didn't exist at the start of the method
         * the "prepared" sample file did exist, which should be impossible

Right? Or am I missing something?

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: probably need to clean up this. Because `upa` and `genome_up` in different format makes things unnecessary complicate.

MrCreosote: maybe
* pass everything as standard upa `a/b/c` and have a `to_filesafe_upa(upa)` method
* Make an UPA class where `str(new UPA(`a/b/c`)) == "a/b/c"`and `new UPA('a/b/c').filesafe() == `a_b_c`

src/loaders/common/loader_common_names.py
Tianhao-Gu: 👍

src/loaders/common/loader_common_names.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: I just pass UPA in 'a/b/c' format for this PR. Unifying upa format across this class is way too beyond the scope.

Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: you are right. updated logic should prevent this happening.

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍  should not happen with updated logic.

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: ok, would you mind making github issue to track it?

Tianhao-Gu: 👍 https://github.com/kbase/collections/issues/331

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -280,29 +299,192 @@ def process_input(conf):
...
+    return (file_key_name in metadata and
+            metadata[file_key_name] == Path(file_path).name and
+            os.path.isfile(file_path))
+
+
+def _retrieve_sample(conf: Conf, upa: str) -> dict[str, Any] | None:
+    # retrieve sample data from sample service
+
+    # retrieve data links associated with upa
+    links_ret = conf.ss.get_data_links_from_data({"upa": upa})
```
MrCreosote, line 418: sorry, I keep thinking of more complications - for reproducibility, can you add the `effective_time`from this call to the metadata? Since data links can expire or be re-linked we should store the time at which the links were retrieved so we can reproduce the data if needed

Tianhao-Gu: 👍

src/loaders/common/loader_common_names.py

src/loaders/common/loader_common_names.py

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: `time.time()` is going to have a different time format than the `effective_time` returned from the server - we should just use the server time

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: Hmm, thinking through how this should work here's what I would do:

* check the metadata file for the 3 sample keys. If they're both there, `return`.
* Pull the sample data
* Save both the files no matter what
* Open the metadata file again to try to avoid race conditions (we should do this in a DB later)
* Add the sample keys and save.

This would mean
* We don't pull the sample when we already have the data
* There's no chance of mixing data between 2 different samples if the sample for an object gets relinked

Tianhao-Gu: I pulled it every time because I think ws can link to different samples?

Tianhao-Gu: by the way, once we download samples the first time, how likely will we ever rerun it again?

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: > I pulled it every time because I think ws can link to different samples?

It's a possibility, but for our current use case, not very likely. If we wanted to check that I'd just add a force toggle for sample downloads

MrCreosote: > by the way, once we download samples the first time, how likely will we ever rerun it again?

For our current use case pretty unlikely I'd guess, but possibly if we find errors in the sample, that kind of thing

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: I mean literally take the value of the `effective_time` key from the return of the `get_data_links_from_data` call and put that in the metadata

Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: 👍

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
MrCreosote: :+1:

PR #329, opened 2023-06-13
RE2022-196 Deal with poor quality genomes in the data pipeline

Xiangs18:
This PR address [this comment](https://kbase.slack.com/archives/C03FK9RKSBX/p1686343775793539)

PR #333, opened 2023-06-19
RE2022-196: Bug fix in the data pipeline

Xiangs18:
This PR address the following

1. bug fixes for [RE2022-196](https://kbase-jira.atlassian.net/jira/software/projects/RE2022/boards/78?selectedIssue=RE2022-196)
2. add cases that have no classified attributes output from gtdb_tk tool into fatal error file. 

ENIGMA data now is up in CI ArangoDB!

[RE2022-196]: https://kbase-jira.atlassian.net/browse/RE2022-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -73,14 +74,23 @@ def _run_gtdb_tk(ids_to_files: Dict[Path, str], output_dir: Path, threads: int,
...
                                            summary_df[loader_common_names.GTDB_CLASSIFICATION_COL]):
+            genome_ids.add(genome_id)
             if classify_res.startswith(loader_common_names.GTDB_UNCLASSIFIED):
                 kbase_id = meta_dict[genome_id]
                 source_file_path = ids_to_files[genome_id]
                 error_message = f"GTDB_tk classification failed: {classify_res}"
                 fatal_tuples.append(FatalTuple(kbase_id, error_message, str(source_file_path), None))
     
+    miss_genome_ids = set(meta_dict.keys()) - genome_ids
+    if miss_genome_ids:
```
Tianhao-Gu, line 85: the `if` statement can be removed.

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -73,14 +74,23 @@ def _run_gtdb_tk(ids_to_files: Dict[Path, str], output_dir: Path, threads: int,
...
+        for miss_genome_id in miss_genome_ids:
+            kbase_id = meta_dict[miss_genome_id]
+            source_file_path = ids_to_files[miss_genome_id]
+            error_message = f"Missing classification attribute from {miss_genome_id}"
+            fatal_tuples.append(FatalTuple(kbase_id, error_message, str(source_file_path), None))
```
Tianhao-Gu, line 90: besides `error_message`, logic is duplicate.

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Updated!

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Updated!

Tianhao-Gu (dismissed):
LGTM

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -73,14 +73,20 @@ def _run_gtdb_tk(ids_to_files: Dict[Path, str], output_dir: Path, threads: int,
...
-                kbase_id = meta_dict[genome_id]
-                source_file_path = ids_to_files[genome_id]
                 error_message = f"GTDB_tk classification failed: {classify_res}"
-                fatal_tuples.append(FatalTuple(kbase_id, error_message, str(source_file_path), None))
+                fatal_tuple = create_gtdbtk_fatal_tuple(genome_id, meta_dict, ids_to_files, error_message, None)
+                fatal_tuples.append(fatal_tuple)
     
+    miss_genome_ids = set(meta_dict.keys()) - genome_ids
+    for miss_genome_id in miss_genome_ids:
+        error_message = f"Missing classification attribute from {miss_genome_id}"
```
MrCreosote, line 84: From reading the code it seems like there was no result for the genome at all, not just the classification attribute.

Is there anything in the GDTB_tk logs about missing data?

src/loaders/compute_tools/tool_common.py
```diff
@@ -269,6 +270,9 @@ def parallel_single_execution(self, tool_callable: Callable[[str, Path, Path, bo
                           meta[loader_common_names.META_SOURCE_FILE]),
                  output_dir,
                  self._debug))
+        
+        _create_metadata_file(genomes_meta, batch_dir)
```
MrCreosote, line 274: Can you explain why this needed to be moved from below? If there's an error in the execution stage that means the metadata file could exist without the tool output. I'm not sure what effect that might have.

src/loaders/compute_tools/tool_common.py
```diff
@@ -575,7 +579,7 @@ def write_fatal_tuples_to_dict(fatal_tuples: List[FatalTuple], output_dir: Path)
 
     fatal_error_path = os.path.join(output_dir, loader_common_names.FATAL_ERROR_FILE)
     with open(fatal_error_path, "w") as outfile:
-        outfile.dump(fatal_dict, outfile)
+        json.dump(fatal_dict, outfile)
```
MrCreosote, line 582: Should this have an indent? Otherwise I'd assume it's really hard to read

src/loaders/compute_tools/tool_common.py
```diff
@@ -584,6 +588,17 @@ def find_gtdbtk_summary_files(output_dir: Path):
     return summary_files
 
 
+def create_gtdbtk_fatal_tuple(genome_id: str, 
+                              meta_dict: Dict[str, str], 
+                              ids_to_files: Dict[str, Path], 
+                              error_message: str,
+                              stacktrace: str | None):
```
MrCreosote, line 595:
```suggestion
def create_gtdbtk_fatal_tuple(
    genome_id: str, 
    meta_dict: Dict[str, str], 
    ids_to_files: Dict[str, Path], 
    error_message: str,
    stacktrace: str | None
):
```

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -467,7 +467,7 @@ def _process_fatal_error_tools(check_fatal_error_tools: set[str],
     fatal_error_path = os.path.join(import_dir, fatal_output)
     print(f"Creating a merged {loader_common_names.FATAL_ERROR_FILE}: {fatal_error_path}")
     with open(fatal_error_path, "w") as outfile:
-        outfile.dump(fatal_dict, outfile)
+        json.dump(fatal_dict, outfile)
```
MrCreosote, line 470: Indent?

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -839,7 +839,7 @@ def gtdb_tk(root_dir, kbase_collection, load_ver, fatal_ids):
     result_dir = _locate_dir(root_dir, kbase_collection, load_ver, tool='gtdb_tk')
     batch_dirs = _get_batch_dirs(result_dir)
 
-    genome_id_col = loader_common_names.GENOME_ID_COL
+    genome_id_col = loader_common_names.GTDB_GENOME_ID_COL
```
MrCreosote, line 842: Hmm... some of these look like further evidence that we should start writing tests as we're getting regressions

src/loaders/compute_tools/tool_common.py
MrCreosote: Oh, maybe it's so the file exists when the tool tries to parse it. In that case I would
* Move the metadata file creation back to where it was
* Pass the `kbase_id` into the tools. My first thought would be to change the first argument to a list of named tuples with the `genome_id`, the `kbase_id`, and the file path

MrCreosote (changes requested):

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Right, there are basically 2 records that have no result at all from gtdb_tk tool.  

From what I see, I did not find anything in the GDTB_tk logs related to missing data.

src/loaders/compute_tools/tool_common.py
Xiangs18: https://github.com/kbase/collections/pull/329#discussion_r1228619440

We kinda agreed upon to get kbase_id from meta file. And meta file must exist before we can parse them. If that's the case, we need to move metafile create step before [multiprocessing self.excute step](https://github.com/kbase/collections/blob/0b4f64ee1583038d60978508d93222b58a620dde/src/loaders/compute_tools/tool_common.py#LL277C1-L277C1)

src/loaders/compute_tools/tool_common.py
Xiangs18: I think i do have an indent within context manager. You mean have a blank line between line 580 and 581?

src/loaders/compute_tools/tool_common.py
Xiangs18: Updated!

src/loaders/genome_collection/parse_tool_results.py
Xiangs18: Updated!

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: That seems very weird. Are there any docs about whether GTDB_tk should ever silently ignore genomes? I'm wondering if there's a bug in their or our code.

src/loaders/compute_tools/tool_common.py
MrCreosote: Well, that's the solution I naively proposed then without realizing that the metadata file didn't exist at that step. Since that's not the case, we should do something else rather than create a potential race condition; plus it makes the API more explicit if we pass in what we need rather than writing a file and having the code read it.

src/loaders/compute_tools/tool_common.py

src/loaders/genome_collection/parse_tool_results.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: It seems ppl had similar issues before with GTDB_tk: https://github.com/Ecogenomics/GTDBTk/issues/356

I `cat` every single file in the batch_dir and found these 2 records in `align/gtdbtk.ar53.filtered.tsv`. It seems they get filtered out because they are not identified as bacterial? GTDB_tk should have put them into *warnings.log or  *failed.genomes.tsv files, which makes more sense to me.

MrCreosote: Can you find any documentation about these files describing their expected contents?

Xiangs18: https://ecogenomics.github.io/GTDBTk/files/filtered.tsv.html#files-filtered-tsv

I am also looking at list of args [here](https://ecogenomics.github.io/GTDBTk/commands/classify_wf.html#commands-classify-wf) that can stop excluding genomes.

MrCreosote: Ok, it looks to me that we should get the genome IDs from `filtered.tsv` and `failed.genomes.tsv` (it'd be nice if they were both in the same file) and throw an error for any missing IDs that aren't in either of those lists

Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: Can you explain why there is a race condition here? A metafile is written to disk per batch_dir before any multiprocessing step is triggered though.

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Ah, all our gtdbtk.failed_genomes.tsv files are empty. I am not sure how gtdbtk.failed_genomes.tsv is formatted exactly. But based on their [doc](https://ecogenomics.github.io/GTDBTk/files/failed_genomes.tsv.html), its first col is not genome_id though. 

Have you seen any gtdbtk.failed_genomes.tsv with contents? @Tianhao-Gu

src/loaders/compute_tools/tool_common.py
MrCreosote: Because the metadata file signals to subsequent steps that the processing is done (even if currently the processing steps don't throw an exception if the metadata file is missing - they probably should if not). If a tool does a partial write the next step will see a metadata file and assume the tool completed successfully.

Regardless of that, there's zero reason not to just pass in the necessary data to the tool processing code explicitly rather than passing in some of it and writing some of it to a file. That's just confusing.

Xiangs18: genome_id: "69036_857_1_kbase" --> kbase_id: "69036_857_1"
Are we falling back to certain structure/pattern?  https://github.com/kbase/collections/blob/a1c4012e885b8a26ce1a9fd3ab159766014274a3/src/loaders/compute_tools/tool_common.py#L51

MrCreosote: Actually, it's not a race condition - it can't be since there's no concurrency here. I tricked myself due to the out of order operations. Not sure what it'd be called, ambiguous termination maybe?

Anyway, the reasons we should avoid it are the same

MrCreosote: Re the kbase ID - I don't actually remember why we're appending that suffix - @Tianhao-Gu?

But anyway, we should just treat the genome_Id as an opaque string and the mapping we create in `ToolRunner` (which is written to the metadata file) as the canonical way to translate `genome_id`<->`kbase_id`. That way if we want to change the suffix, need to MD5, etc. the `genome_id` in the future no other code changes will be needed

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: There's a couple things you could try:
1. Make a small test set containing some bad data (in `sourcedata/test` ?). Copy in a few good genomes, a couple of the filtered genomes, and copy in a couple of other genomes and edit the fasta file to truncate it so the sequences are short. That might prevent prodigal from calling genes. This is probably a helpful thing to do anyway for testing failure conditions.
2. Read the prodigal docs and see if they mention under what conditions it won't call genes, and see if you can create a fasta file with those conditions.
3. Ask in the #subject-matter-experts channel to see if anyone knows what would cause prodigal / gtdb_tk to fail and if they have any failing fasta files we could use

Xiangs18: I took the first approach and verified gtdbtk.failed_genomes.tsv's first col is indeed genome_id. 

Screen Shot 2023-06-22 at 2 51 06 PM

Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: I assume we are still taking the mapping approach for now. Anything else I need to fix for this conversation?

MrCreosote: Not sure what you mean by the mapping approach

Xiangs18: Anything else we need to do? Current codebase is already using the metadata file to map genome_id<->kbase_id.

MrCreosote: Yes, we should put the metadata file creation back to after processing is done and pass in the kbase_id to the tool method instead

Xiangs18: Sounds logical so that we don't need the mapping to do the conversion genome_id<->kbase_id, but why we are appending that suffix to begin with? @Tianhao-Gu

Tianhao-Gu: for GTDB genomes, GTDB-TK sometimes fails due to same reference genome id. That's why we append `_kbase` to the genome_id when run GTDB-TK tool.

Xiangs18: Hi, I put the metadata file creation back to after processing is done and passed in the genome_id<->kbase_id mapping into batch_execution and kbase_id into single_execution. Also tested in all tools for ENIGMA data.

Xiangs18: Updated!

MrCreosote: Rather than adding another argument just expand the first argument from a genome_id -> path mapping to either a list of named tuples or a mapping of genome_id -> NamedTuple[Path, kbase_id] or something like that.

Also, the single batch execution method should pass in the gneome_id as it did originally, but here we should add an argument for the kbase_id

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -18,7 +29,8 @@ def _run_checkm2(ids_to_files: Dict[str, Path], output_dir: Path, threads: int,
                '--output-directory', str(output_dir),
                '--threads', str(threads),
                '--force',  # will overwrite output directory contents
-               '--input'] + [str(v) for v in ids_to_files.values()]
+               '--input'] + [str(getattr(v, loader_common_names.META_SOURCE_FILE))
```
MrCreosote, line 32: You might as well just use the `.` syntax here, unless there's something I'm not understanding / remembering

MrCreosote: Same for more spots below

src/loaders/compute_tools/tool_common.py
```diff
@@ -231,11 +250,12 @@ def parallel_single_execution(self, tool_callable: Callable[[str, Path, Path, bo
                   each individual genome directory. Parser program will parse the result file in each individual genome
                   directory.
 
-        tool_callable - the callable for the tool that takes four arguments:
-            * The data ID
+        tool_callable - the callable for the tool that takes 4 arguments:
```
MrCreosote, line 253:
```suggestion
        tool_callable - the callable for the tool that takes 5 arguments:
```

src/loaders/compute_tools/tool_common.py
```diff
@@ -231,11 +250,12 @@ def parallel_single_execution(self, tool_callable: Callable[[str, Path, Path, bo
...
+            * The tool identifier
```
MrCreosote, line 254: This is always confusing to me - does this mean like `gtdb_tk` or the ID used to identify a genome to the tool? I know it's the latter but someone new to the code would be confused I think

src/loaders/compute_tools/tool_common.py
```diff
@@ -218,7 +237,7 @@ def _get_data_ids(self):
             data_ids = all_data_ids
         return list(set(data_ids))
 
-    def parallel_single_execution(self, tool_callable: Callable[[str, Path, Path, bool], None], unzip=False):
+    def parallel_single_execution(self, tool_callable: Callable[[str, Path, Path, bool, str], None], unzip=False):
```
MrCreosote, line 240: Putting the data ID at the end after debug seems weird. I'd put it right next to the genome_id / tool identifier

src/loaders/compute_tools/tool_common.py
```diff
@@ -292,7 +314,7 @@ def parallel_batch_execution(self, tool_callable: Callable[[Dict[str, Path], Pat
                   Batching genomes for gtdb_tk execution improves overall throughput.
 
         tool_callable - the callable for the tool that takes 4 arguments:
-            * A dictionary of the data_id to the source file path
+            * A dictionary of the tool identifier to the GenomeTuple
```
MrCreosote, line 317: Same thing re tool identifier

src/loaders/compute_tools/tool_common.py
```diff
@@ -584,6 +595,46 @@ def find_gtdbtk_summary_files(output_dir: Path):
     return summary_files
 
 
+def get_filtered_or_failed_genome_ids(output_dir: Path):
```
MrCreosote, line 598: This looks specific to gtdbtk, and so should live in the gtdb specific code

src/loaders/compute_tools/tool_common.py
```diff
@@ -584,6 +595,46 @@ def find_gtdbtk_summary_files(output_dir: Path):
...
+    genome_ids = list()
+
+    # process filtered.tsv files
+    align_dir = output_dir / "align"
+    filter_files = [file_name for file_name in os.listdir(align_dir) if 
+                    re.search(loader_common_names.GTDB_FILTER_FILE_PATTERN, file_name)]
+    if not filter_files or len(filter_files) > 2:
+        raise ValueError("Parsed unexpected filterd.tsv files {filter_files}.")
```
MrCreosote, line 606:
```suggestion
        raise ValueError("Parsed unexpected filtered.tsv files {filter_files}.")
```

MrCreosote, line 606: Can you be more specific about why the files were unexpected in the error message?

src/loaders/compute_tools/tool_common.py
```diff
@@ -584,6 +595,46 @@ def find_gtdbtk_summary_files(output_dir: Path):
...
+    return set(genome_ids)
+
+
+def _get_genome_ids_from_tsv_file(file_path: str):
+    with open(file_path, "r") as f:
+        res = [line.strip().split("\t")[0] for line in f]
+    return res
+
+
+def create_gtdbtk_fatal_tuple(
```
MrCreosote, line 625: This doesn't seem gtdbtk specific, looks like it could work for any of the batch methods

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -73,14 +72,30 @@ def _run_gtdb_tk(ids_to_files: Dict[Path, str], output_dir: Path, threads: int,
...
+
+    if error_genome_ids:
+        raise ValueError(
+            f"Miss {error_genome_ids} that are not found in either "
+            f"{loader_common_names.GTDB_FILTER_FILE_PATTERN} or "
+            f"{loader_common_names.GTDB_FAIL_GENOME_FILE}"
+        )
     
+    for miss_genome_id in miss_genome_ids:
+        error_message = f"No result for the genome {miss_genome_id}"
```
MrCreosote, line 93: Why not include the error message from the filtered or failed output files from GTDB_tk?

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -29,8 +35,8 @@ def _run_gtdb_tk(ids_to_files: Dict[Path, str], output_dir: Path, threads: int,
     # tab separated in 2 columns (FASTA file, genome ID)
     batch_file_path = output_dir / f'genome.fasta.list'
     with open(batch_file_path, "w") as batch_file:
-        for genome_id, source_file in ids_to_files.items():
-            batch_file.write(f'{source_file}\t{genome_id}\n')
+        for genome_id, genome_tuple in ids_to_files.items():
+            batch_file.write(f'{getattr(genome_tuple, loader_common_names.META_SOURCE_FILE)}\t{genome_id}\n')
```
MrCreosote, line 39: `.` syntax unless there's a reason I'm missing

src/loaders/compute_tools/tool_common.py
```diff
@@ -584,6 +595,46 @@ def find_gtdbtk_summary_files(output_dir: Path):
...
+        genome_id: str,
+        ids_to_files: Dict[str, GenomeTuple],
+        error_message: str,
+        stacktrace: str = None,
+):
+    genome_tuple = ids_to_files[genome_id]
+    kbase_id = getattr(genome_tuple, loader_common_names.META_DATA_ID)
+    source_file_path = getattr(genome_tuple, loader_common_names.META_SOURCE_FILE)
```
MrCreosote, line 633: `.` syntax

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
Xiangs18: Please note v is a NameTuple here. 

.syntax will result in AttributeError and we have discussed about this in the past PR.

src/loaders/compute_tools/tool_common.py
Xiangs18: I will use genome_id instead of tool identifier. It's less confusing to me as well. 

Also we kinda use data ID and kbase ID interchangeably. Any preference here?

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: That was a specific case IIRC. Declaring the tuple names with constants from loader_common_names probably isn't necessary for this case. You can just do something like

```
GenomeTuple = namedtuple("GenomeTuple", ["file", "data_id"])
```

...and then access like `genome_tuple.file`

MrCreosote: I remember doing it with getattr for the fatal tuple but I don't remember why we did it that way, other than the original syntax of `fatal_tuple.loader_common_names.ERROR` or whatever wouldn't work

src/loaders/compute_tools/tool_common.py
MrCreosote: For data ID, let's stick with data ID for now as I suspect we'll want to change the `kbase_id` in the final output to something other than the UPA at some point and it'd be good to have different names.

MrCreosote: I'm not really a fan of `genome_id` either because it doesn't really describe the purpose of the ID and why there's a data_id and a genome_id that both refer to the same thing? 

Maybe `tool_safe_data_id`? That's a bit long but more understandable

src/loaders/compute_tools/checkm2/checkm2.py
Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: Moved to gtdb_tk.py file

src/loaders/compute_tools/tool_common.py
Xiangs18: Updated!

Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: updated function name to create_fatal_tuple()

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Updated!

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
Xiangs18: Updated!

src/loaders/compute_tools/checkm2/checkm2.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: :+1:

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -54,33 +90,42 @@ def _run_gtdb_tk(ids_to_files: Dict[Path, str], output_dir: Path, threads: int,
...
+                fatal_tuple = create_fatal_tuple(tool_safe_data_id, ids_to_files, error_message)
+                fatal_tuples.append(fatal_tuple)
     
+    miss_tool_safe_data_ids = set(ids_to_files.keys()) - tool_safe_data_ids
+    filtered_or_failed_genome_id_mapping = _get_id_and_error_message_mapping_from_tsv_files(output_dir)
+    error_tool_safe_data_ids = miss_tool_safe_data_ids - set(filtered_or_failed_genome_id_mapping.keys())
+
+    if error_tool_safe_data_ids:
+        raise ValueError(
+            f"Miss {error_tool_safe_data_ids} that are not found in either "
```
MrCreosote, line 117:
```suggestion
            f"Missing IDs {error_tool_safe_data_ids} that are not found in either "
```

MrCreosote (commented):
Just about there I think

src/loaders/compute_tools/tool_common.py
```diff
@@ -49,6 +50,18 @@
...
 
+FatalTuple = namedtuple(
+    "FatalTuple",
+    [
+        loader_common_names.FATAL_ID,
+        loader_common_names.FATAL_ERROR,
+        loader_common_names.FATAL_FILE,
+        loader_common_names.FATAL_STACKTRACE,
+    ]
+)
```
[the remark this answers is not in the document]
Xiangs18: I prefer .syntax over getattr as well. Let's switch to .syntax then.

src/loaders/common/loader_common_names.py
```diff
@@ -88,6 +88,8 @@
 GTDB_CLASSIFICATION_COL = "classification"
 GTDB_UNCLASSIFIED = "Unclassified"
 GTDB_SUMMARY_FILE_PATTERN = "gtdbtk.*.summary.tsv"
+GTDB_FILTER_FILE_PATTERN = "gtdbtk.*.filtered.tsv"
+GTDB_FAIL_GENOME_FILE = "gtdbtk.failed_genomes.tsv"
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -12,15 +12,51 @@
...
+    genome_dict = dict()
+
+    # process filtered.tsv files
+    align_dir = output_dir / "align"
+    filter_files = [file_name for file_name in os.listdir(align_dir) if 
+                    re.search(loader_common_names.GTDB_FILTER_FILE_PATTERN, file_name)]
+    
+    if not filter_files or len(filter_files) > 2:
+        raise ValueError(f"Either or both gtdbtk.ar53.filtered.tsv and "
+                         f"gtdbtk.bac120.filtered.tsv files must exist.")
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/compute_tools/tool_common.py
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Xiangs18: Updated!

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py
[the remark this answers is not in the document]

src/loaders/common/loader_common_names.py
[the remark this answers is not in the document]

MrCreosote (approved):

Tianhao-Gu (approved):
👍

PR #336, opened 2023-06-23
Move shared names to common names

MrCreosote:
Some of the loader names are used or will need to be used in the service, so moved them to the common names file.


MrCreosote:
@Tianhao-Gu would you mind checking that the sample downloader still works? I just ran the built in tests

Tianhao-Gu (approved):
👍

PR #337, opened 2023-06-23
RE2022-203: Factor out genome attribs models for reuse

MrCreosote:
Also update shared query specfiers to the new `Annotated` syntax before reusing them with samples


Tianhao-Gu (approved):
👍

PR #339, opened 2023-06-26
RE2022-23: Reuse genome attribs query validators for samples

MrCreosote:
... and the special match and selection fields


Tianhao-Gu (approved):
LGTM

PR #349, opened 2023-06-28
RE2022-203: Move subset processing method from heatmap for reuse

MrCreosote:
Move the method that determines the applicable load version and either gets or starts the match and selection processes to common module for reuse with samples.

Also just put the match and selection prefixes in one place for now

### DIFF
--- a/src/service/data_products/table_models.py
+++ b/src/service/data_products/table_models.py
@@ -13,7 +13,7 @@ class AttributeName(BaseModel):
     )
 
 
-class TableAttributes(BaseModel, extra=Extra.allow):
+class TableAttributes(BaseModel):
     """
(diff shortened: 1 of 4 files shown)


src/service/data_products/table_models.py
```diff
@@ -13,7 +13,7 @@ class AttributeName(BaseModel):
     )
 
 
-class TableAttributes(BaseModel, extra=Extra.allow):
+class TableAttributes(BaseModel):
```
MrCreosote, line 16: This seems to be ancient cruft that's not currently needed

Tianhao-Gu (approved):
LGTM. Sry it took long. Didn't notice I have a pending review request.

PR #352, opened 2023-06-29
RE2022-189: Refactor NCBI downloader

Xiangs18:


src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,74 +1,76 @@
...
-  --source SOURCE       Source of data (default: GTDB)
-  --threads THREADS     Number of threads. (default: half of system cpu count)
-  --overwrite           Overwrite existing files.
+  --kbase_collection {GTDB}
+                        Create a collection and link in data to that collection from the overall source data dir (default: GTDB)
+  --root_dir ROOT_DIR   Root directory (default: /global/cfs/cdirs/kbase/collections)
+  --source {NCBI}       Source of data (default: NCBI)
+  --threads THREADS     Number of threads
+  --overwrite           Overwrite existing files
   --exclude_name_substring EXCLUDE_NAME_SUBSTRING [EXCLUDE_NAME_SUBSTRING ...]
```
Tianhao-Gu, line 24: I think we should make `['cds_from', 'rna_from', 'ERR']` a default `exclude_name_substring` for GTDB collection.

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,74 +1,76 @@
...
+                          [--kbase_collection {GTDB}] [--root_dir ROOT_DIR] [--source {NCBI}] [--threads THREADS] [--overwrite]
                           [--exclude_name_substring EXCLUDE_NAME_SUBSTRING [EXCLUDE_NAME_SUBSTRING ...]]
 
+PROTOTYPE - Download genome files from NCBI FTP server.
+
 options:
   -h, --help            show this help message and exit
 
 required named arguments:
   --download_file_ext DOWNLOAD_FILE_EXT [DOWNLOAD_FILE_EXT ...]
```
Tianhao-Gu, line 12: we should probably make `genomic.fna.gz` default extension. I think `xxx.genomic.fna.gz` the only file we want to download from NCBI.

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: in NCBI, I have seen `xxx. cds_from_genomic.fna.gz`, `xxx.rna_from_genomic.fna.gz` and `xxx.ERR_genomic.fna.gz`. Those are the files I think we should by default exclude.

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,74 +1,76 @@
...
-                        Download only files that match given extensions.
-  --release_ver RELEASE_VER
+                        Download only files that match given extensions
+  --release_ver {202,207,214,80,83,86,89,95}
```
Tianhao-Gu, line 14: I assume this is just from the `-h` output? But `release_ver` choices are not hard coded. Instead it's parsed dynamically from the NCBI website every time this script is called. We should probably add a comment here.

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -231,9 +288,26 @@ def download_genome_files(gene_ids: list[str], target_file_ext: list[str], exclu
     return failed_ids
 
 
+def create_softlink_between_csd_and_work_dir(
```
Tianhao-Gu, line 291: besides genome files, I think we also need to link the taxonomy file downloaded from `_fetch_gtdb_genome_ids`. Those files are used by taxa count parser. Unless we want to reference to the sourcedata directory when we run taxa count parser.

src/loaders/ncbi_downloader/ncbi_downloader.py
Tianhao-Gu: Just found out we have something defined here. Maybe just move it to a common module. 

https://github.com/kbase/collections/blob/main/src/loaders/compute_tools/tool_common.py#L48

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -231,9 +288,26 @@ def download_genome_files(gene_ids: list[str], target_file_ext: list[str], exclu
...
+        csd: str,
+        work_dir: str,
+        genome_ids_unprocssed: list[str],
+        failed_ids: list[str] = []
+) -> None:
+    genome_ids = set(genome_ids_unprocssed) - set(failed_ids)
+    for genome_id in genome_ids:
+        genome_dir = os.path.join(work_dir, genome_id)
+        csd_genome_dir = os.path.join(csd, genome_id)
+        loader_helper.create_softlink(csd_genome_dir, genome_dir)
```
Tianhao-Gu, line 301: We need to create the `csd_genome_dir` and link the files within. Just same way as WS downloader. Otherwise, it will break the tool runner which assumes this is a directory. 
https://github.com/kbase/collections/blob/main/src/loaders/compute_tools/tool_common.py#L195-L196

Tianhao-Gu: NV. I looked again, I think WS downloader and NCBI downloader does behaviors the same. I will need to fix the tool runner.

Xiangs18: It is indeed a directory after softlink. I am not sure why you still need to fix the tool runner for that.

Tianhao-Gu: hmmmm. I am seeing those are all links not directories. 

```
(base) tgu@login23 2023.06  $ pwd
/global/u2/t/tgu/collectionssource/CI/GROW/2023.06
(base) tgu@login23 2023.06  $ file -b *
symbolic link to /global/homes/t/tgu/sourcedata/WS/CI/69037/69037_10_1
symbolic link to /global/homes/t/tgu/sourcedata/WS/CI/69037/69037_2_1
symbolic link to /global/homes/t/tgu/sourcedata/WS/CI/69037/69037_4_1
symbolic link to /global/homes/t/tgu/sourcedata/WS/CI/69037/69037_6_1
symbolic link to /global/homes/t/tgu/sourcedata/WS/CI/69037/69037_8_1
```

Xiangs18: where do you get those? Do they exist in /global?

Tianhao-Gu: I download those in using the WS downloader. The full GROW set is in `/global/cfs/cdirs/kbase/collections/collectionssource/CI/GROW` 

```
(base) tgu@login23 GROW  $ pwd
/global/cfs/cdirs/kbase/collections/collectionssource/CI/GROW
(base) tgu@login23 GROW  $ ls
2023.06
(base) tgu@login23 GROW  $ file -b *
setgid, directory
(base) tgu@login23 GROW  $ cd 2023.06/
(base) tgu@login23 2023.06  $ file -b *
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_1000_1
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_100_1
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_1002_1
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_1004_1
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_1006_1
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_1008_1
symbolic link to /global/cfs/cdirs/kbase/collections/sourcedata/WS/CI/69037/69037_10_1
```

Xiangs18: They are dirs and all linked.

Screen Shot 2023-06-30 at 12 00 46 PM

Tianhao-Gu: the `ls -l` shows each genome directory is a link. They are not directories. I lost my clue why `os.path.isdir` works in NERSC but not locally in my container. 

It breaks tool runner when I test locally. But it might work in NERSC? IDK. 
https://github.com/kbase/collections/blob/main/src/loaders/compute_tools/tool_common.py#L195-L196

in NERSC:
```
Python 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:17) [GCC 12.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.12.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os

In [2]: os.path.isdir('/global/cfs/cdirs/kbase/collections/collectionssource/NONE/GTDB/207/GCA_002506415.1')
Out[2]: True

In [3]: os.path.isdir('/global/cfs/cdirs/kbase/collections/collectionssource/NONE/GTDB/207')
Out[3]: True
```

VS in my local container. 
```
Python 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 18:08:17) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.isdir('root/collectionssource/NONE/GTDB/207/GCF_000980105.1')
False
>>> os.path.isdir('root/collectionssource/NONE/GTDB/207')
True
```

Xiangs18: Hi, on my local, it is still dir.

Tianhao-Gu: NV. I figured. I should just mount the source directory (`sourcedata/NCBI/NONE`). Instead I was mounting the directory (`collectionssource/NONE/GTDB/207`)  with softlinks.

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: I'm assuming at some point we'll want to download gbk or gff so we can upload genomes as well as assemblies

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -1,74 +1,76 @@
 """
-PROTOTYPE - Download genome files from NCBI FTP server.
-
-usage: ncbi_downloader.py [-h] --download_file_ext DOWNLOAD_FILE_EXT [DOWNLOAD_FILE_EXT ...] --release_ver
-                          RELEASE_VER [--root_dir ROOT_DIR] [--source SOURCE] [--threads THREADS]
-                          [--overwrite]
+usage: ncbi_downloader.py [-h] --download_file_ext DOWNLOAD_FILE_EXT [DOWNLOAD_FILE_EXT ...] --release_ver {202,207,214,80,83,86,89,95}
```
MrCreosote, line 2: We talked about splitting this into two files:

* A helper module for downloading files from NCBI that just takes a list of IDs and downloads them to an output dir.
* A GTDB specific module that takes CLI arguments, downloads the GTDB specific files, parses them, and uses the helper module to download the NCBI files based on the IDs from the GTDB files.
* There's actually probably a 3rd file (or maybe this goes in loader_helpers) for creating the softlinks in `collectionssource`. That presumably could be reused by future NCBI related downloaders as well

This allows us to add more sources of IDs in the future and reuse the common NCBI related code.

That seems like it's going to change the PR a lot so it probably makes sense to do that and then I'll come back and review it in depth

MrCreosote (changes requested):

src/loaders/ncbi_downloader/ncbi_downloader.py
```diff
@@ -169,32 +175,83 @@ def _fetch_gtdb_genome_ids(release_ver, work_dir):
...
+    os.makedirs(work_dir, exist_ok=True)
+    return work_dir
 
-    if source == 'GTDB':
-        work_dir = os.path.join(root_dir, source_data_dir, source, f'r{release_ver}')
-    else:
-        raise ValueError(f'Unexpected source: {source}')
 
-    os.makedirs(work_dir, exist_ok=True)
+def _make_collection_source_dir(
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader.py
Xiangs18: Updated!

src/loaders/common/loader_helper.py
```diff
@@ -184,6 +184,69 @@ def is_upa_info_complete(upa_dir: str):
...
+def make_collection_source_dir(
+        root_dir: str,
+        collection_source_dir: str,
+        env: str,
+        collection: str,
+        release_ver: str
+) -> str:
+    """
+    Helper function that creates a collection & source_version and link in data
+    to that colleciton from the overall source data dir.
```
Tianhao-Gu, line 196: *collection

src/loaders/common/loader_helper.py
```diff
@@ -184,6 +184,69 @@ def is_upa_info_complete(upa_dir: str):
     return True
 
 
+def make_collection_source_dir(
+        root_dir: str,
+        collection_source_dir: str,
```
Tianhao-Gu, line 189: do you think we can make `collection_source_dir` defaults to `loader_common_names.COLLECTION_SOURCE_DIR`? 
And we can probably remove `form_source_dir` and replace usage with this function.

src/loaders/common/loader_helper.py
```diff
@@ -184,6 +184,69 @@ def is_upa_info_complete(upa_dir: str):
...
+    """
+    Helper function that creates a collection & source_version and link in data
+    to that colleciton from the overall source data dir.
+    """
+    csd = os.path.join(root_dir, collection_source_dir, env, collection, release_ver)
+    os.makedirs(csd, exist_ok=True)
+    return csd
+
+
+def create_softlinks_in_csd(csd: str, work_dir: str, genome_ids: list[str], taxonomy_files: list[str] = []) -> None:
```
Tianhao-Gu, line 203: avoid mutable defaults. 

```suggestion
def create_softlinks_in_csd(csd: str, work_dir: str, genome_ids: list[str], taxonomy_files: list[str] = []) -> None:
```

    if taxonomy_files is None:
        taxonomy_files = []

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
+"""
+usage: ncbi_downloader.py [-h] --download_file_ext DOWNLOAD_FILE_EXT [DOWNLOAD_FILE_EXT ...] --release_ver {202,207,214,80,83,86,89,95}
```
Tianhao-Gu, line 2: description and usage probably also needs to be updated accordingly.

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
...
+        "--release_ver",
+        required=True,
+        type=str,
+        choices=gtdb_release_vers,
+        help="GTDB release version",
+    )
+
+    # Optional argument
+    optional.add_argument(
+        f"--{loader_common_names.KBASE_COLLECTION_ARG_NAME}",
```
Tianhao-Gu, line 209: Since we isolate gtdb downloading process, do we still need `kbase_collection` argument? Will we ever have a case where this won't be 'GTDB' if we call `gtdb.py`?

MrCreosote: I have a question around that here: https://kbase-jira.atlassian.net/browse/RE2022-209?focusedCommentId=91635

Namely, we're going to have 2 similar data sets - the data that was downloaded from NCBI using the GTDB IDs, which lives in `sourcedata/NCBI` and `collectionssource/<env>/<GTDB?>`, and the data that was loaded to the workspace, which lives in `sourcedata/WS/<env>` (I think) and `collectionssource/<env>/<GTDB?>`.

Those `<GTDB?>`s probably need to be different

Xiangs18: `collectionssource/<env>/GTDB_LOADED` makes sense to me.

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
...
+        root_dir, loader_common_names.SOURCE_DATA_DIR, source, loader_common_names.DEFAULT_ENV
+    )
+    csd = loader_helper.make_collection_source_dir(
+        root_dir,
+        loader_common_names.COLLECTION_SOURCE_DIR,
+        loader_common_names.DEFAULT_ENV,
+        kbase_collection,
+        release_ver,
+    )
+    genome_ids_unprocssed, taxonomy_urls = _fetch_gtdb_genome_ids(release_ver, work_dir)
```
MrCreosote, line 260:
```suggestion
    genome_ids_unprocessed, taxonomy_urls = _fetch_gtdb_genome_ids(release_ver, work_dir)
```

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
...
+            ):
+                ext_count += 1
+
+        if ext_count != target_ext_count:
+            genome_ids.append(genome_id)
+
+    return genome_ids
+
+
+def main():
```
MrCreosote, line 180: main is 137 lines which is way too long, and needs some splitting up. Arg parsing is one sub function that could be split out, for example

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
...
+        overwrite,
+    )
+    if not genome_ids:
+        print(f"All {len(genome_ids_unprocssed)} genomes files already exist in {work_dir}")
+        loader_helper.create_softlinks_in_csd(csd, work_dir, genome_ids_unprocssed, taxonomy_files)
+        return
+
+    if not threads:
+        threads = max(int(cpu_count() * min(SYSTEM_UTILIZATION, 1)), 1)
+    threads = max(1, threads)
```
MrCreosote, line 276: We should move all the threading stuff into the ncbi downloader helper so it can be shared. No need to reinvent the the multiprocessing wheel every time we want to call those functions

MrCreosote (changes requested):

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: Actually, it's not a problem, since they'll have different <env> values

As such, to my mind we could hard code the GTDB collection name here

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if not success:
+        raise ValueError(f'Download Failed for {gene_id} after {max_attempts} attempts!')
+
+
+def download_genome_files(
+        gene_ids: list[str],
+        target_file_ext: list[str],
+        exclude_name_substring: list[str],
+        result_dir: str,
```
MrCreosote, line 58: @Tianhao-Gu I've been flip flopping on whether this module should understand the `collections` file structure and just take a root dir argument or be more flexible with where it stores the output. 

It seems weird to me to effectively pass in the source ID (NCBI) and hardcode it in the GTDB specific code when this code knows exactly what the source is.

src/loaders/common/loader_helper.py
```diff
@@ -184,6 +184,69 @@ def is_upa_info_complete(upa_dir: str):
     return True
 
 
+def make_collection_source_dir(
+        root_dir: str,
+        collection_source_dir: str,
+        env: str,
+        collection: str,
+        release_ver: str
```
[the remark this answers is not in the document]
Xiangs18: If my understanding is correct, it is release_ver here.

src/loaders/ncbi_downloader/ncbi_downloader.py
MrCreosote: :+1:

src/loaders/common/loader_helper.py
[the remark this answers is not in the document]
MrCreosote: It's making a dir in `collectionssource` right? If so that's def a source_ver

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Tianhao-Gu: hmmm I didn't catch what you mean. We need to form the result dir somewhere essentially.

MrCreosote: IOW do we leave this as is, where it's just any directory, or instead do we pass in the collections root dir and this code appends `sourcedir/NCBI` (or whatever). For the latter, then the calling code doesn't have to hardcode `NCBI` anywhere, which really it shouldn't - this NCBI downloading code should know that it's supposed to save stuff in the NCBI directory.

Tianhao-Gu: I see. Yea. I think the latter makes sense.

src/loaders/common/loader_helper.py
Xiangs18: Updated!

src/loaders/common/loader_helper.py
Xiangs18: Updated!

src/loaders/common/loader_helper.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/common/loader_helper.py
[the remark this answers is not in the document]
Xiangs18: Updated!

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
...
+    optional = parser.add_argument_group("optional arguments")
+
+    # Required flag argument
+    required.add_argument(
+        "--download_file_ext",
+        required=True,
+        type=str,
+        nargs="+",
+        help="Download only files that match given extensions",
+    )
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,311 @@
...
+    optional.add_argument(
+        "--root_dir",
+        type=str,
+        default=loader_common_names.ROOT_DIR,
+        help="Root directory",
+    )
+    optional.add_argument(
+        "--source", type=str, default="NCBI", choices=SOURCE, help="Source of data"
+    )
```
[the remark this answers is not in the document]
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: Updated!

test/src/loaders/ncbi_downloader/ncbi_downloader_test.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: The threading stuff is still all in gtdb.py?

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: This code is still just taking any old result dir - did you maybe not push your latest commit(s)?

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: https://github.com/kbase/collections/blob/a4c61eaa0b1bb6d0caba947564ec3cd3b9f939cd/src/loaders/ncbi_downloader/gtdb.py#L262

src/loaders/ncbi_downloader/gtdb.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/ncbi_downloader/gtdb.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: Still kind of long but moving the threading code to the ncbi helper script will help so for now :+1:

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: I'm talking about the entire threading mechanism, not just calculating the thread count. E.g. if I call the ncbi helper from another script, I just pass it the system utilization and/ or thread count I want and it does the rest

MrCreosote: Otherwise we have to repeat the threading code for every script that calls the ncbi helper

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,297 @@
+"""
+usage: gtdb.py [-h] --release_ver {202,207,214,80,83,86,89,95} [--root_dir ROOT_DIR]
```
MrCreosote, line 2: The markdown documentation will need to be updated as well.

src/loaders/compute_tools/tool_common.py
```diff
@@ -617,6 +610,13 @@ def create_fatal_tuple(
     return fatal_tuple
 
 
+def get_threads(system_utilization: float, threads: int = None) -> int:
+    if not threads:
```
MrCreosote, line 614: If it's a public method it needs pydoc

src/loaders/compute_tools/tool_common.py
```diff
@@ -617,6 +610,13 @@ def create_fatal_tuple(
     return fatal_tuple
 
 
+def get_threads(system_utilization: float, threads: int = None) -> int:
```
MrCreosote, line 613: The tool_common file should only contain helper methods that are exclusive to tools. If a method is used by other parts of the pipeline it should go in loader helpers

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: I misunderstood the conversation and made the change [here](https://github.com/kbase/collections/blob/a4c61eaa0b1bb6d0caba947564ec3cd3b9f939cd/src/loaders/ncbi_downloader/gtdb.py#L241). But looking it now since we already created work_dir/result previously in the script, why cannot we just pass in? Also, is more general and handy for testing purpose. Just my thoughts

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Hi, may I know exactly which part I need to update here?

src/loaders/compute_tools/tool_common.py
Xiangs18: then I assume threading stuff should be put into loader_helpers instead of ncbi helper?

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: it's basically the reason given in the OP - it's weird for another script to be telling the NCBI script that the source of the data is NCBI. The ncbi script knows that, and so should be able to handle it itself. Maybe the way to do is have a function in the ncbi helper like

```
get_work_dir(root: Path):
     return root / loader_common_names.SOURCE_DATA_DIR / loader_common_names.DEFAULT_ENV
```
and call that from gtdb

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: Anything refers to the old ncbi script which now doesn't exist, and the replacement has different arguments

src/loaders/compute_tools/tool_common.py
MrCreosote: Hmm... if you can make it generic enough for reuse in other places sure. That being said, AFAIK there's nothing else making use of it right now (or am I wrong about that?) and so it could live in the ncbi code for now and we could do the refactoring when we actually need it and have a real use case for figuring out which parts are sharable and which aren't

Xiangs18: tool_common.py is also using it for sure.

MrCreosote: In the same or very close way? If so then yeah, it'd make sense to move it into the common loader code or its own module

Xiangs18: nah, I feel they are kinda diff in terms of tool_common.py has more data processing in between get_ threads() and multiprocessing.Pool call. Also multiprocessing.Pool call is nested inside _execute() func as well. Not quite sure how can I make it generic enough that can reuse in both places other than get_threads().

MrCreosote: Ok, I'd leave it in the ncbi code for now then

Xiangs18: Just to make sure we are on the same page, we are not changing multithreading stuff in gtdb.py file for now.

MrCreosote: All the multithreading code should move from gtdb.py to the ncbi helper so it can be reused when we write another script that downloads stuff from ncbi

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated! Moved into ncbi helper.

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated! But please let me know if I missing anything again:)

src/loaders/compute_tools/tool_common.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: The NCBI helper code needs to create the work dir as there's no guarantee it's created when the code is called

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: You're probably more familiar with the markdown documentation than I am at this point.

That being said, there's no changes to the markdown docs in the PR AFAICT

src/loaders/compute_tools/tool_common.py
MrCreosote: This is now moot, method deleted

src/loaders/compute_tools/tool_common.py

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
```diff
@@ -0,0 +1,144 @@
...
+        genome_ids: list[str],
+        target_file_ext: list[str],
+        exclude_name_substring: list[str],
+        system_utilization: float,
+        threads: int = None,
+        overwrite: bool = False
+) -> list[list[str]]:
+    """
+    Download genome files from NCBI FTP server in parallel using multiprocessing
+    """
```
MrCreosote, line 117: This really needs documentation of the arguments like the single file download method above

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: ah, this function is get_work_dir, not make_work_dir. A work_dir will be created when _make_work_dir is called.

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,287 @@
...
+    work_dir = _make_work_dir(root_dir)
+    csd = loader_helper.make_collection_source_dir(
+        root_dir,
+        loader_common_names.DEFAULT_ENV,
+        KBASE_COLLECTION,
+        release_ver,
+    )
+    genome_ids_unprocessed, taxonomy_urls = _fetch_gtdb_genome_ids(release_ver, work_dir)
+    taxonomy_files = [os.path.basename(url) for url in taxonomy_urls]
+    genome_ids = _process_genome_ids(
```
MrCreosote, line 247: Now I look at it, _process_genome_ids is non-gtdb specific as well and should be moved into the ncbi code for reuse, probably with a better name, maybe something like `remove_ids_with_existing_data` or something

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: That function is never called in the ncbi code.

Xiangs18: But we already had https://github.com/kbase/collections/blob/89402df9d134b2fc93c06fee645e1754a4e2e499/src/loaders/ncbi_downloader/ncbi_downloader_helper.py#L82

MrCreosote: Ah, I missed that. I was expecting the work dir to be created before the threads were fired off so there was only one attempt to create it rather than `n_threads`.

Still probably fine as is so :+1:

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated markdown. Also add env that's missing

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: I moved everything ato get_work_dir in ncbi helper. Updated!

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: :+1:

docs/generate_genome_data.md
```diff
@@ -11,7 +11,7 @@ The scripts used in the process are located in the `src/loaders` directory.
 1. Retrieve Source Genome Files
     * From NCBI
```
MrCreosote, line 12:
```suggestion
    * From GTDB
```

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
```diff
@@ -0,0 +1,196 @@
...
+) -> list[list[str]]:
+    """
+    Download genome files from NCBI FTP server in parallel using multiprocessing
+
+    root_dir: root directory for the collections project.
+    genome_ids: genome ids that parsed from GTDB metadata file (e.g. GB_GCA_000016605.1, RS_GCF_000968435.2).
+    target_file_ext: download only files that match given extensions.
+    exclude_name_substring: exclude files that contain given substrings in their names.
+    system_utilization: fraction of CPU cores to use.
+    threads: number of threads to use.
```
MrCreosote, line 122:
```suggestion
    threads: number of threads to use. Takes precedence over system_utilization if supplied.
```

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
```diff
@@ -0,0 +1,196 @@
...
+        exclude_name_substring: list[str],
+        system_utilization: float,
+        threads: int = None,
+        overwrite: bool = False
+) -> list[list[str]]:
+    """
+    Download genome files from NCBI FTP server in parallel using multiprocessing
+
+    root_dir: root directory for the collections project.
+    genome_ids: genome ids that parsed from GTDB metadata file (e.g. GB_GCA_000016605.1, RS_GCF_000968435.2).
```
MrCreosote, line 118:
```suggestion
    genome_ids: NCBI genome ids to download (e.g. GB_GCA_000016605.1, RS_GCF_000968435.2).
```

May or may not be parsed from the GTDB file in future

MrCreosote: Same comment for the docs in other 2 functions

MrCreosote (commented):
Easy stuff

src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -0,0 +1,245 @@
...
+    return release_ver
+
+
+def _form_gtdb_taxonomy_file_url(release_ver: str) -> list[str]:
+    # form GTDB taxonomy URL for specific GTDB version.
+    # e.g. https://data.gtdb.ecogenomic.org/releases/release207/207.0/ar53_taxonomy_r207.tsv
+
+    file_dir_url = (
+        GTDB_DOMAIN + f"release{release_ver}/" + f"{release_ver}.0/"
+    )  # TODO: add dot support, relase version 86 has '86.1' and '86.2'
```
MrCreosote, line 89:
```suggestion
    )  # TODO: add dot support, release version 86 has '86.1' and '86.2'
```

docs/generate_genome_data.md
Xiangs18: Why GTDB? we download files from the NCBI FTP server though.

MrCreosote: Because the collection we're creating is GTDB, and we're downloading the NCBI IDs that occur in that collection from GTDB.

Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: Updated!

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
Xiangs18: Updated!

docs/generate_genome_data.md
MrCreosote: :+1:

src/loaders/ncbi_downloader/gtdb.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: :+1:

src/loaders/ncbi_downloader/ncbi_downloader_helper.py
MrCreosote: :+1:

MrCreosote (commented):

MrCreosote (approved):
Doh. If @Tianhao-Gu approves so do I agani

Tianhao-Gu (approved):
👍

PR #354, opened 2023-07-01
Document data paging options

MrCreosote:


MrCreosote:
No idea how adding a text file can affect coverage

dauglyon (approved):
Looks to cover everything we discussed!

design/paging.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Skip and limit / offset and limit
+
+This is the standard paging solution of providing a count of documents to skip and a count
+of documents to return.
+
+#### Pros
+
+* Can page to any location in the data
+* Easily understood by users
```
dauglyon, line 28: I'm not sure they pay too much attention to anything besides the page number. Is that what you're getting at here?

design/paging.md
```diff
@@ -0,0 +1,119 @@
...
+* Prior `start_after` values can be cached by the front end to allow for returning to a
+  previous page
+* At worst linear time complexity
+
+#### Cons
+
+* Can only navigate to the next page after the current page. Navigating to page 10 from page
+  1 would require 9 full page loads.
+* May not work well on non-unique fields
+  * If a page has more than a page size's number of identical values in the field being paged over,
```
dauglyon, line 64: Couldn't this be solved by having an numbered index per sortable column? That might be a bit bloated though...

design/paging.md
```diff
@@ -0,0 +1,119 @@
...
+
+* Any data filtering or deletion disrupts the ordering and may result in pages less than the
+  expected size or entirely empty pages
+
+#### Notes
+
+* This is one of the paging strategies the KBase workspace supports, using the object ID within
+  a particular workspace.
+
+### Calculate and cache paging
```
dauglyon, line 94: From a UI perspective (ignoring the server side complexity) this is my preferred solution as it can be implemented similarly to skip&limit which is well-supported in our query lib.

dauglyon (approved):
Looks to cover everything we discussed!

design/paging.md
MrCreosote: It's just that they're familiar with it and aren't likely to be confused when presented with the UI

design/paging.md
MrCreosote: That's essentially solution #3

design/paging.md
MrCreosote: Well... maybe. 

* You have to provide the query ID every time you page
* It may take a loooong time to return the first page (or in a hybrid system, the first time the full query index is calculated)
* For background index building, you'll have to check and see if a page has been indexed yet before you can actually go there (or the server will response with some sort of `page not ready` error

design/paging.md
MrCreosote: If you're going to do that you might was well use that numbered index as the `start_after` value

MrCreosote: Added to the doc

design/paging.md
dauglyon: Hmm good point. I may have to build some prototypes and see where the sticky bits are.

MrCreosote: There's no guarantee we're going to go this route so I wouldn't spend a ton of time on it

dauglyon (approved):

PR #363, opened 2023-07-05
RE2022-177: refactor taskfarmer with env

Tianhao-Gu:


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -338,8 +373,17 @@ def main():
 
     image_str = _fetch_image(REGISTRY, tool, job_dir, tag=args.image_tag, force_pull=not args.use_cached_image)
     wrapper_file = _create_shifter_wrapper(job_dir, image_str)
-    task_list_file, n_jobs = _create_task_list(source_data_dir, kbase_collection, load_ver, tool, wrapper_file, job_dir,
-                                               root_dir, source_file_ext=source_file_ext)
+    task_list_file, n_jobs = _create_task_list(
+        source_data_dir,
```
MrCreosote, line 377: Source data dir is redundant with the other args in this method, right? You can construct it from them

Tianhao-Gu: 👍 yea. Added `form_source_dir` helper func so that we can reuse it.

MrCreosote (approved):
LGTM

PR #365, opened 2023-07-05
RE2022-203: Refactor count methods for reuse in samples

MrCreosote:
& share code with genome attribs and heatmaps


Tianhao-Gu (approved):
LGTM

PR #369, opened 2023-07-05
RE2022-203: Add beginnings of samples API

MrCreosote:
just counts rows in the DB for now.

### DIFF
--- a/src/service/data_products/samples.py
+++ b/src/service/data_products/samples.py
@@ -1,0 +1,219 @@
+"""
+The samples data product, which provides sample information for a collection.
+"""
+
+from collections import defaultdict
+import logging
+
+from fastapi import APIRouter, Request, Depends, Query
+from pydantic import BaseModel, Field, Extra
+import src.common.storage.collection_and_field_names as names
+from src.common.product_models.common_models import SubsetProcessStates
+from src.service import app_state
+from src.service.app_state_data_structures import CollectionsState, PickleableDependencies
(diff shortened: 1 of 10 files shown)


src/service/data_products/samples.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+
+
+# At some point we're going to want to filter/sort on fields. We may want a list of fields
+# somewhere to check input fields are ok... but really we could just fetch the first document
+# in the collection and check the fields 
+@_ROUTER.get(
+    "/",
+    response_model=SamplesTable,
+    description="Get the corresponding sample attribues for each data unit in the collection, "
```
Tianhao-Gu, line 137: *attributes

Tianhao-Gu (dismissed):
other than the typo, LGTM. There are a few unused imports, those are there for future usage?

src/service/data_products/samples.py
MrCreosote: fixed

MrCreosote:
> There are a few unused imports, those are there for future usage?

Maybe, but probably also because I've tried to get linting working several times in VSC (which is something that should Just Work, the amount of JSON based config you have to do is ridiculous) and can never get it to recognize imports in the same project (and some other quirks). On the last PR for samples I'll clean up any extraneous imports.

Tianhao-Gu (approved):
👍

PR #370, opened 2023-07-05
RE2022-177: fix import requirments

Tianhao-Gu:
avoid adding all dependencies from loader_helper for all images.


MrCreosote:
> avoid adding all dependencies from loader_helper for all images.

Why is that an issue?

Tianhao-Gu:
> 

tool images doesn't have libraries like `jsonlines`. Unless we want to install `jsonlines` to all images?

MrCreosote (approved):
Ok I see. This seems fine then

PR #371, opened 2023-07-06
RE2022-177: refactor parser with env

Tianhao-Gu:


src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -355,9 +365,9 @@ def _process_mash_tool(root_dir: str,
                 raise ValueError(f'Unable to locate the sketch file: {sketch_file} for genome: {data_id}')
             sketch_files.append(sketch_file)
 
-    create_import_files(root_dir, f'{kbase_collection}_{load_ver}_{SEQ_METADATA}', seq_meta)
+    create_import_files(root_dir, env, f'{kbase_collection}_{load_ver}_{SEQ_METADATA}', seq_meta)
 
-    import_dir = Path(root_dir, loader_common_names.IMPORT_DIR)
+    import_dir = Path(root_dir, loader_common_names.IMPORT_DIR, env)
     os.makedirs(import_dir, exist_ok=True)
```
MrCreosote, line 371: Maybe there should be common code for making  the import dir

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -863,10 +876,10 @@ def checkm2(root_dir, kbase_collection, load_ver, fatal_ids):
...
 
 
-def _retrieve_sample(root_dir, kbase_collection, load_ver):
-    print(f'Parsing sample data for {kbase_collection} collection, version {load_ver}.')
-
-    source_dir = os.path.join(root_dir, loader_common_names.COLLECTION_SOURCE_DIR, kbase_collection, load_ver)
+def _retrieve_sample(root_dir, env, kbase_collection, source_ver, load_ver):
+    print(f'Parsing sample data for {kbase_collection} collection, load version {load_ver}, '
+          f'source version {source_ver}.')
+    source_dir = form_source_dir(root_dir, env, kbase_collection, source_ver)
```
MrCreosote, line 882: Hmm, sample data is kind of odd because there's nothing in `collectionsdata`, unlike all the other data, right? I'm wondering if we should have softlinks or something rather than having this one off. I think that's the only reason this script needs to know the `source_ver` as well, right?

On the other hand, this might be a lot easier to deal with once we swap all these files for a database, since we could record the load_ver -> source_ver relationship when it's first created.

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -902,13 +918,24 @@ def main():
...
+    optional.add_argument(
+        f"--{loader_common_names.ENV_ARG_NAME}",
+        type=str,
+        choices=loader_common_names.KB_ENV + [loader_common_names.DEFAULT_ENV],
+        default='PROD',
+        help="Environment containing the data to be processed. (default: PROD)",
+    )
+
+    optional.add_argument(f'--{loader_common_names.LOAD_VER_ARG_NAME}', type=str,
+                          help=loader_common_names.LOAD_VER_DESCR + ' (defaults to the source version)')
```
MrCreosote, line 937: What happens if someone makes a mistake and forgets to enter the correct new load_ver, and the old load_ver that == source_ver already exists?

Tianhao-Gu: * if load_ver == source_ver does exist, it will just overwrites existing result jsonl files with exactly the same content. 
* if loand_ver == source_ver doesn't exist, it will error due to result dir cannot be found. 
We should be fine in either situation.

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: we don't specify load version during download process. Where are we establishing the softlinks?

MrCreosote: > we don't specify load version during download process.

Good point. I guess it's either an oddball here or an oddball there

> Where are we establishing the softlinks?

I was thinking of something along the lines of linking `collectionsdata/<env>/<coll>/<load_ver>` to `collectionssource/<env>/<coll>/<source_ver>`. Moot anyway given the above.

MrCreosote: Although - samples' load_ver will always be the same as the source_ver since there's no calculation step. Anyway, probably easier to deal with in the database based implementation

src/loaders/genome_collection/parse_tool_results.py
MrCreosote: > if load_ver == source_ver does exist, it will just overwrites existing result jsonl files with exactly the same content.

Ok, you're right for the parser. What about for the tool runner and tool images?

src/loaders/genome_collection/parse_tool_results.py
MrCreosote: I was thinking `make_import_dir` would go in `loader_helpers` and be used by the function that writes the import files as well

Tianhao-Gu: yup yup. I realized the same after commit. Made the update.

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: if load_ver == source_ver already exists, tool runner will warn user and user has to use --force.
if load_ver == source_ver doesn't exist, the tool runner will run and create a source_ver dir in the collectionsdata dir.

src/loaders/genome_collection/parse_tool_results.py

src/loaders/genome_collection/parse_tool_results.py
MrCreosote: :+1:

MrCreosote (approved):

PR #377, opened 2023-07-11
fix typos and spelling errors

random415:
Fixing some minor typos and spelling errors which should not affect functionality but improve the overall quality of documentation.


MrCreosote:
Hi @random415, thanks for this. You likely couldn't have known this but the code in `src/clients` is generated from specs in other repos so any fixes here will get blown away the next time they're regenerated. Also there was a key name change that needed testing.

As it happens I was also working on some code cleanup and so I incorporated your changes in #378 and tested them along with my changes.  

Thanks again for the fixes.

PR #380, opened 2023-07-17
RE2022-105: build push tool img using GHA

Tianhao-Gu:


.github/workflows/build-push-tool-images.yml
```diff
@@ -0,0 +1,66 @@
+name: Build & Push Tool Image to GHCR
```
MrCreosote, line 1: I was thinking we would only trigger image builds as needed, not on every PR. Most PRs they probably won't change

.github/workflows/build-push-tool-images.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          tools=(
+            mash
+            gtdb_tk
+            checkm2
+            microtrait)
+
+          for tool in "${tools[@]}"; do
+            dockerfile="./src/loaders/compute_tools/$tool/Dockerfile"
+            tag="ghcr.io/${{ github.event.repository.owner.login }}/${{ github.event.repository.name }}:$tool"
```
MrCreosote, line 54: Unless I'm mistaken this tag will be the same every time there's an new image, right? So there's no way to select older images?

MrCreosote, line 54: Also once we start adding version tags to the repo those should be incorporated into the tool tags, I'd think

.github/workflows/build-push-tool-images.yml
```diff
@@ -0,0 +1,66 @@
...
+
+            echo "Building and pushing $tool image..."
+            echo "tagging image with $tag"
+
+            docker buildx build --file "$dockerfile" --tag "$tag" \
+              --build-arg "BUILD_DATE=${{ steps.date.outputs.date }}" \
+              --build-arg "VCS_REF=${{ github.sha }}" \
+              --build-arg "BRANCH=${{ github.ref }}" \
+              --build-arg "TAG=${{ github.ref }}" \
```
MrCreosote, line 63: Should this be $tag or is github.ref intentional?

.github/workflows/build-push-tool-images.yml
Tianhao-Gu: I did add `path`. Only one of the tool Dockerfile changes, this will be triggered.

.github/workflows/build-push-tool-images.yml
Tianhao-Gu: yea, tag will be the same for every image. Setting tags to be collections:tool_name is preferred by sysadmin.
https://kbase.slack.com/archives/C02MHNMCQ/p1689612602761919

But if you have suggestions, I can change how to tag forms. Will there be a case where we need to run tools with old builds?

.github/workflows/build-push-tool-images.yml
Tianhao-Gu: hmmm I had the same question. But I copied this from the template. Tag does seem to be overwritten by explicitly defined `tags`. So I didn't change anything just to be safe. 
https://github.com/kbase/.github/blob/main/.github/workflows/reusable_build-push.yml#L57-L60

.github/workflows/build-push-tool-images.yml
Tianhao-Gu: I think `github.event.repository.name` will always be `collections`. Depends on how we add version tags to the main repo, this might need to be updated in the future. yes.

.github/workflows/build-push-tool-images.yml
MrCreosote: That means if one gets rebuilt all get rebuilt right? I was thinking we'd just have a button per tool or something

.github/workflows/build-push-tool-images.yml
MrCreosote: I didn't get the impression that they were stating a preference, just providing options, and the same tag is the easiest.

Anyway, to me having the same tag for all images seems like a debugging / reproducibility nightmare. I'd either incorporate the hash or the version into the tag, so something like mash_<githash> or match_v.0.4.1

MrCreosote: Yeah the version would have to go into the image tag (and maybe the image metadata). 

Why not just figure it out now? Then we don't have to worry about it in the future. The way tagging works in other repos is that a manual trigger is required; it might make sense to do the same thing here

.github/workflows/build-push-tool-images.yml
MrCreosote: Fair enough

:+1:

.github/workflows/build-push-tool-images.yml
Tianhao-Gu: how do you want the tag to be tracked across the repo? it needs to be pull by task runner.

Tianhao-Gu: we don't have a collection release version. I don't know what needs to be figure out here. I also didn't get why/how collection version needs to be tagged with tool image. How do you want checkm2 being tagged for collections:0.0.1 vs collections:0.0.2?

.github/workflows/build-push-tool-images.yml
Tianhao-Gu: you mean having GHA figure out which Docker file is changed and just trigger that tool build?

.github/workflows/build-push-tool-images.yml
MrCreosote: All 3 of these comments are related to each other, I think, so I'll consolidate the conversation here.

The way I would do this is
* build all the tools every PR but don't push
    * This is relatively quick and is parallel to the tests, and makes sure the tool doesn't break
* Add an action for each tool, similar to the manual release action, that builds and pushes an image with a tag something like `<tool_name>_<githhash>`
    * When we need a new tool image we just manually activate that action, which probably isn't very often
* Add an action for each tool (unless they can be combined into the action above) that builds and pushes an image like the above but uses a tool specific release version, so we could tag the repo with `mash_0.1.0` or something and release from that.
    * This allows releasing new versions of tools without having to release a new version of the rest of the code / other tools, etc.
    * Maybe there's a way to parameterize the actions with the tool name in the actions UI so that we don't need a separate action for each tool
* By default the tool runner would just pull the latest version (and record the githash / version when we get to that), but it could have an argument for supplying specific tool tags.

Tianhao-Gu:
@MrCreosote Updated workflow to monitor a version file for each tool.

MrCreosote:
Once this is agreeable to everyone maybe this PR should be closed and one with just one commit made

Or squash merge, although I'm generally not a fan of that

src/loaders/compute_tools/tool_version.py
```diff
@@ -0,0 +1,36 @@
+import sys
+
+import yaml
```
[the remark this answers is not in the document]
Tianhao-Gu: hmmm. GHA didn't complain about it. I will need to use this function in my next PR to retrieve version for taskfarmer. If I cannot import, I will update pipfile.

src/loaders/compute_tools/tool_version.py
```diff
@@ -0,0 +1,36 @@
+import sys
```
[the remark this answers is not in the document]
Tianhao-Gu: yup.

.github/workflows/build-push-tool-images.yml
```diff
@@ -0,0 +1,59 @@
...
+          password: '${{ secrets.GHCR_TOKEN }}'
+
+      - name: Retrieve latest tool version
+        id: extract_tool_version
+        run: |
+          tool_name="${{ inputs.tool_name }}"
+          current_version=$(python src/loaders/compute_tools/tool_version.py ${{ inputs.version_file }})
+          tool_version="${tool_name}_${current_version}"
+          echo "Current tool version: $tool_version"
+          echo "tool_version=$tool_version" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
Tianhao-Gu: echo to `$GITHUB_OUTPUT` is the new preferred way to set environment var while `set-output` is retiring. 
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/build-push-tool-images.yml
```diff
@@ -0,0 +1,59 @@
...
+          tag="ghcr.io/${{ github.event.repository.owner.login }}/${{ github.event.repository.name }}:${{ steps.extract_tool_version.outputs.tool_version }}"
+
+          echo "Building and pushing ${{ inputs.tool_name }} image..."
+          echo "tagging image with $tag"
+
+          docker buildx build --file "$dockerfile" --tag "$tag" \
+            --build-arg "BUILD_DATE=${{ steps.date.outputs.date }}" \
+            --build-arg "VCS_REF=${{ github.sha }}" \
+            --build-arg "BRANCH=${{ github.ref }}" \
+            --build-arg "TAG=${{ github.ref }}" \
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/compute_tools/checkm2/versions.yaml
```diff
@@ -0,0 +1,3 @@
+versions:
+  - version: 0.0.1
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_version.py
```diff
@@ -0,0 +1,36 @@
...
+
+import yaml
+
+
+def extract_latest_version(file_path: str) -> str:
+    """
+    Extracts the latest version from a YAML file based on the date specified.
```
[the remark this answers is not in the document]
Tianhao-Gu: updated the doc. I want to refer to the date specified inside the version file.

src/loaders/compute_tools/tool_version.py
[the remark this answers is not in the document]
MrCreosote: Probably because a dependency in the Pipfile has a dependency on it. But that means if we update that dependency and it no longer depends on pyyaml or if we remove that dependency things will break. We should explicitly add it to the Pipfile if we depend on it directly

src/loaders/compute_tools/tool_version.py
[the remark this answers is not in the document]
MrCreosote: :

.github/workflows/build-push-tool-images.yml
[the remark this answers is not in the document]
MrCreosote: :+1:

.github/workflows/build-push-tool-images.yml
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/tool_version.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/tool_version.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍 added pyyaml to pipfile.

.github/workflows/build-push-tool-images.yml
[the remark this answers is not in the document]
MrCreosote: So actually what you did was good, but I was thinking of adding just the version string (e.g. 0.1.0 or whatever) as well for easier parsing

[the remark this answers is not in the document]
MrCreosote: or you could just replace it with the version string since the tag has redundant info

src/loaders/compute_tools/tool_version.py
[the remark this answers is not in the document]
MrCreosote: :+1:

.github/workflows/build-push-tool-images.yml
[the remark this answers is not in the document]
Tianhao-Gu: 👍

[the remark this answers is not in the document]
MrCreosote: :+1:

.github/workflows/build-push-tool-images.yml
```diff
@@ -0,0 +1,61 @@
...
+          tool_version="${tool_name}_${current_version}"
+          echo "Current tool version: $tool_version"
+          echo "tool_version=$tool_version" >> $GITHUB_OUTPUT
+          echo "version_str=$current_version" >> $GITHUB_OUTPUT
+
+      - name: Build and push tool images
+        id: build-and-push
+        run: |
+          dockerfile="./src/loaders/compute_tools/${{ inputs.tool_name }}/Dockerfile"
+          tag="ghcr.io/${{ github.event.repository.owner.login }}/${{ github.event.repository.name }}:${{ steps.extract_tool_version.outputs.tool_version }}"
```
MrCreosote, line 50: Hold on - shouldn't the tag just be `<tool>_<ver>`? Or am I misremembering how this works?

Tianhao-Gu: I learnt it from the template. It's not just the <tool>_<ver>. 
https://github.com/kbase/.github/blob/main/.github/workflows/reusable_build-push.yml#L46

MrCreosote: :+1:

MrCreosote:
This seems fine, I'm just thinking it might make sense to close this PR and make a new one with just one commit rather than a large number of really small commits all going to the same greenfield files

PR #388, opened 2023-07-21
RE2022-105: download img from ghcr

Tianhao-Gu:


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -365,7 +373,7 @@ def main():
     job_dir = task_mgr.job_dir
     _create_job_dir(job_dir, destroy_old_job_dir=args.force)
 
-    image_str = _fetch_image(REGISTRY, tool, job_dir, tag=args.image_tag, force_pull=not args.use_cached_image)
```
MrCreosote, line 368: Similarly I'm wondering if the use cached image flag is still needed since we're now tagging the images with an explicit version, so we should always use that version and the exact image it tags. As such either it's on the system or it isn't and we don't need to worry about the cache, right...?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -110,13 +112,19 @@ def _write_to_file(file_path, content):
...
+def _fetch_image(registry, tool, job_dir, force_pull=True):
     """
     Fetches the specified Shifter image if it is not already present on the system.
 
     When force_pull is set to True, the image is always pulled from the registry
     """
-    image_str = f"{registry}/{image_name}:{tag}"
+
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    compute_tools_dir = os.path.join(current_dir, '../../compute_tools')
```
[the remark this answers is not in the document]

src/loaders/jobs/taskfarmer/task_generator.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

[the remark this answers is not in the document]
Tianhao-Gu: 👍

Tianhao-Gu: 👍

src/loaders/jobs/taskfarmer/task_generator.py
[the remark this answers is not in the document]

src/loaders/jobs/taskfarmer/task_generator.py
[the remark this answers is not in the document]
MrCreosote: :+1:

MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -110,13 +107,19 @@ def _write_to_file(file_path, content):
...
-    image_str = f"{registry}/{image_name}:{tag}"
+
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    compute_tools_dir = os.path.join(current_dir, COMPUTE_TOOLS_DIR)
+    version_file = os.path.join(compute_tools_dir, tool, VERSION_FILE)
+    tool_img_ver = extract_latest_version(version_file)
+    tool_img_tag = f'{tool}_{tool_img_ver}'
+    image_str = f'{registry}:{tool_img_tag}'
 
     if not force_pull:
```
MrCreosote, line 124: It seems like this code is never executed since `force_pull` is always `True`. To my mind this code should always be executed, right?

It also seems like the force_pull arg could be removed since it's always the same, unless there's some reason to keep it I'm not seeing.

Tianhao-Gu: removed unused codes. I think we should always run `shifterimg pull`.

Tianhao-Gu: ok. reversed changes.

MrCreosote: I'd still like to understand why you thought we should pull the image every time, maybe I'm missing something

Tianhao-Gu: I just thought it would be safer to pull every time. And this operation is not very time consuming.

MrCreosote: Why would it be safer? (I'm not trying to be argumentative here, just in case I'm coming off that way, genuinely want to understand) 

If it's cheap then maybe pulling out all that code is the right way to go

Tianhao-Gu: I think it's safer because it's pulling every time.

MrCreosote: right, but at the risk of sounding like a broken record, why is it safer to pull every time if the image has a specific tag that doesn't change?

Tianhao-Gu: Yea. I think what you said makes sense. We are using a specific tag.

MrCreosote (approved):

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: Fair enough

PR #394, opened 2023-07-30
RE2022-218: move checkm2 parsing logic to tool computation

Tianhao-Gu:
moving pre-processing logic to container.


src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -37,8 +46,28 @@ def _run_checkm2(
...
+
+    _process_checkm2_result(output_dir, meta)
+
+
+def _process_checkm2_result(
+        output_dir: Path,
+        meta: Dict[str, str],
+):
+    # create metadata file before parsing the result
+    _create_metadata_file(meta, output_dir)
```
MrCreosote, line 60: The reason the metadata file was written after the tool execution finished was so that if tool execution failed partway it'd be clear to the parser that only partial results were available and it should fail. Since the jsonl files will now serve that purpose the meta files could be written out prior to the tool execution and you wouldn't have to pass the meta info into the tools, I think.

MrCreosote: Alternatively, if the meta file is only needed for this step, it looks like from the parsing code all you need is the tool identifier -> data-id mapping, so maybe you could just pass that around rather than creating a file.

src/loaders/compute_tools/tool_result_parser.py
MrCreosote, line 1: noop test file?

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if selected_cols and genome_id_col:
+        selected_cols.add(genome_id_col)
+
+    df = pd.read_csv(file_path, sep='\t', keep_default_na=False, usecols=selected_cols)
+
+    return df
+
+
+def _create_meta_lookup(batch_result_dir):
```
MrCreosote, line 112: Seems to be unused?

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -0,0 +1,131 @@
...
+import pandas as pd
+
+from src.loaders.common import loader_common_names
+
+# used as collection_and_field_names.FLD_KBASE_ID
+TOOL_KBASE_ID = "kbase_id"
+TOOL_GENOME_ATTRI_FILE = "genome_attribs.jsonl"
+
+
+def _create_metadata_file(
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍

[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_common.py
```diff
@@ -367,9 +365,6 @@ def parallel_batch_execution(self, tool_callable: Callable[[Dict[str, GenomeTupl
                 for file in unzipped_files_to_delete:
                     os.remove(file)
 
-        for meta in metas:
-            _create_metadata_file(*meta)
```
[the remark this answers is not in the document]
Tianhao-Gu: yea, gtdb needs this. I am planning to address those in the next PRs when I am at gtdb and non-batch tools. The containers for those tools are not updated so nothing will break.

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -11,13 +10,23 @@
     ToolRunner,
     run_command,
 )
+from src.loaders.compute_tools.tool_result_parser import (
+    _read_genome_attri_result,
+    _create_jsonl_files,
+    _create_metadata_file, TOOL_GENOME_ATTRI_FILE,
```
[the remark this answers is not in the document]

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -11,13 +10,23 @@
...
+# If empty, select all available fields
+SELECTED_CHECKM2_FEATURES = {'Completeness', 'Contamination'}
 
 
 def _run_checkm2(
         ids_to_files: Dict[str, GenomeTuple],
         output_dir: Path,
         threads: int,
         debug: bool,
+        meta: Dict[str, str],
```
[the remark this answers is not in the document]

src/loaders/compute_tools/tool_result_parser.py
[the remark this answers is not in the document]
Tianhao-Gu: yea. those are pretty much C&P.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: I have to write the metadata after tool execution because checkm2 needs a clean directory to execute. The metadata file is needed for other things like compute failure ids  as well.

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: 👍 removed.

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: I see, ok.

In that case, after reading through things a bit more, it looks like the reason the checkm2 code needs the metadata file is just to get the tool identifier -> data id mapping. However, that mapping is already available in the `ids_to_files` argument. So what I'd do here is just pass the mapping to `read_genome_attri_result` (rather than parsing the metadata file in the method) and leave the metadata file generation where it was.

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_result_parser.py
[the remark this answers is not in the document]
MrCreosote: :+1:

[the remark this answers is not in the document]
MrCreosote: Just to be clear, the other implementations of these methods will be deleted in a future PR (probably the gtdb_tk PR)?

src/loaders/compute_tools/checkm2/checkm2.py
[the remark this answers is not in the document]

src/loaders/compute_tools/checkm2/checkm2.py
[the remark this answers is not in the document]
MrCreosote: I was talking about `tool_callable` in `arallel_batch_execution` but based on a different comment I think this new argument might go away

src/loaders/compute_tools/checkm2/checkm2.py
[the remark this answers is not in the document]

src/loaders/compute_tools/tool_result_parser.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: what's wrong with keeping the existing logic and making minimum changes to code base? We need to generate the metadata file down the road anyways.

src/loaders/compute_tools/tool_result_parser.py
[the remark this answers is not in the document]
Tianhao-Gu: Are you referring to `_create_metadata_file` method? I made it public in tool_common and remove the rest of implementations.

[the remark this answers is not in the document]
MrCreosote: I'm talking about the stuff in `tool_result_parser.py`

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: What I'm talking about would reduce changes to the code base AFAICT, and it would mean someone implementing a new tool doesn't have to worry about writing the metadata file. 

Maybe we should huddle quickly on this one if we're still in disagreement

src/loaders/compute_tools/tool_result_parser.py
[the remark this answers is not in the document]
Tianhao-Gu: yea those duplicate methods will be removed.

src/loaders/compute_tools/tool_common.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/compute_tools/checkm2/checkm2.py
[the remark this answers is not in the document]

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: 👍 updated to use `ids_to_files`

src/loaders/compute_tools/tool_result_parser.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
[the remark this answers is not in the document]
MrCreosote: moot now

src/loaders/compute_tools/tool_common.py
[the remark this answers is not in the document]
MrCreosote: moot now

src/loaders/compute_tools/tool_common.py
```diff
@@ -368,13 +368,13 @@ def parallel_batch_execution(self, tool_callable: Callable[[Dict[str, GenomeTupl
...
         for meta in metas:
-            _create_metadata_file(*meta)
+            create_metadata_file(*meta)
 
     def _execute(
             self,
             threads: int,
             tool_callable: Callable[..., None],
-            args: List[Tuple[Any]],
+            args: List[Tuple[Any, ...]],
```
MrCreosote, line 377: I think this is cruft

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -0,0 +1,105 @@
...
+from pathlib import Path
+from typing import Dict, List, Set
+
+import jsonlines
+import pandas as pd
+
+from src.loaders.compute_tools.tool_common import GenomeTuple
+
+# used as collection_and_field_names.FLD_KBASE_ID
+TOOL_KBASE_ID = "kbase_id"
```
MrCreosote, line 12: Shouldn't this be imported from `collection_and_field_names`?

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    process the output file generated by the tool (checkm2, gtdb-tk, etc) to create a format suitable for
+    importing into ArangoDB
+
+    :param batch_result_dir: the directory where the tool result files are stored
+    :param tool_file_name: the name of the tool result file
+    :param features: a list of features to be retrieved from the tool result file
+    :param genome_id_col: the column name of the genome id in the tool result file
+    :param ids_to_files: a dictionary of tool genome ids to the corresponding GenomeTuple
+    :param prefix: the prefix to append to the genome id from the tool result file
```
MrCreosote, line 46: do you mean the prefix to prepend?

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -0,0 +1,105 @@
...
+
+    :return: a dictionary of genome attributes
+    """
+    tool_genome_map = {tool_id: genome_tuple.data_id for tool_id, genome_tuple in ids_to_files.items()}
+
+    tool_file = os.path.join(batch_result_dir, tool_file_name)
+    docs = list()
+    if os.path.exists(tool_file):
```
MrCreosote, line 54: I'm sure we discussed this before but silently ignoring missing files makes me nervous. What was the reasoning behind this again?

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -0,0 +1,105 @@
...
+    return docs
+
+
+def _row_to_doc(row, features, tool_genome_map, genome_id_col, prefix):
+    # Transforms a row from tool result file into ArangoDB collection document
+
+    try:
+        genome_id = tool_genome_map[row[genome_id_col]]
+    except KeyError as e:
+        raise ValueError('Unable to find genome ID') from e
```
MrCreosote, line 69: When can this happen?

MrCreosote (commented):
just about done I think

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: I need the logic compatible with python 3.8. I cannot import collection_and_field_names

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: raise error if file doesn't exist. I copy the logic from parse_tool_result.py. I don't remember exactly, but it's probably due to gtdb_tk can produce both or either result files. I think I will just make sure when it's used by gtdb_tk, I will check the result file existence.

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: hmmm maybe it's just a safe check. would you rather not having this? It was inherited from the parse_tool_result.py too.

MrCreosote: If we're sure it can't happen we should remove it, but if we're not sure it seems helpful to keep it

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: Oh right, crap. Ok, maybe we should separate the main file from the collection annotations. If you want to do that here that's fine, otherwise could you make an issue

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: Maybe add a bool flag to determine whether to throw an error if the file is missing or not. Then it's clear from the signature what's going to happen

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: 👍 add another AC in the ticcket. https://kbase-jira.atlassian.net/browse/RE2022-218

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: it doesn't hurt. I think I will just keep it.

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_result_parser.py
MrCreosote: We need to be a bit careful about this though - the reason I put it all in one place was so the CLI could pull the collection information easily from the file. If we split the collection names from the annotations we might forget to add the annotations

MrCreosote (approved):

PR #402, opened 2023-08-04
RE2022-218: refactor gtdb_tk parsing logic

Tianhao-Gu:


src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -122,14 +131,37 @@ def _run_gtdb_tk(
...
+):
+    genome_id_col = loader_common_names.GTDB_GENOME_ID_COL
+    gtdb_tk_docs = list()
+    for tool_file_name in summary_files:
+        docs = read_genome_attri_result(
+            output_dir,
+            tool_file_name,
+            SELECTED_GTDBTK_SUMMARY_FEATURES,
+            genome_id_col,
+            ids_to_files)
```
MrCreosote, line 158: Does this need to set the check file flag to false since both summary files aren't necessarily there?

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -122,14 +131,37 @@ def _run_gtdb_tk(
...
         error_message = filtered_or_failed_genome_id_mapping[miss_tool_safe_data_id]
         fatal_tuple = create_fatal_tuple(miss_tool_safe_data_id, ids_to_files, error_message)
         fatal_tuples.append(fatal_tuple)
 
     write_fatal_tuples_to_dict(fatal_tuples, output_dir)
 
+    _process_gtdb_result(output_dir, ids_to_files, summary_files)
+
+
+def _process_gtdb_result(
```
MrCreosote, line 145: this code looks really similar to the checkm2 code. Can it be shared?

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Tianhao-Gu: I looked at the code again. The func that returns gtdb_tk result files actually will only return the ones generated. So the file will guaranteed to be there. 
https://github.com/kbase/collections/blob/main/src/loaders/compute_tools/tool_common.py#L604-L607
Do you want me to remove the check flag or just leave it there? It doesn't hurt to just leave it there I think.

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
Tianhao-Gu: 👍

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: I think it's fine as is

src/loaders/compute_tools/tool_result_parser.py
```diff
@@ -12,6 +12,45 @@
...
+    :param genome_id_col: the column name of the genome id in the tool result file
+    :param ids_to_files: a dictionary of tool genome ids to the corresponding GenomeTuple
+    :param result_files: the list of tool result files
+    :param check_file_exists: if True, check if the tool result file exists
+    :param prefix: the prefix of the tool result file
+    """
+
+    genome_attri_docs = list()
+    for tool_file_name in result_files:
+        docs = read_genome_attri_result(
```
MrCreosote, line 39: I think you can just inline this method

Tianhao-Gu: 👍

MrCreosote: By inline I mean incorporate the code into the calling method and just not have a 2nd method

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_result_parser.py
Tianhao-Gu: 👍

MrCreosote (approved):

PR #403, opened 2023-08-04
RE2022-218: separate field name used by parser

Tianhao-Gu:


MrCreosote (approved):
Not what I was thinking exactly, but this'll work

PR #411, opened 2023-08-08
RE2022-218: update genome attribs global parser

Tianhao-Gu:


src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -476,14 +435,10 @@ def _process_genome_attri_tools(genome_attr_tools: set[str],
...
     docs = list()
     for tool in genome_attr_tools:
-        try:
-            parse_ops = getattr(sys.modules[__name__], tool)
-        except AttributeError as e:
-            raise ValueError(f'Please implement parsing method for: [{tool}]') from e
-
-        docs.extend(parse_ops(root_dir, env, kbase_collection, load_ver, fatal_ids))
-
+        tool_docs = _retrieve_pre_processed_docs(tool, root_dir, env, kbase_collection, load_ver, fatal_ids)
```
MrCreosote, line 438: Does this mean we could add more tools to the parser whose output ends up in genome attribs just by adding the tool name to the list of GA tools?

MrCreosote (approved):

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: yea. That's the idea. We can include additional tools and pack results together given we pre-process the tool result similar to checkm2/gtdb_tk.

PR #412, opened 2023-08-09
RE2022-218: save jsonl trait count

Tianhao-Gu:


MrCreosote:
So I was thinking the tool would build the cell and row documents like the parser currently does, not just save the csv data as jsonl instead

Tianhao-Gu:
> So I was thinking the tool would build the cell and row documents like the parser currently does, not just save the csv data as jsonl instead. That doesn't really do anything helpful since it's just an intermediate format IIUC

It cannot. We need the whole collection to be finished before compute heatmap cell/rows.

MrCreosote:
Why is that?

MrCreosote:
> cell needs to have a global id

We could do that by making it a UUID or `<kbase_id>_<per row incrementing integer id>`

> we want to make tool container light

While true, I don't think adding the data transformation code we're talking about is adding a lot of weight. I was thinking more of things like adding more 3rd party tools, large python libs, etc.

Worst case we could set things up so that a post processing script ran on the compute nodes after the tool container (or we could containerize that as well) but at least naively I don't think that's an issue here.

> we want to compute max/min value for each row, etc.

It's for all rows, not per row right? But that we would have to keep in the parser, yes. That being said, it'd be easy - just load each row doc and run through the cells. If we start storing the results in the DB it could be done with an `AGGREGATE` AQL call

Tianhao-Gu:
> > cell needs to have a global id
> 
> We could do that by making it a UUID or `<kbase_id>_<per row incrementing integer id>`
> 
> > we want to make tool container light
> 
> While true, I don't think adding the data transformation code we're talking about is adding a lot of weight. I was thinking more of things like adding more 3rd party tools, large python libs, etc.
> 
> Worst case we could set things up so that a post processing script ran on the compute nodes after the tool container (or we could containerize that as well) but at least naively I don't think that's an issue here.
> 
> > we want to compute max/min value for each row, etc.
> 
> It's for all rows, not per row right? But that we would have to keep in the parser, yes. That being said, it'd be easy - just load each row doc and run through the cells. If we start storing the results in the DB it could be done with an `AGGREGATE` AQL call

alright. It's your call. But we are adding intermediate steps, introducing unnecessary variables/dependencies to break a global operation into parts.

MrCreosote:
> But we are adding intermediate steps, introducing unnecessary variables/dependencies to break a global operation into parts.

I'm not quite sure I understand this. If there's negative repercussions to doing this that I'm not getting, maybe we should do a quick zoom and talk this out

PR #414, opened 2023-08-11
Update dependencies

MrCreosote:
Note that `fastapi` was not updated, as that bumps `pydantic` from v1 to v2, which has a very large number of breaking changes. That will need to be updated in its own PR and tested extensively


Tianhao-Gu (approved):
LGTM

PR #418, opened 2023-08-15
RE2022-232: Add metadata per heatmap row for biolog media

MrCreosote:
Could be used for things other than the media in the future


MrCreosote:
image
image

MrCreosote:
> Though is there a more descriptive name than "media"?

That's the standard term in wetlab world afaik. We could be more specific with "growth_media" maybe?

Tianhao-Gu (approved):
LGTM

PR #425, opened 2023-08-23
Add genome attributes design

MrCreosote:


design/genome_attributes_filtering.md
```diff
@@ -0,0 +1,54 @@
...
+      * No need for this currently AFAWK
+    * Filter strategies
+      * prefix search
+      * full text search
+        * requires ArangoSearch View or inverted index
+      * substring search
+        * cannot use index without ArangoSearch View or (maybe?) inverted index
+          * TBD - will an inverted index work?
+        * Out of scope for v1
+  * Enum
```
Tianhao-Gu, line 26: enum includes both list and dictionary type?

MrCreosote: Not sure what you mean - an enum is a set of strings

MrCreosote: I still have no idea what you mean - there aren't any dictionary values for enums

dauglyon (approved):

design/genome_attributes_filtering.md
MrCreosote: I don't think there is a dictionary column type. Which column were you thinking of?

Tianhao-Gu: currently I only see `_upas`.

MrCreosote: Those aren't visible to the user, they're for the service only

MrCreosote: https://github.com/kbase/collections/blob/main/src/service/data_products/genome_attributes.py#L212

Tianhao-Gu: 👍

Tianhao-Gu (approved):
👍

Issue #428, opened 2023-08-24
Update `example` to `examples` in pydantic models when Swagger fixes bug


MrCreosote:
`example` is now deprecated in FastAPI but the replacement, `examples`, does not show up in the Swagger docs, so we can't switch to it for now.

https://github.com/swagger-api/swagger-ui/issues/9045

PR #429, opened 2023-08-24
Update fastapi & pydantic

MrCreosote:
Pydantic v2 has a number of breaking changes.

Also see https://github.com/kbase/collections/issues/428


src/service/data_products/common_models.py
```diff
@@ -82,10 +81,10 @@ class DataProductMissingIDs(SubsetProcessStates):
     )
 
 
-QUERY_VALIDATOR_LOAD_VERSION_OVERRIDE = Annotated[str | None, Query(
+QUERY_VALIDATOR_LOAD_VERSION_OVERRIDE = Annotated[str, Query(
```
MrCreosote, line 84: For `Path` and `Query` components, adding `None` to the type annotation causes the input restrictions not to be shown in the OpenAPI documentation. Not sure why

Tianhao-Gu (approved):
the changes look good. But did you test tool runners with pydantic 2? I know at least microtrait is explicitly using `pydantic=1.10.12`.

MrCreosote:
I didn't think of that, but looking at the code changes, they shouldn't break anything since microtrait doesn't actually use the models any more, just the field names. Therefore the imports just need to work.

Would it be ok if I asked you to bump pydantic next time you work on the tools? I've literally never run them so testing them is a bit of a pain for me currently (which at some point I need to fix)

Tianhao-Gu (approved):
LGTM

src/service/data_products/common_models.py
bio-boris: Does it happen with a Optional[Str]  Or Union[] ?

MrCreosote: I didn't try because the current way works well enough for now and I don't want to go back to the Bad Old Ways of type hints

PR #431, opened 2023-08-25
Fix bugs related to pydantic 2 conversion

MrCreosote:
In some cases optional fields need to be Annotated and explicitly assigned, and some cases they don't. I'm not quite sure which is which yet.


Tianhao-Gu (approved):
LGTM

PR #433, opened 2023-08-28
RE2022-239: Add genome attribs metadata endpoint to service

MrCreosote:
also improve range validation

### DIFF
--- a/src/service/data_products/genome_attributes.py
+++ b/src/service/data_products/genome_attributes.py
@@ -22,6 +22,7 @@ from src.service.data_products.common_functions import (
     remove_marked_subset,
     override_load_version,
     query_table,
+    get_collection_singleton_from_db,
 )
 from src.service.data_products.data_product_processing import (
     MATCH_ID_PREFIX,
@@ -34,6 +35,7 @@ from src.service.routes_common import PATH_VALIDATOR_COLLECTION_ID
 from src.service.storage_arango import ArangoStorage, remove_arango_keys
 from src.service.timestamp import now_epoch_millis
 from typing import Any, Callable
+from src.common.product_models import columnar_attribs_common_models as col_models
 
 # Implementation note - we know FLD_KBASE_ID is unique per collection id /
 # load version combination since the loader uses those 3 fields as the arango _key
@@ -173,6 +175,10 @@ GENOME_ATTRIBS_SPEC = GenomeAttribsSpec(
     router=_ROUTER,
     db_collections=[
         common_models.DBCollection(
+            name=names.COLL_GENOME_ATTRIBS_META,
+            indexes=[]  # lookup is by key
+        ),
+        common_models.DBCollection(
             name=names.COLL_GENOME_ATTRIBS,
             indexes=[
                 [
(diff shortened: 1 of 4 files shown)


src/service/data_products/genome_attributes.py
```diff
@@ -222,18 +228,51 @@ def _remove_keys(doc):
...
+    storage = app_state.get_app_state(r).arangostorage
+    _, load_ver = await get_load_version(storage, collection_id, ID, load_ver_override, user)
+    doc = await get_collection_singleton_from_db(
+            storage,
+            names.COLL_GENOME_ATTRIBS_META,
+            collection_id,
+            load_ver,
+            bool(load_ver_override)
+    )
+    doc[col_models.FIELD_COLUMNS] = [col_models.AttributesColumn.model_construct(**d)
```
dauglyon, line 254: Is this always ordered (ie is this list always in the same order as the columns in the table?)

dauglyon: I'm kinda confused on if this func returns a dict or a list, I think I need to learn more about the pydantic lib to parse this PR

Xiangs18, line 254: cannot find model_construct method. Is this for later PR?

MrCreosote: `model_construct` is a pydantic method

MrCreosote: `model_construct` constructs an instance of the model class upon which it's called, in this case `AttributesColumn`

Xiangs18: I know `doc` return by get_collection_singleton_from_db is a dict. Can you provide a sample doc? Would like to know more what keys/values are inside.

MrCreosote: No samples docs exist currently; doc creation will need to be added to the pipeline parser. Also the contents of the doc will be different depending on the arango collection specified in the arguments. For this case, though, you can see what fields the doc will need to have by examining the model that will be constructed: https://github.com/kbase/collections/blob/main/src/common/product_models/columnar_attribs_common_models.py

Xiangs18: doc["columns"] will be a list of col names or dict?

MrCreosote: https://github.com/kbase/collections/blob/main/src/common/product_models/columnar_attribs_common_models.py#L106

Xiangs18: The part I am trying to understand is L254-255. Maybe I confuse myself.

You already had a list of AttributesColumn and why you construct them again and reassign to doc["columns"]. M

MrCreosote: What's coming back from the database is just dicts and lists. The response model defined for this method is a `ColumnarAttributesMeta`, and so L254-255 just construct that response model from the dict/list data structure. L254 is necessary because `model_construct` is not recursive, so any enclosed models need to be explicitly created.

Xiangs18 (approved):
LGTM

dauglyon (approved):

PR #447, opened 2023-09-05
RE2022-240: Update filtering design doc

MrCreosote:
* Use arangosearch for all filtering ops
* Note n-gram might be useful for substring search
* Note decision to not add any infrastructure for updating indexes / views


Tianhao-Gu (approved):
LGTM.

PR #449, opened 2023-09-06
RE2022-240: Add timings for n-gram substring search

MrCreosote:
Currently no major difference between `LIKE()` (which requires an index scan) and n-gram search, probably due to the DB size. However, n-gram search handles tokens in the search string while `LIKE()` can't.


Tianhao-Gu (approved):
👍

PR #450, opened 2023-09-07
RE2022-240: Add range filtering representation code

MrCreosote:
Parses the input range filter specification and translates to ArangoSearch AQL.

### DIFF
--- a/src/service/data_products/filtering.py
+++ b/src/service/data_products/filtering.py
@@ -1,0 +1,167 @@
+"""
+Data structures and methods for parsing, representing, and translating table filters for
+data products like genome attributes or samples.
+"""
+
+from dateutil import parser
+from pydantic import BaseModel, Field
+from src.common.product_models.columnar_attribs_common_models import ColumnType
+
+from types import NotImplementedType
+from typing import Any, Self
+
+
+class SearchQueryPart(BaseModel):
+    aql_lines: list[str] = Field(  # TDOO better docs once I figure out integration
+        description="One or more lines of ArangoSearch AQL representing the filter")
+    bind_vars: dict[str, Any] = Field(
+        description="The bind variables for the AQL lines")
+
+
+def _to_bool_string(b: bool):
+    return "true" if b else "false"
+
+
+class RangeFilter:
+    
+    def __init__(
+        self,
+        type_: ColumnType,
+        low: float | int | str | None = None,
+        high: float | int | str | None = None,
+        low_inclusive: bool = False,
+        high_inclusive: bool = False
+    ):
+        """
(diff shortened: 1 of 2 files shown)


src/service/data_products/filtering.py
```diff
@@ -0,0 +1,167 @@
...
+        return ret
+
+    @classmethod
+    def _parse_inclusivity(self, part, start):
+        inclusive = False
+        if start:
+            if part.startswith("["):
+                inclusive = True
+            if part.startswith("(") or part.startswith("["):
+                part = part[1:]
```
Tianhao-Gu, line 83: raise if range string is not None and doesn't start/end with `[` or `(`?

src/service/data_products/filtering.py
```diff
@@ -0,0 +1,167 @@
...
+    def __repr__(self):
+        l, h = self.low, self.high
+        if self.type == ColumnType.DATE:
+            l = f'"{l}"'
+            h = f'"{h}"'
+        return (f"RangeFilter(ColumnType.{self.type.name}, {l}, {h}, "
+            + f"{self.low_inclusive}, {self.high_inclusive})")
+
+    @classmethod
+    def from_string(cls, type_: ColumnType, string: str) -> Self:
```
Tianhao-Gu, line 121: we didn't do any validation for `string`. Should we validate it before calling filter? i.e. If I pass `<2, 3)`, it will raise at `_parse_val` but we could raise earlier.

src/service/data_products/filtering.py
```diff
@@ -0,0 +1,167 @@
...
+                part = part[1:]
+        else:
+            if part.endswith("]"):
+                inclusive = True
+            if part.endswith(")") or part.endswith("]"):
+                part = part[:-1]
+        return part, inclusive
+    
+    def _parse_val(self, val, name):
+        if val is None or val == "":
```
Tianhao-Gu, line 92: when will `val == ""` happen? `string = '(1, )'`?

src/service/data_products/filtering.py
MrCreosote: The string specification says that `(` and `)` can be omitted without changing the semantics

Tianhao-Gu: 👍

src/service/data_products/filtering.py
Tianhao-Gu: got it. 👍

src/service/data_products/filtering.py
Tianhao-Gu: 👍

Tianhao-Gu (approved):
LGTM

PR #451, opened 2023-09-07
RE2022-240: Add a string filter representation class

MrCreosote:
Represents a filter for a string column and translates to ArangoSearch AQL lines.

### DIFF
--- a/src/service/data_products/filtering.py
+++ b/src/service/data_products/filtering.py
@@ -53,7 +53,7 @@ class AbstractFilter(ABC):
             `f"{identifer} > 47`
         var_prefix - a prefix to apply to variable names, including bind variables,
             to prevent collisions between multiple filters.
-        analyzer - the analyzer to use for processing user input strings.
+        analyzer - the analyzer to use for the search.
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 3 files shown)


test/src/service/data_products/filtering_test.py
```diff
@@ -68,37 +77,79 @@ def test_to_arangosearch_aql():
...
+
+def test_rangefilter_from_string_fail():
     test_set = (
-        (ColumnType.INT, "   \t   ", "^Missing range information$"),
+        (ColumnType.INT, "   \t   ", "Missing range information"),
         (
             ColumnType.INT,
             " [  \t   ",
-            "^Invalid range specification; expected exactly one comma:  \\[  \t   $"
+            "Invalid range specification; expected exactly one comma:  \\[  \t   "
```
Xiangs18, line 89: There are a bunch of  `\\` inside test_rangefilter_from_string_fail function. Why are they?

MrCreosote: Regex escape sequences

Xiangs18: Oh, I see. But why not using "\"? You also want to match a single "\" in your value as well?

MrCreosote: Python complains if you escape an invalid escape character:

```
  /home/crushingismybusiness/github/kbase/collections/test/src/service/data_products/filtering_test.py:89: DeprecationWarning: invalid escape sequence '\['
```

src/service/data_products/filtering.py
```diff
@@ -3,26 +3,69 @@
...
-from src.common.product_models.columnar_attribs_common_models import ColumnType
+from src.common.product_models.columnar_attribs_common_models import ColumnType, FilterStrategy
 
 from types import NotImplementedType
-from typing import Any, Self
+from typing import Annotated, Any, Self
 
 
 class SearchQueryPart(BaseModel):
+    variable_assignments: Annotated[dict[str, str], Field(
```
Xiangs18, line 16: should be Annotated[dict[str, str] | `None`, Field(...)] = None?

Xiangs18:
Somehow I cannot left a comment there, but [this](https://github.com/kbase/collections/pull/451/files#diff-9fbc55937877d77ed5cb163f16993b7fdaef56d0757c8f3dff12b22c82e141f8L76) looks like a @staticmethod rather than @classmethod

src/service/data_products/filtering.py
MrCreosote: Fixed

MrCreosote:
> change _parse_inclusivity(self...) -> _parse_inclusivity(cls...)

I think it originally was a regular method, but when I switched it to a classmethod I forgot to switch the self var. Anyway, fixed

src/service/data_products/filtering.py
```diff
@@ -3,26 +3,69 @@
...
+    def from_string(self, type_: ColumnType, strategy: FilterStrategy, string: str) -> Self:
+        """
+        Parse the filter from a filter string. The syntax of the filter string is dependent
+        on the filter implementation.
+        
+        type_ - the column type to which the filter will apply.
+        strategy - the strategy for the filter.
+        string - the string to parse to create the filter.
+        """
+        raise NotImplementedError()
```
Xiangs18, line 43: Ah, not sure about this line. Why we need to raise NotImplementedError()? 

I thought a TypeError will be automatically raised if any of abstract methods is not implemented.

MrCreosote: Just to make it clear to the subclass implementer that there's no need to call the super method

Xiangs18 (approved):
LGTM!

PR #454, opened 2023-09-14
RE2022-240: Add load_ver, skip, and limit to arangosearch AQL generator

MrCreosote:
Needed prior to any integration into the service

### DIFF
--- a/test/src/service/data_products/filtering_test.py
+++ b/test/src/service/data_products/filtering_test.py
@@ -11,7 +11,8 @@ from src.common.product_models.columnar_attribs_common_models import (
 from src.service.data_products.filtering import (
     RangeFilter,
     StringFilter,
-    SearchQueryPart, FilterSet
+    SearchQueryPart,
+    FilterSet
 )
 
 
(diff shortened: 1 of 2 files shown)


test/src/service/data_products/filtering_test.py
```diff
@@ -190,9 +191,14 @@ def test_filterset_w_defaults():
         ANALYZER(v4_prefixes ALL == doc.fulltextfield, "text_rs")
         AND
         doc.datefield <= @v5_high
+        LIMIT @skip, @limit
         RETURN doc
 """.strip() + "\n"
     assert bind_vars == {
+        "@view": "my_search_view",
```
Xiangs18, line 198: why var view has an extra @?

MrCreosote: I'm actually not sure of the reasoning, it's not explained: https://www.arangodb.com/docs/stable/aql/fundamentals-bind-parameters.html#syntax

Xiangs18: > A special type of bind parameter exists for injecting collection names. This type of bind parameter has a name prefixed with an additional @ symbol, so @@name in the query.

I see. Make sense now!

Xiangs18 (approved):
LGTM!

PR #455, opened 2023-09-14
RE2022-240: Add simple dynamic configuration system

MrCreosote:
Needed to configure the arangosearch view name for filtering operations

Also adds a `__len__` method to `FilterSet`


src/service/config_dynamic.py
```diff
@@ -0,0 +1,47 @@
...
+        # I think setting the last update to prior to the big bang is probably safe
+        self._last = -14000000000 * 365 * 24 * 60 * 60
+        self._timestamp = timestamp_fn
+    
+    async def get_config(self) -> DynamicConfig:
+        """ Get the service config """
+        ts = self._timestamp()
+        if ts > self._last + self._cache_for_sec:
+            self._config = await self._db.get_dynamic_config()
+            self._last = ts
```
Tianhao-Gu, line 46: should we do `self._last = time.time()` here? In case `get_dynamic_config` takes long. If `get_dynamic_config` takes longer than `_cache_for_sec`, we will never get new service config.

Tianhao-Gu (approved):
LGTM other than the comment regarding `self._last`.

src/service/config_dynamic.py
MrCreosote: I'm not sure I understand. if `get_dynamic_config` takes longer than the cache time (which seems really unlikely), then `last + cache` will always be < than the current time, so the config will be pulled from the db on every request

Tianhao-Gu: I see. yea. you are right. If `get_dynamic_config` takes long, it will always pull from DB. I misread the `>`.

Tianhao-Gu:
test failed on `codecov`. But I think that should be ok.

PR #456, opened 2023-09-14
Make test file names consistent

MrCreosote:
All test files except these 3 are named *_test. Make these the same


Tianhao-Gu (approved):
👍

PR #458, opened 2023-09-15
RE2022-240: Move filtering code into own package

MrCreosote:
Going to need to add some more modules, makes sense to make it non-data product specific


Tianhao-Gu (approved):
👍

PR #461, opened 2023-09-18
RE2022-240: Add identity string filter strategy

MrCreosote:
Will want to do exact matches on some columns without stemming, particlarly `kbase_id`


Xiangs18 (approved):
LGTM!

PR #463, opened 2023-09-19
Ensure min / max are ints for biolog

MrCreosote:
Since biolog only has bools, need to covert them to ints. This bug was hidden in microtrait since there are always 0 and 1+ values in the dataset.

Also fix some warnings when building return models


Pipfile
```diff
@@ -43,6 +43,7 @@ pymongo = "4.4.1"
 # Note aioarango forces an old version of httpcore
 dnspython = "==2.3.0"
 sourmash = "==4.8.3"
+prometheus-client = "==0.17.1"
```
MrCreosote, line 46: This is useful for parsing the arango metrics data

Tianhao-Gu (approved):
👍

PR #464, opened 2023-09-19
RE2022-240: Add specs for genome attribs columns & loader

MrCreosote:


src/common/collection_column_specs/genome_attribs-GROW.yml
```diff
@@ -0,0 +1 @@
+genome_attribs-PMI.yml
```
MrCreosote, line 1: softlink

src/common/collection_column_specs/genome_attribs-GTDB.yml
```diff
@@ -0,0 +1,160 @@
...
+   - key: checkm_completeness
+     type: float
+
+   - key: checkm_contamination
+     type: float
+
+   - key: checkm_marker_count
+     type: int
+
+   - key: checkm_marker_lineage  # GTDB node, might need better tokentizing or substring search
```
Tianhao-Gu, line 19: * tokenizing

Tianhao-Gu (commented):
spec and code changes look good to me. Do you think it's a good idea to add a note for `SELECTED_FEATURES` to remind us if we make any changes to selected features should prompt corresponding modifications to the spec file?

MrCreosote:
> Do you think it's a good idea to add a note for SELECTED_FEATURES to remind us if we make any changes to selected features should prompt corresponding modifications to the spec file?

Maybe it should be the other way around and the specs determine the selected features, possibly with a mapping from the original name to a new name? And If the output is missing a key an error is thrown?

src/common/collection_column_specs/genome_attribs-GTDB.yml
MrCreosote: fixed

Tianhao-Gu:
> > Do you think it's a good idea to add a note for SELECTED_FEATURES to remind us if we make any changes to selected features should prompt corresponding modifications to the spec file?
> 
> Maybe it should be the other way around and the specs determine the selected features, possibly with a mapping from the original name to a new name? And If the output is missing a key an error is thrown?

specs determines SELECTED_FEATURES makes more sense to me. But this PR LGTM.

Tianhao-Gu (approved):

PR #469, opened 2023-09-21
RE2022-240: Merge column specs for collection data products

MrCreosote:


src/common/collection_column_specs/load_specs.py
```diff
@@ -10,13 +10,34 @@
 """
 import yaml
 from pathlib import Path
-from src.common.product_models.columnar_attribs_common_models import ColumnarAttributesSpec
-
+from src.common.product_models.columnar_attribs_common_models import (
+    ColumnarAttributesSpec,
```
Xiangs18, line 14: Should be alphabetical order

src/common/collection_column_specs/load_specs.py
```diff
@@ -10,13 +10,34 @@
...
+    AttributesColumnSpec,
```
Xiangs18, line 15: Should be alphabetical order

MrCreosote: fixed

src/common/collection_column_specs/load_specs.py
```diff
@@ -26,8 +47,29 @@ def load_spec(data_product: str, collection: str = None) -> ColumnarAttributesSp
...
+    columns = []
+    for k in keys:
+        first = None
+        for coll, spec in coll2spec.items():
+            if k in spec:
+                if not first:
+                    first = (coll, spec[k])
+                else:
+                    if spec[k] != first[1]:
+                        colls = sorted([coll, first[0]])
```
Xiangs18, line 71: Not quite sure why you are sorting here. Shouldn't raise error directly?

MrCreosote: If I don't sort, then which coll name goes where in the error message is indeterminate, and the test passing is therefore indeterminate. The lack of a sort is why the test were passing locally but not in GHA

Xiangs18 (approved):
LGTM!

PR #470, opened 2023-09-22
RE2022-240: Add global columns to column specs

MrCreosote:
Def need to index the coll, load_ver, and _mtchsel columns in arangosearch

### DIFF
--- a/src/service/filtering/analyzers.py
+++ b/src/service/filtering/analyzers.py
@@ -36,9 +36,16 @@ _CUSTOM_ANALYZERS = {
     )
 }
 
-def get_analyzer(strategy: FilterStrategy | None) -> str:
-    """ Get the name of the appropriate ArangoDB analyzer to use for a given filter strategy. """
-    return _COL2ANALYZER.get(strategy, DEFAULT_ANALYZER)
+def get_analyzer(strategy: FilterStrategy | None, return_none_for_default_analyzer: bool = False
+    ) -> str:
+    """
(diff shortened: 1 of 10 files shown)


src/service/filtering/analyzers.py
```diff
@@ -36,9 +36,16 @@
...
-    return _COL2ANALYZER.get(strategy, DEFAULT_ANALYZER)
+def get_analyzer(strategy: FilterStrategy | None, return_none_for_default_analyzer: bool = False
+    ) -> str:
+    """
+    Get the name of the appropriate ArangoDB analyzer to use for a given filter strategy.
+    
+    strategy - the strategy to translate to an analyzer
+    return_none_for_default_analyzer - rather than returning the default analyzer, return None.
+    """
+    df = None if return_none_for_default_analyzer else DEFAULT_ANALYZER
```
Tianhao-Gu, line 47: can we rename `df` to `default_analyzer`?  It's a common convention to use df for dataframes. Or we can just do `return _COL2ANALYZER.get(strategy, None if return_none_for_default_analyzer else DEFAULT_ANALYZER)`

Tianhao-Gu (approved):
LGTM other than a potentially misleading variable name.

src/service/filtering/analyzers.py
MrCreosote: yeah... if pandas were used in this module I'd agree, but not using any variables that are frequently used in entirely different contexts doesn't seem reasonable to me. Means we shouldn't use np either, for example.

PR #472, opened 2023-09-29
RE2022-240: Move and rename init_storage script

MrCreosote:
Will also be used for creating and updating views, so will get quite a bit more complex and be used more often, so putting in a more obvious place. Will update documentation once the script is updated. This commit just moves the script to avoid mixing a file move and substantial file changes in the same commit.


Xiangs18 (approved):
LGTM

PR #473, opened 2023-09-29
RE2022-240: Improve service_manager usability

MrCreosote:
... in preparation for ArangoSearch view creation.

First cut was just getting it working without much thought for ease of use. Make it simpler to use since we'll be using it to create views as well, and the collections need to be created before the views.


Tianhao-Gu (approved):
LGTM

PR #488, opened 2023-10-03
Bulk dependency update

MrCreosote:
...and fix service startup killing bug I missed in the last PR. Only tested it with the CLI I'm working on


MrCreosote:
Not sure why the images are building

Tianhao-Gu (approved):
LGTM. 

tools are rebuilding as expected because 
https://github.com/kbase/collections/blob/main/.github/workflows/build-push-checkm2-image.yml#L13

PR #492, opened 2023-10-03
RE2022-240: Add CLI to create arangosearch views based on view specs

MrCreosote:


src/common/product_models/columnar_attribs_common_models.py
```diff
@@ -83,6 +84,9 @@ class ColumnarAttributesSpec(BaseModel):
     columns: Annotated[list[AttributesColumnSpec], Field(
         description="The set of columns in the table."
     )]
+    spec_files: Annotated[list[Path], Field(
+        description="Paths to the spec files from which the specs where loaded."
```
Tianhao-Gu, line 88: * spec_files are just the name of the spec files rather than path, right? e.g. `[genome_attribs-PMI.yml]`. 
* in which circumstances will the `ColumnarAttributesMeta` be linked to multiple spec files? I was thinking that each collection (PMI, GORW, etc.) would have its own column metadata, resulting with only one spec file per `ColumnarAttributesMeta`.

MrCreosote: * It's the full path
* When creating the view I have to merge all the spec files together to create the view spec

Xiangs18 (approved):
LGTM!

PR #494, opened 2023-10-05
RE2022-240: Fix filter aql bug

MrCreosote:
Variables declared in filters shouldn't be marked as bind variables


Tianhao-Gu (approved):
LGTM

PR #497, opened 2023-10-06
RE2022-240: Support count in arangosearch filters

MrCreosote:


src/service/data_products/genome_attributes.py
```diff
@@ -454,29 +453,38 @@ async def _get_internal_match_id(
...
-            doc = _remove_keys(doc)
-            if output_table:
-                data.append([doc[k] for k in sorted(doc)])
-            else:
-                data.append({k: doc[k] for k in sorted(doc)})
+        if filters.count:
+            count = await cur.next()
+            return {
+                _FLD_SKIP: 0,
+                _FLD_LIMIT: 0,
```
Tianhao-Gu, line 462: is `skip` and `limit` still a necessary return in this case?

MrCreosote: It's not super important, but it does signify that any skip and limit parameters that the user provided were ignored, and also is consistent with all the other return options

Tianhao-Gu: 👍

PR #498, opened 2023-10-09
RE2022-240: Support sorting and AND / OR for arangosearch filters

MrCreosote:


src/service/data_products/genome_attributes.py
```diff
@@ -364,6 +371,9 @@ async def get_genome_attributes(
     limit: common_models.QUERY_VALIDATOR_LIMIT = 1000,
     output_table: common_models.QUERY_VALIDATOR_OUTPUT_TABLE = True,
     count: common_models.QUERY_VALIDATOR_COUNT = False,
+    conjunction: Annotated[bool, Query(
```
Xiangs18, line 374: conjunction is not in common_models like others?

MrCreosote: No, this is the only place it exists for now

Xiangs18 (approved):
LGTM!

PR #500, opened 2023-10-11
RE2022-240: Simplify subset handling code

MrCreosote:


src/service/processing.py
```diff
@@ -256,3 +256,69 @@ async def heartbeat(millis: int):
...
+        subset_process - a process containing the ID of the subset. Ignored if internal_subset_id
+            is provided or the process is not complete.
+        mark_only - whether the subset should be filtered out (False) or just marked (True).
+        prefix - a prefix to apply to the subset ID.
+        """
+        self.internal_subset_id = None
+        if subset_process and subset_process.is_complete():
+            self.internal_subset_id = subset_process.internal_id
+        if internal_subset_id:
+            self.internal_subset_id = internal_subset_id
```
Tianhao-Gu, line 293:
```suggestion
        if internal_subset_id:
            self.internal_subset_id = internal_subset_id
        elif subset_process and subset_process.is_complete():
            self.internal_subset_id = subset_process.internal_id
```

src/service/processing.py
```diff
@@ -256,3 +256,69 @@ async def heartbeat(millis: int):
...
+            mark_only: bool = False,
+            prefix: str = None,
+        ):
+        """
+        Create the subset specification. If neither internal_subset_id nor a completed
+        subset_process are provided, the resultant instance is a null specification.
+        
+        internal_subset_id - the ID of the subset. If provided, subset_process is ignored.
+        subset_process - a process containing the ID of the subset. Ignored if internal_subset_id
+            is provided or the process is not complete.
```
Tianhao-Gu, line 285: when would both `internal_subset_id` and `subset_process` be provided, and should we ensure that only one of them is provided? Can you also add a comment explaining why is internal_subset_id given precedence over subset_process?

MrCreosote: > when would both internal_subset_id and subset_process be provided

Never

> should we ensure that only one of them is provided? 

I don't really have a strong opinion because there's no reason for this to ever happen. I'm fine with throwing an error if both are provided if that's what you want.

> why is internal_subset_id given precedence over subset_process?

It's more or less arbitrary. Since I never expect both to be provided at the same time I didn't care too much about the interactions

Tianhao-Gu: I would throw an error if both are provided.

src/service/processing.py
MrCreosote: fixed

src/service/processing.py
MrCreosote: fixed

Tianhao-Gu (approved):
👍

PR #501, opened 2023-10-12
RE2022-240: Support matches / selection in arangosearch filters

MrCreosote:


test/src/service/filtering/filters_test.py
```diff
@@ -332,6 +335,10 @@ def test_filterset_w_all_args():
         d.coll == @collid
         AND
         d.load_ver == @load_ver
+        AND
+        d._mtchsel == @internal_match_id
+        AND
+        d._mtchsel == @internal_selection_id
```
Xiangs18, line 341: I get confused here. What is the DS for doc[_mtchsel] here? It can equal to internal_match_id and internal_selection_id at the same time?

MrCreosote: Yes, `d._mtchsel` is an array of match and selection IDs. For regular arango `FILTER` expression it'd look like
```
FILTER @internal_match_id IN d._mtchsel
FILTER @internal_selection_id IN d._mtchsel
```

But in ArangoSearch those `IN`s translate to `==` due to the way ArangoSearch indexes the data.

Xiangs18: yeh, I found them in the [doc](https://docs.arangodb.com/3.11/index-and-search/arangosearch/) as well.

Xiangs18 (approved):
LGTM!

PR #504, opened 2023-10-16
Remove internal fields from genome attribs meta endpoint

MrCreosote:
`load_ver` shouldn't be exposed to users here and `coll` is redundant. Also remove the `inarray` filter strategy as arangosearch treats arrays and strings effectively identically when it comes to matching, so there's no need for it


Xiangs18 (approved):
LGTM!

PR #508, opened 2023-10-18
Add 2D plots design doc

MrCreosote:


dauglyon (approved):
Looks good to me. Will we also need to add filter support to our other data products? (mostly thinking about the taxonomy bar chart)

MrCreosote:
So far I have heard nothing along those lines. I'd guess we'd add filtering to samples before the taxa chart though

PR #511, opened 2023-10-23
Add provenance design notes

MrCreosote:


design/provenance.md
```diff
@@ -0,0 +1,58 @@
...
+(1) Note that collection versions are ephemeral and may be replaced at any time, but
+knowing the version can aid in debugging.
+```
+
+* Note that the provenance as a whole can be no more than 1MB. Since matches can be
+  no more  than 10k UPAs this is probably ok.
+* For the match parameters to be provided, the UI will need to send the match ID to the server
+  with the request.
+* Note that it's possible for a user to create a match, create a selection from a match, and
+  then alter the selection. Should the match information be included in that case? Should we
```
dauglyon, line 49: I like what you did above with the bool. We can mark it as something like "Selection derived from match" though I think this only makes sense if there are actual members of that match in the selection?

dauglyon (approved):

design/provenance.md
MrCreosote: yep, that seems reasonable

scanon: This seems good.  Does the matcher_id get garbage collected?  I guess it is still a hash of the parameters.  Correct?

scanon (approved):

MrCreosote:
> Does the matcher_id get garbage collected?
After 2 weeks without being accessed it'll get moved to a deleted state which will eventually be permanently deleted. If a new match with the same parameters is made during that time it creates a new match record.

> I guess it is still a hash of the parameters. Correct?
Right

design/provenance.md
MrCreosote: > Does the matcher_id get garbage collected?

After 2 weeks without being accessed it'll get moved to a deleted state which will eventually be permanently deleted. If a new match with the same parameters is made during that time it creates a new match record.

> I guess it is still a hash of the parameters. Correct?

Right

PR #512, opened 2023-10-25
explicitly provide ipv4 to callback server

Tianhao-Gu:
without explicitly provide `CALLBACK_IP`, the jobrunner container will return an ipv6 address. 

But the JobRunner code should return an IPV4 instead. I am not sure if it has something to do with how the image was built (there is no dockerfile in that repo). 
https://github.com/kbase/JobRunner/blob/master/JobRunner/JobRunner.py#L202-L204

```
(base) tgu@login39 tgu  $ podman run --name tian_callback -it --rm \
> --net host \
> -e KB_* \
> -e CALLBACK_PORT=9999 \
> -e KB_AUTH_TOKEN=$KB_AUTH_TOKEN \
> -e KB_BASE_URL=https://ci.kbase.us/services/ \
> -e JOB_DIR \
> -e DEBUG_MODE=true \
> -v $JOB_DIR:$JOB_DIR \
> kbase/callback:test
WARN[0000] Path "/etc/SUSEConnect" from "/etc/containers/mounts.conf" doesn't exist, skipping 
WARN[0000] Path "/etc/zypp/credentials.d/SCCcredentials" from "/etc/containers/mounts.conf" doesn't exist, skipping 
Warning: Missing admin token needed for volume mounts.
INFO:jr:Logger initialized for test
Callback IP provided (2620:0:28b0:64:0:ff:fe00:31f)
Job runner recieved Callback URL http://2620:0:28b0:64:0:ff:fe00:31f:9999/
```

### DIFF
--- a/src/loaders/workspace_downloader/workspace_downloader_helper.py
+++ b/src/loaders/workspace_downloader/workspace_downloader_helper.py
@@ -46,6 +46,7 @@ class Conf:
         token = loader_helper.get_token(token_filepath)
         self.retrieve_sample = retrieve_sample
         self.ignore_no_sample_error = ignore_no_sample_error
+        self.ipv4 = loader_helper.get_ip()
         self._start_callback_server(
             docker.from_env(),
             uuid.uuid4().hex,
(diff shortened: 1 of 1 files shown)


src/loaders/workspace_downloader/workspace_downloader_helper.py
```diff
@@ -46,18 +46,20 @@ def __init__(
         token = loader_helper.get_token(token_filepath)
         self.retrieve_sample = retrieve_sample
         self.ignore_no_sample_error = ignore_no_sample_error
+        self.ipv4 = loader_helper.get_ip()
```
MrCreosote, line 49: Doesn't seem like there's any need for this to be an instance variable

Tianhao-Gu: 👍

PR #513, opened 2023-10-27
RE2022-247: upgrade GTDB to 2.3.2

Tianhao-Gu:


MrCreosote:
The code looks fine, I just don't understand the logic of deleting the sketch db every time. Is it sketching the input or something?

Tianhao-Gu:
> The code looks fine, I just don't understand the logic of deleting the sketch db every time. Is it sketching the input or something?

I delete it just in case GTDB-TK was using the old mash reference from previous runs.
'--mash_db: path to save/read (if exists) the Mash reference sketch database (.msh)'

MrCreosote:
This is sketching the GTDB database, not the inputs?

Tianhao-Gu:
> This is sketching the GTDB database, not the inputs?

It‘s sketching the inputs.

MrCreosote:
Ok, can you add a comment to that effect to make it clear why the db is always deleted? Otherwise should be g2g

Tianhao-Gu:
> Ok, can you add a comment to that effect to make it clear why the db is always deleted? Otherwise should be g2g

Did already add comments. No clear enough?

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -33,6 +33,12 @@
 # If empty, select all available fields
 SELECTED_GTDBTK_SUMMARY_FEATURES = set()
 
+# GTDB-TK V2.2+ requirement
+# path of the sketched Mash database required for ANI screening
```
MrCreosote, line 37:
```suggestion
# path of the sketched Mash database from the input assembly files required for ANI screening
```

src/loaders/compute_tools/gtdb_tk/gtdb_tk.py
```diff
@@ -74,6 +80,12 @@ def _run_gtdb_tk(
     print(f'Start executing GTDB-TK for {size} genomes')
     start = time.time()
 
+    mash_db_path = output_dir / MASH_DB_PATH
+    if mash_db_path.exists():
+        # in the event of a rerun, remove an existing Mash database since we want GTDB-TK to create a new one
```
MrCreosote, line 85:
```suggestion
        # in the event of a rerun, remove an existing Mash database since we want GTDB-TK to create a new one
        # as the inputs might be a different set of assembly files
```

MrCreosote:
Made some suggestions that makes things clearer for me anyway

Tianhao-Gu:
> Made some suggestions that makes things clearer for me anyway, assuming I'm understanding things correclty

👍  thanks.

MrCreosote (approved):

PR #514, opened 2023-11-01
Refactor filters.py to generate AQL

MrCreosote:
as well as ArangoSearch. This'll allow us to put the table search aql generation code in one place instead of the current 3 and abstract away the difference between arangosearch and regular queries.

Next step is updating the code to use the filterset everywhere tables are queried.


Tianhao-Gu (approved):
GLTM.

PR #528, opened 2023-11-02
Bugfix for match and selection handling in standard AQL generation

MrCreosote:


Tianhao-Gu (dismissed):
👍

Tianhao-Gu (approved):
LGTM. Those bugs are really hard to be spotted in PR reviews.

MrCreosote:
Yeah, they're hard to catch until you actually use the code, which is what's happening now

PR #529, opened 2023-11-02
Refactor all table queries to use FilterSet

MrCreosote:
Cleans up very similar code in several places

### DIFF
--- a/src/service/data_products/samples.py
+++ b/src/service/data_products/samples.py
@@ -11,10 +11,10 @@ from src.common.product_models.common_models import SubsetProcessStates
 from src.service import app_state, errors, kb_auth, models
 from src.service.data_products.common_functions import (
     remove_collection_keys,
-    count_simple_collection_list,
     remove_marked_subset,
     query_table,
-    get_load_version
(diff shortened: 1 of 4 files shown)


MrCreosote:
image
Aw yiss

src/service/data_products/samples.py
```diff
@@ -203,71 +204,49 @@ async def get_samples(
...
+            skip=res.skip,
+            limit=res.limit,
+            count=res.count,
+            match_state=dp_match.state if dp_match else None,
+            selection_state=dp_sel.state if dp_sel else None,
+            fields=res.fields,
+            table=res.table,
+            data=res.data,
+        )
+    else:
```
Xiangs18, line 243: I would personally avoid `else` here.

MrCreosote: Here we have a difference of style - either way is technically correct, but to me having both returns in the if/else block

* makes it easier to read
* makes it easier to add more elif blocks in the future

So I prefer it as is

Xiangs18 (approved):
:100:

PR #530, opened 2023-11-05
RE2022-247: increase execution time for GTDB-TK

Tianhao-Gu:


MrCreosote:
I'm confused -

* If the time per genome is about the same for the 3 collections, why was only PMI timing out?
* Does 20m / genome mean that a batch of 100 will take 33 hours of compute? That seems insane. How long does it take with the pplacer etc. classification strategy?

Tianhao-Gu:
> I'm confused -
> 
> * If the time per genome is about the same for the 3 collections, why was only PMI timing out?
> * Does 20m / genome mean that a batch of 100 will take 33 hours of compute? That seems insane. How long does it take with the pplacer etc. classification strategy?

PMI times out consistently. It took about 7hrs to run PMI. Other collections did time out the first time I run the tool. And finished the second time. There must be some parallelizations when executing batch of genomes. But the experiment I run shows that genomes across all collections are slow for this latest GTDB-TK. If we were keeping this `ani_screen` step, the only option we have is decreasing batch sizes, increasing resources (CPU, time out threshold)  so that job can be finished within time out threshold.

MrCreosote:
* So why did the other 2 collections pass and PMI fail the 2nd round if PMI is on average the fastest to complete?
* How long does it take to complete a batch of 100 genomes?
* How long does it take if not using mash / ANI?

Tianhao-Gu:
> * So why did the other 2 collections pass and PMI fail the 2nd round if PMI is on average the fastest to complete?
I honestly have no idea. I am not able to answer you this question. Maybe we need to consult an SME someone really understands genomes and the tool. I think I might just get lucky when I ran GROW/ENIGMA because I scheduled those jobs pretty late and the node might be more efficient. 

> * How long does it take to complete a batch of 100 genomes?
If I batch PMI into 100 per batch, only 2 batches (one with 100 genomes, one with 50 genomes) finish within 5 hrs old timeout threshold with 32 CPU. All 4 other 100 genomes batch cannot be finished within 5 hours threshold. 

> * How long does it take if not using mash / ANI?
Without pre-screening, I was able to get all collections finished within 5 hours timeout threshold and utilize only 32 CPUs per task.

PR #533, opened 2023-11-06
RE2022-258: Add genome_attribs histogram endpoint

MrCreosote:


src/service/data_products/genome_attributes.py
```diff
@@ -258,19 +295,52 @@ async def _get_filters(
...
+    load_ver: str,
+    load_ver_override: bool,
+    view_name: str = None,
+    count: bool = False,
+    sort_on: str = None,
+    sort_desc: bool = False,
+    filter_conjunction: bool = True,
+    match_spec: SubsetSpecification = None,
+    selection_spec: SubsetSpecification = None,
+    keep: dict[str, set[col_models.ColumnType]] = None,
```
Xiangs18, line 314: I thought a column can only have one column type. Why we have a set here?

src/service/data_products/genome_attributes.py
```diff
@@ -451,12 +499,84 @@ async def get_genome_attributes(
     }
 
 
+class Histogram(BaseModel):
+    
+    bins: Annotated[list[float], Field(
```
Tianhao-Gu, line 504: bins should be `list[int]` and values should be `list[float]` right?

Xiangs18: bins is a list of float. Let's say [2.5, 3.5, 4.5], then we have [2.5, 3.5), [3.5, 4.5), and (4.5, upperbound]
values should be a list of int, indicating how many counts in that particular bin.

src/service/data_products/genome_attributes.py
```diff
@@ -451,12 +499,84 @@ async def get_genome_attributes(
...
+        example=[2.5, 3.5, 4.5],
+        description="The location of the histogram bins. Each bin starts at index i, "
+            + "inclusive, and ends at index i + 1, exclusive, except for the last bin which is "
+            + "inclusive at both sides."
+    )]
+    values: Annotated[list[int], Field(
+        example=[78.5, 96, 1],
```
Xiangs18, line 511: 78.5 is not an int

src/service/data_products/genome_attributes.py
Tianhao-Gu: 👍 I see. the `example ` appears conflicting.

src/service/data_products/genome_attributes.py
MrCreosote: Because the keep parameter specifies the allowable column types, for which there could be > 1

src/service/data_products/genome_attributes.py
MrCreosote: fixed

src/service/data_products/genome_attributes.py
MrCreosote: fixed

Tianhao-Gu (approved):
LGTM

src/service/data_products/genome_attributes.py
Xiangs18: ah, I assume the `key` of keep variable refers to a column name? A column's data type should be unique and consistent. Unless it refers to a data model or something else ...

MrCreosote: > ah, I assume the key of keep variable refers to a column name?

Correct

> A column's data type should be unique and consistent.

Which they are, e.g. https://github.com/kbase/collections/blob/main/src/common/collection_column_specs/genome_attribs-PMI.yml

Xiangs18: if they are unique and consistent, why each col could have a set of types?

MrCreosote: It's not the columns having a set of types, it's that the allowed column types for the histogram operation are a set. IOW, if the provided column is in the set {int, float} the operation can proceed, but if not, then an error is thrown. Take a look how the sets are used in the `_get_filters` method.

MrCreosote: E.g. https://github.com/kbase/collections/pull/533/files#diff-21772263d31206be1473129ce3db2950fcf3be1d52d3f702ea3d75497387f08aR338-R341

Xiangs18: https://github.com/kbase/collections/pull/533/files#diff-21772263d31206be1473129ce3db2950fcf3be1d52d3f702ea3d75497387f08aR561

Make more sense now. Indeed, It does not conflict with my point: https://github.com/kbase/collections/pull/533#discussion_r1385319584

PR #535, opened 2023-11-08
Monthly updates

MrCreosote:
Notes:
* Didn't update python to 3.12.0, it's only a month old
* Didn't update urrlib3, see Pipfile for notes


Tianhao-Gu (approved):
👍

PR #536, opened 2023-11-08
NCBI downloader includes Genomes

Xiangs18:


MrCreosote (approved):
Well that was easy

MrCreosote:
actually on second thought - are there tests for this that need to be updated? I think there might be

Xiangs18:
still need approval to merge

MrCreosote:
See comment re tests above

MrCreosote (commented):
We'll probably want to add more & better tests in the future but this seems GEFN.

@Tianhao-Gu can you take a look and approve if things look ok?

Tianhao-Gu (commented):
It will work. 

There are logic in place to prevent redownloading files in `remove_ids_with_existing_data`. But say if we run the download script for R207 with only `fna.gz` and then run the script for R214 with both `fna.gz` and `gbff.gz`. It will redownload and rewrite all existing `fna.gz` files. 

If that's an ok behavior, this PR looks good to me.

Tianhao-Gu:
> It will work.
> 
> There are logic in place to prevent redownloading files in `remove_ids_with_existing_data`. But say if we run the download script for R207 with only `fna.gz` and then run the script for R214 with both `fna.gz` and `gbff.gz`. It will redownload and rewrite all existing `fna.gz` files.
> 
> If that's an ok behavior, this PR looks good to me.

If we want to prevent redownloading, we need to limit `download_file_ext` to one in order for the avoid redownloading logic to work every time. Otherwise, we need to build logic and figure out target_file_ext for each genome and pass correct input to `download_genome_files` which is the right way to go in my option but involves a lot of refactoring work.

MrCreosote:
Good point. IMO we should make it work correctly and only download missing files rather than redownload the entire directory if any of the required files are missing

Xiangs18:
Not sure what is going on ... but the code only downloads missing files, not the entire directory 
https://github.com/kbase/collections/blob/2727b88ccab4de0ce2d1cd68b90878660bb31936/src/loaders/ncbi_downloader/ncbi_downloader_helper.py#L52

Tianhao-Gu:
> Not sure what is going on ... but the code only downloads missing files, not the entire directory
> 
> https://github.com/kbase/collections/blob/2727b88ccab4de0ce2d1cd68b90878660bb31936/src/loaders/ncbi_downloader/ncbi_downloader_helper.py#L52

I see. In that case, this PR LGTM.

Tianhao-Gu (approved):

Xiangs18:
@MrCreosote May I also get an approval from you before merging?

MrCreosote:
Tian's is sufficient

PR #537, opened 2023-11-09
RE2022-261: add fatal error file for checkm2

Tianhao-Gu:


src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -5,10 +5,13 @@
 from pathlib import Path
```
MrCreosote, line 5: I assume we don't have any tests for the tools yet right?

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -47,12 +50,23 @@ def _run_checkm2(
...
+    genome_attri_docs = process_genome_attri_result(output_dir,
+                                                    SELECTED_CHECKM2_FEATURES,
+                                                    genome_id_col,
+                                                    ids_to_files,
+                                                    [tool_file_name],
+                                                    )
+
+    # create fatal error file if any genome is missing from the result file
+    fatal_tuples = []
+    error_message = f"Result missing in the CheckM2 output file."
```
MrCreosote, line 62: is it not possible to be more specific than this?

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -47,12 +50,23 @@ def _run_checkm2(
...
+    success_ids = [genome_attri_doc[FLD_KBASE_ID] for genome_attri_doc in genome_attri_docs]
```
MrCreosote, line 63: have we actually had checkm fail yet?

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: no we don't. No tests for any tool computation codes.

src/loaders/compute_tools/checkm2/checkm2.py

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: Encountered only once (with a single missing genome out of 550), that's how I realized the absence of a fatal error file for CheckM. However, upon rerunning, the result file has been successfully recovered. So technically, no failure for checkm.

MrCreosote: How did you test this code worked as expected in that case?

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: So if checkm fails it just leaves a blank spot in the report? Are there any logs we could look at to get a better idea of the problem?

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Fair enough :+1:

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: yea. The result file just miss one of the genomes. And checkm finished fine with no errors. I don't recall I saw anything in the logs. But after rerunning we lost logs from the previous one. I found out the bug upon running the parser script.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: To test it, I manually remove a random id from `success_ids` and build a customized image in my personal registry.

MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Ok, I guess this is the best we can do for now then. If we start seeing a genome that consistently causes a fail we can look again and see if we can improve the error message

MrCreosote (approved):

PR #538, opened 2023-11-09
[RE2022-264] NCBI downloader fix

Xiangs18:


src/loaders/ncbi_downloader/gtdb.py
```diff
@@ -242,11 +242,12 @@ def main():
...
         print(f"\nSuccessfully downloaded {len(genome_ids)} genome files")
 
-    genome_ids_clean = set(genome_ids) - set(failed_ids)
+    # include a full set of genome_ids for each release version inside collectionssource dir
+    genome_ids_success = list(set(genome_ids_unprocessed) - set(failed_ids))
     loader_helper.create_softlinks_in_collection_source_dir(
         collection_source_dir,
         work_dir,
-        list(genome_ids_clean),
+        genome_ids_success,
```
MrCreosote, line 250: Are there any tests that need to be updated / written for this code?

MrCreosote (commented):
Code changes look good to me, but @Tianhao-Gu should definitely take a look. If he approves so do I

src/loaders/ncbi_downloader/gtdb.py
Xiangs18: This helper function comes from loader_helper file, which does not have any unit tests set up yet. I can write some for key functions if necessary.

MrCreosote: We're still not officially writing tests for this repo, so if there are no tests that need updating don't worry about it

Xiangs18: Then in that case, no tests need updating.

Tianhao-Gu (approved):
LGTM.

PR #539, opened 2023-11-10
RE2022-262: restruct import files

Tianhao-Gu:


MrCreosote (approved):
I'm assuming here you've caught all the spots that need updates due to the changes in the import creation methods. LGTM otherwise

Tianhao-Gu:
> I'm assuming here you've caught all the spots that need updates due to the changes in the import creation methods. LGTM otherwise

Yea. I think I get all places that use this method.

PR #540, opened 2023-11-10
RE2022-246: Add ngram based string matching

MrCreosote:
... and apply it to the classification field

### DIFF
--- a/src/common/collection_column_specs/genome_attribs-GTDB.yml
+++ b/src/common/collection_column_specs/genome_attribs-GTDB.yml
@@ -49,7 +49,7 @@ columns:
 
    - key: classification  # GTDB classification, might need better tokenizing
      type: string
-     filter_strategy: fulltext
+     filter_strategy: ngram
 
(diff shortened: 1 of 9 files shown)


src/common/collection_column_specs/genome_attribs-GTDB.yml
```diff
@@ -49,7 +49,7 @@ columns:
 
    - key: classification  # GTDB classification, might need better tokenizing
      type: string
-     filter_strategy: fulltext
```
Tianhao-Gu, line 52: Do we need to update genome_attribs-GROW/ENIGMA.yml as well?

MrCreosote: ENIGMA and GROW are just softlinks to PMI

Tianhao-Gu: I see.

Tianhao-Gu (approved):
👍

PR #541, opened 2023-11-10
RE2022-246: Set ngram threshold to 1

MrCreosote:
Make the match non-fuzzy

Also bump arango to 3.11 in the dockerfile


Tianhao-Gu (approved):
LGTM

Issue #294, opened 2023-05-04
Always return match and selection counts with matches and selections


MrCreosote:
... even if the lists aren't returned.

PR #543, opened 2023-11-12
Add counts to matches and selections

MrCreosote:
Allows getting the size of the match / selection without pulling back the lists of all the (possibly many) matches and selections

Closes https://github.com/kbase/collections/issues/294


Xiangs18 (approved):
LGTM!

PR #544, opened 2023-11-13
RE2022-268: Add basic provenance information when saving sets

MrCreosote:
... as per /design/provenance.md

Also throw an error if the description is > 800 chars

Match information is still do do

### DIFF
--- a/src/service/processing_selections.py
+++ b/src/service/processing_selections.py
@@ -21,6 +21,10 @@ from src.service.workspace_wrapper import WorkspaceWrapper, SetSpec
 
 MAX_SELECTION_IDS = 10000
 
+# The set descriptions go in the workspace metadata, which has a maximum size of 900 for the
+# key and value
+_MAX_DESCRIPTION_LENGTH=800
+
 _UTF_8 = "utf-8"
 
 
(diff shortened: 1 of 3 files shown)


src/service/processing_selections.py
```diff
@@ -244,7 +248,8 @@ async def save_set_to_workspace(
     workspace_id: int,
     object_name: str,
     ws_type: str,
-    description: str = None
+    service_ver: str = None,
+    description: str = None,
```
Xiangs18, line 252: Need to update doc string as well.

src/service/processing_selections.py
```diff
@@ -21,6 +21,10 @@
 
 MAX_SELECTION_IDS = 10000
 
+# The set descriptions go in the workspace metadata, which has a maximum size of 900 for the
```
Xiangs18, line 24: 800?

src/service/processing_selections.py
MrCreosote: I've got it done locally; it'll be in the next PR. Cool?

src/service/processing_selections.py
MrCreosote: Workspace limit is 900 for len(key) + len(value), so max description is 900 - len(key). I just rounded down to 800

Xiangs18 (approved):
LGTM!

PR #545, opened 2023-11-14
RE2022-268: Allow specifying a match when saving a selection as a workspace set

MrCreosote:
Match details are added to the set provenance

Still do to: add default parameters to the user submitted parameters when saving the match. Currently default parameters are excluded

### DIFF
--- a/src/service/processing_selections.py
+++ b/src/service/processing_selections.py
@@ -6,7 +6,7 @@ import hashlib
 import logging
 import uuid
 
-from typing import Callable, Awaitable
+from typing import Callable, Awaitable, Any
 
 from src.service.app_state_data_structures import CollectionsState, PickleableDependencies
 from src.service import data_product_specs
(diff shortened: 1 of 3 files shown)


src/service/processing_selections.py
```diff
@@ -260,7 +291,9 @@ async def save_set_to_workspace(
     workspace_id - the ID of the workspace where the data will be saved.
     object_name - the name of the object to save.
     ws_type - the type of data to save.
+    service_ver - the version of the service saving the set.
```
MrCreosote, line 294: as promised

MrCreosote:
There's some overlap but they're all separate things.

Matches are created at https://ci.kbase.us/services/collections/docs#/Collections/match_collections__collection_id__matchers__matcher_id__post

The endpoint allows you to select data in the workspace and match it against data in the collection, creating a match with the matched data as a list.

Selections are created at https://ci.kbase.us/services/collections/docs#/Collections/create_selection_collections__collection_id__selections_post

The endpoint allows you to create a selection by choosing data in the collection directly.


Filters are applied to a data table at https://ci.kbase.us/services/collections/docs#/Genome%20Attributes/get_genome_attributes_collections__collection_id__data_products_genome_attribs__get as well as the hist and scatter endpionts.

The endpoints allow you to apply filters to columns that will include / remove rows in the table from the results, based on the filter. 
You can also specify matches and selections in the endpoint with will filter or mark the results depending on the *_mark parameters

Xiangs18 (approved):
LGTM!
Appreciate the clarification!!

Issue #316, opened 2023-06-02
Populate user parameters with defaults


MrCreosote:
Currently if a user doesn't include user parameters they're included in the match as {}. They're also included in the match md5 ID as an empty dict, which means if the defaults ever change, the match will no longer be valid. The defaults should always be calculated and included in the match and match md5 ID.

MrCreosote:
https://github.com/mnboos/jsonschema-default

MrCreosote:
https://github.com/clarketm/mergedeep

PR #546, opened 2023-11-15
RE2022-268: Add default parameters to match record

MrCreosote:
If the user omitted to include a default match parameter, it would not be included in the match record, but the default value would be used to process the match. This is especially problematic when saving provenance a set from a selection that was created from a match.

Closes https://github.com/kbase/collections/issues/316


src/service/matchers/lineage_matcher.py
```diff
@@ -102,8 +103,7 @@ def generate_match_process(
         token - the user's token.
```
Xiangs18, line 103: token param is not used in this fun?

MrCreosote: Not for this matcher, but the minhash matcher (which implements the same method) does need it

src/service/processing_matches.py
```diff
@@ -81,6 +83,19 @@ async def create_match(
     return curr_match
 
 
+def _process_user_params(params: dict[str, Any], matcher: Matcher):
+    ret = jsonschema_default.create_from(matcher.user_parameters)
+    if params:
+        ret = mergedeep.merge(ret, params, strategy=mergedeep.Strategy.REPLACE)
```
Xiangs18, line 89: Just wanna clarify here: 
When `ret` and `params` keys are the same, we take the value from `params`?

Xiangs18: That does not sound right ...

MrCreosote: Right. 2nd argument keys replace 1st

Xiangs18: that does not sound right ...

MrCreosote: Why not?

Xiangs18 (approved):
LGTM!

PR #548, opened 2023-11-17
RE2022-263: assign default file group permission

Tianhao-Gu:


src/loaders/genome_collection/compute_genome_taxa_count.py
```diff
@@ -34,6 +34,9 @@
   --root_dir ROOT_DIR   Root directory for the collections project. (default: /global/cfs/cdirs/kbase/collections)
   --input_source {GTDB,genome_attributes}
                         Input file source
+  --file_group FILE_GROUP
+                        Assign file group (default: File group permission for created
+                        data files. Default is kbase)
```
MrCreosote, line 39: There should probably be a way to say "don't change the file group"

src/loaders/genome_collection/parse_PMI_biolog_data.py
```diff
@@ -171,6 +172,7 @@ def generate_pmi_biolog_heatmap_data(
     :param kbase_collection: the name of the KBase collection
     :param root_dir: the root directory for the data to be processed
     :param upa_map_file: the path to the JSON file containing the mapping between KBase PROD UPA to other environment UPA
+    :param file_group: the file group permission for created data files
```
MrCreosote, line 175: If it's None what's the behavior?

src/loaders/common/loader_common_names.py
```diff
@@ -31,6 +31,10 @@
 # Name for environment argument
```
MrCreosote, line 31: This covers the import files, but are there any other steps in the process that could produce unreadable files in perlmutter without a group change?

src/loaders/genome_collection/compute_genome_taxa_count.py
Tianhao-Gu: 👍

src/loaders/genome_collection/parse_PMI_biolog_data.py
Tianhao-Gu: if file_group is set to None, we will skip os.chown.

MrCreosote: Is that a legal group name?

src/loaders/genome_collection/compute_genome_taxa_count.py
MrCreosote: Is `None` a legal group name?

src/loaders/common/loader_common_names.py
Tianhao-Gu: I think import files might be it. Other files created in the process are not supposed to be manually handled.

src/loaders/genome_collection/parse_PMI_biolog_data.py
MrCreosote: Add to docstring?

src/loaders/common/loader_common_names.py
MrCreosote: What if we
* change tool code
* change the parser

... and someone wants to rerun with those changes against unreadable files?

Tianhao-Gu: I am not getting what you mean. It doesn't sound like a common case. There are a lot of files being created/downloaded. Unless we want to run chown on every files, there will be some unreadable files in some rare cases.

src/loaders/genome_collection/parse_PMI_biolog_data.py
Tianhao-Gu: 👍

src/loaders/genome_collection/compute_genome_taxa_count.py
Tianhao-Gu: yea. it's technically legal.

src/loaders/common/loader_common_names.py
MrCreosote: So we download the files from workspace A and they're only readable by user B. We find a bug in the gtdb tool container. User C wants to rerun the pipeline but can't because the files aren't readble.

Same thing if the user C just wants to rerun the parser on user B's output

src/loaders/genome_collection/parse_PMI_biolog_data.py
MrCreosote: :+1:

src/loaders/genome_collection/compute_genome_taxa_count.py
MrCreosote: Hm, is there a way to make the argument an illegal name?

Tianhao-Gu: I don't have permissions to try to create users group and test illegal user group names in NERSC. I think it's pretty safe to use None here. In order for this to break, someone (Shane) will actually create a users group called None and make it the default users group for our project.

src/loaders/common/loader_common_names.py
Tianhao-Gu: yea. Like I mentioned, in order to completely prevent this, we will need to refactor every script that handles files.

src/loaders/genome_collection/compute_genome_taxa_count.py
MrCreosote: $ sudo groupadd no:group
groupadd: 'no:group' is not a valid group name

src/loaders/common/loader_common_names.py
MrCreosote: Ok, so you don't want to do that right now, got it. Ticket instead? At some point this'll cause problems for us, but maybe not for a while

src/loaders/genome_collection/compute_genome_taxa_count.py
Tianhao-Gu: 👍 Set KEEP_FILE_GROUP = 'no:group'. Did you test it on NERSC? That user group is completely legal in mac.

MrCreosote: No, just my laptop. Let's see if @scanon has admin on NERSC

PR #549, opened 2023-11-18
RE2022-263, RE2022-271: set default file group

Tianhao-Gu:
after running those two comments on /global/cfs/cdirs/kbase/collections, tested rerun tools and parsers, all newly generated files are on kbase group now.

### DIFF
--- a/docs/data_pipeline_procedure.md
+++ b/docs/data_pipeline_procedure.md
@@ -4,11 +4,12 @@
    * Set working directory and its subdirectories to the default group
       ```commandline
       file_group=kbase
-      find /global/cfs/cdirs/kbase/collections/working_dir -type d -exec chgrp $file_group {} \;
+      working_dir=dir_to_set_group
+      find /global/cfs/cdirs/kbase/collections/$working_dir -type d -exec chgrp $file_group {} \;
       ```
    * set the "setgid" bit so that anything created or modified in that direcotry will automatically be set to the default group
       ```commandline
-      find /global/cfs/cdirs/kbase/collections/working_dir -type d -exec chmod g+s {} \;
+      find /global/cfs/cdirs/kbase/collections/$working_dir -type d -exec chmod g+s {} \;
       ```
 
 ## Download Source Data
--- a/docs/data_pipeline_procedure.md
+++ b/docs/data_pipeline_procedure.md
@@ -1,5 +1,16 @@
 # Data Pipeline Procedure
 
+## Set working directory to default group
+   * Set working directory and its subdirectories to the default group
+      ```commandline
+      file_group=kbase
+      find /global/cfs/cdirs/kbase/collections/working_dir -type d -exec chgrp $file_group {} \;
+      ```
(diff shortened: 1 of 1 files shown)


docs/data_pipeline_procedure.md
```diff
@@ -1,5 +1,17 @@
 # Data Pipeline Procedure
 
+## Set working directory to default group
+   * Set working directory and its subdirectories to the default group
+      ```commandline
+      file_group=kbase
+      working_dir=dir_to_set_group
+      find /global/cfs/cdirs/kbase/collections/$working_dir -type d -exec chgrp $file_group {} \;
```
MrCreosote, line 8: why the braces and backslash?

MrCreosote: This looks like it sets the group on the directories but not on any existing files. The next command sets the sticky group bit but I don't think that affects any current files either...? Or am I wrong about that

Tianhao-Gu: * it's taking the result of `find /global/cfs/cdirs/kbase/collections/$working_dir -type d` and execute each directories. 
* It only sets the directories and will only affect newly created files/sub-directories moving forward. If needed, we can manually sets the user's group for current files. But that will take quite some time since we do have a lot files. And we almostly only have CI files now.

MrCreosote: yeah, I get what the command does as a whole, I just don't understand the need for the `{} \` part

What I'm thinking about re current files is that the tools create directories when they run, so any files in those directories will be unreadable to anyone except the person that started the job (right?). Same for the downloaders.

MrCreosote: > chgrp and chmod needs to get the dir path returned from find.

ok,, I'm assuming that's what the braces are for but I still don't understand the need for the backslash - that's usually a line continuation marker but there's no line continuation here

re newly created files - ok now I get it. Wasn't thinking that directories would inherit the sticky bit

Tianhao-Gu: the `\`  and `;` are together. Otherwise, `-exec` won't terminate properly. At least in NERSC it won't. 

(base) tgu@login25 collections  $ find /global/cfs/cdirs/kbase/collections/$working_dir -type d -exec chgrp $file_group {}
find: missing argument to `-exec'
(base) tgu@login25 collections  $ find /global/cfs/cdirs/kbase/collections/$working_dir -type d -exec chgrp $file_group {} ;
find: missing argument to `-exec'
(base) tgu@login25 collections  $ find /global/cfs/cdirs/kbase/collections/$working_dir -type d -exec chgrp $file_group {} \;
chgrp: changing group of '/global/cfs/cdirs/kbase/collections/homology_service': Operation not permitted
chgrp: changing group of '/global/cfs/cdirs/kbase/collections/homology_service/CI': Operation not permitted
chgrp: changing group of '/global/cfs/cdirs/kbase/collections/homology_service/CI/archive': Operation not permitted

PR #569, opened 2023-12-04
RE2022-276: add display name to PMI biolog  heatmap

Tianhao-Gu:


MrCreosote (approved):
LGTM. This is the last one for display name updates right?

Tianhao-Gu:
> LGTM. This is the last one for display name updates right?

yes. After that, I am planning to reload updated data.

PR #577, opened 2023-12-08
Fix 500 error when plot data contains nulls pt2

MrCreosote:
Apply filters null filter in genome attribs plot endpoints

Manually tested David's failure mode with the hist and scatter endpoints with and without a filter included in the query


Tianhao-Gu (approved):
👍

PR #578, opened 2023-12-11
RE2022-275: reconstruct heatmap row cells

Tianhao-Gu:
transformed heatmap_data looks like:

{"kbase_id": "69037_10_1", "col_1_cell_id": "efe9dc0c-39c8-48d2-88ea-4fd93de5ec0c", "col_1_val": 0, "col_2_cell_id": "f5145e84-d370-4b1d-9906-edc58d4725ef", "col_2_val": 3, "col_3_cell_id": "a5fb45d9-e42e-44e6-821e-a418fb11e33d", "col_3_val": 0, "col_4_cell_id": "a2f1c72f-7515-4c77-83d9-4cd7b7e04256", "col_4_val": 0, "col_5_cell_id": "4677ebb4-328a-493d-9baa-370f60d9afa3", "col_5_val": 0, "col_6_cell_id": "b7b84f0a-55d3-4b65-a0ec-9ddb0fa138f6", "col_6_val": 0, "col_7_cell_id": "bd0f5d06-1322-44f9-b222-6524075ab095", "col_7_val": 2, ... ...

### DIFF
--- a/src/loaders/common/loader_helper.py
+++ b/src/loaders/common/loader_helper.py
@@ -496,7 +496,7 @@ class ExplicitDefaultsHelpFormatter(argparse.ArgumentDefaultsHelpFormatter):
         return super()._get_help_string(action)
 
 
-def transform_heatmap_row_cells(data: dict):
+def transform_heatmap_row_cells(data: dict[str, Any]):
     """
     Transform the cells structure in a heatmap row to a new structure.
 
--- a/src/loaders/common/loader_helper.py
+++ b/src/loaders/common/loader_helper.py
@@ -22,6 +22,12 @@ from src.common.product_models.columnar_attribs_common_models import (
     AttributesColumn,
     ColumnarAttributesMeta,
 )
+from src.common.product_models.heatmap_common_models import (
+    FIELD_HEATMAP_ROW_CELLS,
+    FIELD_HEATMAP_COL_ID,
+    FIELD_HEATMAP_CELL_ID,
+    FIELD_HEATMAP_CELL_VALUE,
+)
 from src.common.storage.db_doc_conversions import (
     collection_data_id_key,
     collection_load_version_key,
@@ -45,6 +51,9 @@ This module contains helper functions used for loaders (e.g. compute_genome_attr
 """
 
 NONE_STR = ['N/A', 'NA', 'None', 'none', 'null', 'Null', 'NULL', '']
+
+HEATMAP_COL_PREFIX = "col"
+HEATMAP_COL_SEPARATOR = '_'
 
 
 def _convert_to_iso8601(date_string: str) -> str:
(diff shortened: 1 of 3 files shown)


src/loaders/common/loader_helper.py
```diff
@@ -46,6 +52,9 @@
 
 NONE_STR = ['N/A', 'NA', 'None', 'none', 'null', 'Null', 'NULL', '']
 
+HEATMAP_COL_PREFIX = "col"
+HEATMAP_COL_SEPARATOR = '_'
```
MrCreosote, line 56: Will these be needed by the service? If so they should go in common names

src/loaders/common/loader_helper.py
```diff
@@ -485,3 +494,49 @@ def _get_help_string(self, action):
         if action.default is None or action.default is False:
             return action.help
         return super()._get_help_string(action)
+
+
+def transform_heatmap_row_cells(data: dict):
+    """
+    Transform the cells structure in a heatmap row to a new structure.
```
MrCreosote, line 501:
```suggestion
    Transform, in place, the cells structure in a heatmap row to a new structure.
```

src/loaders/common/loader_helper.py
```diff
@@ -485,3 +494,49 @@ def _get_help_string(self, action):
         if action.default is None or action.default is False:
             return action.help
         return super()._get_help_string(action)
+
+
+def transform_heatmap_row_cells(data: dict):
```
MrCreosote, line 499:
```suggestion
def transform_heatmap_row_cells(data: dict[str, Any]):
```

src/loaders/common/loader_helper.py
```diff
@@ -485,3 +494,49 @@ def _get_help_string(self, action):
...
+    The new structure:
+    "col_0_cell_id": "cell_0",
+    "col_0_val": 0.0,
+    "col_1_cell_id": "cell_1",
+    "col_1_val": 1.0
+
+    """
+
+    # Iterate over the 'cells' structure and remove them from the data structure while constructing the new structure
+    for cell in data.pop(FIELD_HEATMAP_ROW_CELLS):
```
MrCreosote, line 531: It does seem a little weird to do the transform here, but I suppose we can push it back to the tools if it becomes a problem.

MrCreosote (commented):
Generally looks good, a couple minor things &  a question

src/loaders/common/loader_helper.py
Tianhao-Gu: I thought about this too. I intended to relocate it to common modules if it turns out we'll utilize it in the service. Otherwise, it will only be used for this module.

MrCreosote: :+1:

src/loaders/common/loader_helper.py
Tianhao-Gu: yea. Other than that, I think it will be easier for us to revert/skip this step if Arango nested search did become available in Community Edition.

MrCreosote (approved):

PR #582, opened 2023-12-14
RE2022-275: create search view for biolog/microtrait

Tianhao-Gu:
local created view for microtrait:
`
{
  "writebufferSizeMax": 33554432,
  "id": "27714",
  "storedValues": [],
  "name": "m1",
  "type": "arangosearch",
  "consolidationPolicy": {
    "type": "tier",
    "segmentsBytesFloor": 2097152,
    "segmentsBytesMax": 5368709120,
    "segmentsMax": 10,
    "segmentsMin": 1,
    "minScore": 0
  },
  "writebufferActive": 0,
  "links": {
    "kbcoll_microtrait_data": {
      "analyzers": [
        "identity"
      ],
      "fields": {
        "kbase_display_name": {
          "analyzers": [
            "kbcoll_en_ngram3"
          ]
        }
      },
      "includeAllFields": true,
      "storeValues": "none",
      "trackListPositions": false
    }
  },
  "commitIntervalMsec": 1000,
  "consolidationIntervalMsec": 1000,
  "globallyUniqueId": "hEED053C208E7/27714",
  "cleanupIntervalStep": 2,
  "primarySort": [],
  "primarySortCompression": "lz4",
  "writebufferIdle": 64
}
`

### DIFF
--- a/src/service/data_products/common_models.py
+++ b/src/service/data_products/common_models.py
@@ -84,15 +84,15 @@ class DataProductSpec(BaseModel):
     def _ensure_only_one_view(cls, v):
         found = False
         for dbc in v:
-            if found and (v.view_required or v.generic_view_required):
+            if found and v.view_required:
                 raise ValueError("More than one db collection requiring a view found")
-            found = found or dbc.view_required
+            found = found or dbc.view_required or dbc.generic_view_required
         return v
     
     def view_required(self):
         """ Check if a search view is required for this data product. """
         for db in self.db_collections:
-            if db.view_required:
+            if db.view_required or db.generic_view_required:
                 return True
         return False
 
--- a/src/service/data_products/common_models.py
+++ b/src/service/data_products/common_models.py
@@ -82,17 +82,15 @@ class DataProductSpec(BaseModel):
     @field_validator("db_collections")
     @classmethod
     def _ensure_only_one_view(cls, v):
-        found = False
-        for dbc in v:
-            if found and v.view_required:
-                raise ValueError("More than one db collection requiring a view found")
-            found = found or dbc.view_required or dbc.generic_view_required
+        view_required_collections = [dbc for dbc in v if dbc.view_required or dbc.generic_view_required]
+        if len(view_required_collections) > 1:
+            raise ValueError("More than one db collection requiring a view found")
         return v
     
     def view_required(self):
         """ Check if a search view is required for this data product. """
         for db in self.db_collections:
-            if db.view_required or db.generic_view_required:
+            if db.view_required:
                 return True
         return False
 
--- a/src/service/data_products/common_models.py
+++ b/src/service/data_products/common_models.py
@@ -88,7 +88,7 @@ class DataProductSpec(BaseModel):
         return v
     
     def view_required(self):
-        """ Check if a search view is required for this data product. """
+        """ Check if a non-generic search view is required for this data product. """
         for db in self.db_collections:
             if db.view_required:
                 return True
--- a/src/service/data_products/common_models.py
+++ b/src/service/data_products/common_models.py
@@ -27,6 +27,12 @@ class DBCollection(BaseModel):
     view associated with it.
     """
 
+    generic_view_required: bool = False
+    """
+    Whether the collection requires a generic ArangoSearch view. If so, the view will be automatically 
+    created by the service manager. `view_required` flag is ignored if this flag is set.
+    """
+
     indexes: list[list[str]]
     """
     The indexes in the collection. Each item in the outer list is an index, with the inner
--- a/src/service/data_products/common_models.py
+++ b/src/service/data_products/common_models.py
@@ -84,7 +84,7 @@ class DataProductSpec(BaseModel):
     def _ensure_only_one_view(cls, v):
         found = False
         for dbc in v:
-            if found and v.view_required:
+            if found and (v.view_required or v.generic_view_required):
                 raise ValueError("More than one db collection requiring a view found")
             found = found or dbc.view_required
         return v
--- a/src/service/data_products/common_models.py
+++ b/src/service/data_products/common_models.py
@@ -29,8 +29,7 @@ class DBCollection(BaseModel):
 
     generic_view_required: bool = False
     """
-    Whether the collection requires a generic ArangoSearch view. If so, the view will be automatically 
-    created by the service manager. `view_required` flag is ignored if this flag is set.
+    Whether the collection requires a generic ArangoSearch view. `view_required` flag is ignored if this flag is set.
     """
 
     indexes: list[list[str]]
--- a/src/service/filtering/generic_view.py
+++ b/src/service/filtering/generic_view.py
@@ -9,8 +9,6 @@ _FIELD_COL = {
 
 _GENERIC_VIEW_NAME_SUFFIX = "_generic_view"
 
-_GENERIC_VIEW_PRODUCTS = ['microtrait', 'biolog']
-
 
 def create_generic_spec() -> ColumnarAttributesSpec:
     """
@@ -23,13 +21,6 @@ def create_generic_spec() -> ColumnarAttributesSpec:
     columns = [AttributesColumnSpec(key=key, **specs) for key, specs in _FIELD_COL.items()]
 
     return ColumnarAttributesSpec(columns=columns)
-
-
-def is_generic_view_product(data_product: str) -> bool:
-    """
-    Check if a data product requires a generic view.
-    """
-    return data_product in _GENERIC_VIEW_PRODUCTS
 
 
 def get_generic_view_name(data_product: str) -> str:
--- a/src/service/filtering/generic_view.py
+++ b/src/service/filtering/generic_view.py
@@ -12,8 +12,8 @@ def create_generic_spec() -> ColumnarAttributesSpec:
     """
     Create a generic spec for a collection.
 
-    Arango search includeAllFields will be set to true. This means that all fields will be indexed
-    by default except for those that are explicitly defined in the _FIELD_COL dict.
+    This spec only includes fields common to most or all collections that require special handling and it is
+    expected that all other fields are handled in a generic way.
     """
 
     columns = [AttributesColumnSpec(key=key, **specs) for key, specs in _FIELD_COL.items()]
--- a/src/service/filtering/generic_view.py
+++ b/src/service/filtering/generic_view.py
@@ -10,7 +10,7 @@ _FIELD_COL = {
 
 def create_generic_spec() -> ColumnarAttributesSpec:
     """
-    Create a generic spec for a collection.
+    Create a generic column spec for a collection.
 
     This spec only includes fields common to most or all collections that require special handling and it is
     expected that all other fields are handled in a generic way.
--- a/src/service/filtering/generic_view.py
+++ b/src/service/filtering/generic_view.py
@@ -1,0 +1,21 @@
+from src.common.product_models.columnar_attribs_common_models import ColumnarAttributesSpec, AttributesColumnSpec
+
+_FIELD_COL = {
+    "kbase_display_name": {
+        "type": "string",
+        "filter_strategy": "ngram",
+    }
+}
+
+
+def create_generic_spec() -> ColumnarAttributesSpec:
+    """
+    Create a generic spec for a collection.
+
+    Arango search includeAllFields will be set to true. This means that all fields will be indexed
+    by default except for those that are explicitly defined in the _FIELD_COL dict.
+    """
+
+    columns = [AttributesColumnSpec(key=key, **specs) for key, specs in _FIELD_COL.items()]
+
+    return ColumnarAttributesSpec(columns=columns)
--- a/src/service/filtering/generic_view.py
+++ b/src/service/filtering/generic_view.py
@@ -7,6 +7,10 @@ _FIELD_COL = {
     }
 }
 
+_GENERIC_VIEW_NAME_SUFFIX = "_generic_view"
+
+_GENERIC_VIEW_PRODUCTS = ['microtrait', 'biolog']
+
 
 def create_generic_spec() -> ColumnarAttributesSpec:
     """
@@ -19,3 +23,17 @@ def create_generic_spec() -> ColumnarAttributesSpec:
     columns = [AttributesColumnSpec(key=key, **specs) for key, specs in _FIELD_COL.items()]
 
     return ColumnarAttributesSpec(columns=columns)
+
+
+def is_generic_view_product(data_product: str) -> bool:
+    """
+    Check if a data product requires a generic view.
+    """
+    return data_product in _GENERIC_VIEW_PRODUCTS
+
+
+def get_generic_view_name(data_product: str) -> str:
+    """
+    Get the name of the generic view for a data product.
+    """
+    return f"{data_product}{_GENERIC_VIEW_NAME_SUFFIX}"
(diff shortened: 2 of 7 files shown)


src/service/data_products/common_models.py
```diff
@@ -27,6 +27,12 @@ class DBCollection(BaseModel):
     view associated with it.
     """
 
+    generic_view_required: bool = False
+    """
+    Whether the collection requires a generic ArangoSearch view. If so, the view will be automatically 
+    created by the service manager. `view_required` flag is ignored if this flag is set.
```
MrCreosote, line 33:
```suggestion
    Whether the collection requires a generic ArangoSearch view. `view_required` flag is ignored if this flag is set.
```
The service code really shouldn't know the service manager exists

src/service/filtering/generic_view.py
```diff
@@ -0,0 +1,21 @@
...
+    }
+}
+
+
+def create_generic_spec() -> ColumnarAttributesSpec:
+    """
+    Create a generic spec for a collection.
+
+    Arango search includeAllFields will be set to true. This means that all fields will be indexed
+    by default except for those that are explicitly defined in the _FIELD_COL dict.
```
MrCreosote, line 16:
```suggestion
    This spec only includes fields common to most or all collections that require special handling and it is
    expected that all other fields are handled in a generic way.
```

Edited to avoid mentioning internals in external docs and remove references to Arango, which this code shouldn't know about.

src/service/filtering/generic_view.py
```diff
@@ -0,0 +1,21 @@
...
+    "kbase_display_name": {
+        "type": "string",
+        "filter_strategy": "ngram",
+    }
+}
+
+
+def create_generic_spec() -> ColumnarAttributesSpec:
+    """
+    Create a generic spec for a collection.
```
MrCreosote, line 13:
```suggestion
    Create a generic column spec for a collection.
```

src/service_manager.py
```diff
@@ -157,13 +157,24 @@ async def _create_view(
...
+
     views = sorted(await store.get_search_views_from_spec(
-        db_collection_name, view_spec, analyzers.get_analyzer))
+        db_collection_name,
+        view_spec,
+        analyzers.get_analyzer,
+        include_all_fields=create_generic_view))
     _print_view_status(data_product, db_collection_name, views)
     if views:
         print("    Setup ok.")
```
MrCreosote, line 180: Below the generic view is assigned a view name by the user. How is the service going to know which view to use?

src/service/storage_arango.py
```diff
@@ -254,7 +254,8 @@ async def create_search_view(
         name: str,
         arango_collection: str,
         view_spec: ColumnarAttributesSpec,
-        analyzer_provider: Callable[[FilterStrategy, bool], str]
+        analyzer_provider: Callable[[FilterStrategy, bool], str],
+        include_all_fields: bool = False
```
MrCreosote, line 258: needs docs

src/service/storage_arango.py
```diff
@@ -300,7 +303,8 @@ async def get_search_views_from_spec(
         self,
         arango_collection: str,
         view_spec: ColumnarAttributesSpec,
-        analyzer_provider: Callable[[FilterStrategy, bool], str]
+        analyzer_provider: Callable[[FilterStrategy, bool], str],
+        include_all_fields: bool = False
```
MrCreosote, line 307: needs docs

src/service_manager.py
```diff
@@ -173,7 +184,11 @@ async def _create_view(
...
             await store.create_search_view(
-                view_name, db_collection_name, view_spec, analyzers.get_analyzer)
+                view_name,
+                db_collection_name,
+                view_spec,
+                analyzers.get_analyzer,
+                include_all_fields=create_generic_view)
             print(f"done in {time.time() - t0:.2f} seconds.")
             return view_name
         except ViewExistsError:
```
MrCreosote, line 194: The prints below this point do apply for generic specs

Tianhao-Gu: 👍

src/service/storage_arango.py
Tianhao-Gu: 👍

src/service_manager.py
Tianhao-Gu: it should be specified at `save collection` step. Same as genome attributes.

src/service/storage_arango.py
Tianhao-Gu: 👍

src/service_manager.py
MrCreosote: But why? it doesn't depend on the a manual spec change so we can do this automatically. There's no reason to put the user in the loop at all

src/service_manager.py
MrCreosote: :

src/service/storage_arango.py
MrCreosote: :+1:

src/service/storage_arango.py
MrCreosote: :+1:

src/service_manager.py
Tianhao-Gu: hmm once the view is created. I don't see why we are doing differently between genome attributes and the heatmap. Maybe we should just always auto link newly created view to the corresponding collection?

MrCreosote: I'm saying there's no reason to link a generic view to a particular collection. It's generic, it applies to all versions of all collections

Tianhao-Gu: 👍 creating a common name for generic view. I also updated `get_data_product` in models.py. If you think we should instead get product exactly as what's stored in DB. I can revert that and get the common search view name when I work on updating FilterSet in heatmap.py.

src/service/filtering/generic_view.py
```diff
@@ -0,0 +1,39 @@
...
+    This spec only includes fields common to most or all collections that require special handling and it is
+    expected that all other fields are handled in a generic way.
+    """
+
+    columns = [AttributesColumnSpec(key=key, **specs) for key, specs in _FIELD_COL.items()]
+
+    return ColumnarAttributesSpec(columns=columns)
+
+
+def is_generic_view_product(data_product: str) -> bool:
```
MrCreosote, line 28: This can be checked by looking for the `requires_generic_view` flag on the data product spec - maybe add a method there to make it easy? Or a method to get the view name from the data product spec class

MrCreosote: E.g. https://github.com/kbase/collections/pull/582/files#diff-ab437ceb2ccb0562a34ddf1b5ecccc3db5b4382fc820c99c8c32b8d7439765f9R92

src/service/models.py
```diff
@@ -294,6 +295,8 @@ def get_data_product(self, data_product) -> DataProduct:
         """ Get a data product by its ID. """
         for dp in self.data_products:
             if dp.product == data_product:
+                if generic_view.is_generic_view_product(data_product):
+                    dp.search_view = generic_view.get_generic_view_name(data_product)
```
MrCreosote, line 299: There's no reason to set a generic search view on a collection, since they're always the same for all collections.

src/service/storage_arango.py
```diff
@@ -311,6 +316,7 @@ async def get_search_views_from_spec(
             an analyzer to use for that strategy. The second argument defines whether to
             return None (True) or the name of the default analyzer (False) when the default
             analyzer is to be returned.
+        include_all_fields - check if returned matching views have the include_all_fields flag set.
```
MrCreosote, line 319:
```suggestion
        include_all_fields - returned matching views must have the include_all_fields flag set.
```

src/service/filtering/generic_view.py
MrCreosote: Or instead of a flag you could maybe just hard code the view name. Haven't really thought that through though

src/service/models.py
Tianhao-Gu: 👍

src/service/filtering/generic_view.py
Tianhao-Gu: removed this method since there is no usage for this method now.

src/service/models.py
MrCreosote: :+1:

src/service/filtering/generic_view.py
MrCreosote: :+1:

src/service_manager.py
MrCreosote: :+1:

src/service/data_products/common_models.py
```diff
@@ -27,6 +27,11 @@ class DBCollection(BaseModel):
     view associated with it.
     """
 
+    generic_view_required: bool = False
```
MrCreosote, line 30: The _ensure_only_one_view method below needs updating for this field

src/service/models.py
```diff
@@ -15,6 +15,7 @@
 from typing import Any, Self, Annotated
 
 from src.service.arg_checkers import contains_control_characters
+from src.service.filtering import generic_view
```
MrCreosote, line 18: unused

src/service/data_products/common_models.py
```diff
@@ -81,13 +86,13 @@ def _ensure_only_one_view(cls, v):
         for dbc in v:
             if found and v.view_required:
                 raise ValueError("More than one db collection requiring a view found")
-            found = found or dbc.view_required
+            found = found or dbc.view_required or dbc.generic_view_required
         return v
```
MrCreosote, line 90: What if there are 2 generic views?

src/service/data_products/common_models.py
```diff
@@ -81,13 +86,13 @@ def _ensure_only_one_view(cls, v):
...
     
     def view_required(self):
         """ Check if a search view is required for this data product. """
         for db in self.db_collections:
-            if db.view_required:
+            if db.view_required or db.generic_view_required:
```
MrCreosote, line 95: This will break updating collections: https://github.com/kbase/collections/blob/24d29cf66451b980472f1e8c6ea8d3dba537d82b/src/service/routes_collections.py#L59

src/service/models.py
MrCreosote: :+1:

MrCreosote:
OMG github sucks at handling comments during pending reviews

src/service/data_products/common_models.py
```diff
@@ -77,11 +82,9 @@ def _check_router_tags(cls, v):  # @NoSelf
...
-        for dbc in v:
-            if found and v.view_required:
-                raise ValueError("More than one db collection requiring a view found")
-            found = found or dbc.view_required
+        view_required_collections = [dbc for dbc in v if dbc.view_required or dbc.generic_view_required]
+        if len(view_required_collections) > 1:
+            raise ValueError("More than one db collection requiring a view found")
         return v
     
     def view_required(self):
```
MrCreosote, line 90: Can you update the method docs to say
```
Check if a non-generic search view is required for this data product. 
```
Unfortunately you can't make a suggestion on a hidden line

src/service/data_products/common_models.py
```diff
@@ -77,11 +82,9 @@ def _check_router_tags(cls, v):  # @NoSelf
...
     @classmethod
     def _ensure_only_one_view(cls, v):
-        found = False
-        for dbc in v:
-            if found and v.view_required:
-                raise ValueError("More than one db collection requiring a view found")
-            found = found or dbc.view_required
+        view_required_collections = [dbc for dbc in v if dbc.view_required or dbc.generic_view_required]
+        if len(view_required_collections) > 1:
+            raise ValueError("More than one db collection requiring a view found")
```
MrCreosote, line 87: So does this actually work? now I distrust how the method is set up since the prior version had a bug in it that seems like it should have puked if the method was run

src/service/data_products/common_models.py
MrCreosote: :+1:

src/service/data_products/common_models.py
MrCreosote: :+1:

src/service/data_products/common_models.py
MrCreosote: :+1:

src/service/data_products/common_models.py
Tianhao-Gu: 👍

src/service/data_products/common_models.py
Tianhao-Gu: I think it's working as expected. 

`
In [10]: DataProductSpec(
    ...:         data_product="test_product",
    ...:         db_collections=[
    ...:             DBCollection(name="Collection1", view_required=True, generic_view_require
    ...: d=False, indexes=[]),
    ...:             DBCollection(name="Collection2", view_required=False, generic_view_requir
    ...: ed=True, indexes=[]),
    ...:         ],
    ...:         router=APIRouter(tags=["tag1"]),
    ...:     )
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[10], line 1
----> 1 DataProductSpec(
      2         data_product="test_product", 
      3         db_collections=[
      4             DBCollection(name="Collection1", view_required=True, generic_view_required=False, indexes=[]),
      5             DBCollection(name="Collection2", view_required=False, generic_view_required=True, indexes=[]),
      6         ],
      7         router=APIRouter(tags=["tag1"]),
      8     )

File /opt/miniconda3/envs/dev/lib/python3.11/site-packages/pydantic/main.py:164, in BaseModel.__init__(__pydantic_self__, **data)
    162 # `__tracebackhide__` tells pytest and some other tools to omit this function from tracebacks
    163 __tracebackhide__ = True
--> 164 __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)

ValidationError: 1 validation error for DataProductSpec
db_collections
  Value error, More than one db collection requiring a view found [type=value_error, input_value=[DBCollection(name='Colle...uired=True, indexes=[])], input_type=list]
    For further information visit https://errors.pydantic.dev/2.5/v/value_error
`

Tianhao-Gu: If I import the code exactly as it is in the main branch, it will raise an expected error: 'list' object has no attribute 'view_required'. Surprisingly, this issue was never not caught.

Screenshot 2023-12-15 at 4 13 41 PM

src/service/data_products/common_models.py
MrCreosote: :+1:

src/service/filtering/generic_view.py
```diff
@@ -0,0 +1,30 @@
...
+    This spec only includes fields common to most or all collections that require special handling and it is
+    expected that all other fields are handled in a generic way.
+    """
+
+    columns = [AttributesColumnSpec(key=key, **specs) for key, specs in _FIELD_COL.items()]
+
+    return ColumnarAttributesSpec(columns=columns)
+
+
+def get_generic_view_name(data_product: str) -> str:
```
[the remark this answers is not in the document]
Tianhao-Gu: I guess we can leave it here until we need to move it to data product spec

src/service/data_products/common_models.py
MrCreosote: yeah, I don't understand how `genome_attribs` was ever able to import like that. It should've choked

MrCreosote: Oh, I bet I know. The `and` causes the `if` to short circuit and since the `dbc` with `view_required` is last there's never a loop with `found==True`

MrCreosote: That means I either never tested it or screwed up the test though...

MrCreosote (approved):

src/service/data_products/common_models.py
Tianhao-Gu: yea. That's exactly what happened. `found` was never True. If we put 'genome_attribs' before 'genome_attribs_meta', this bug will appear.

PR #583, opened 2023-12-15
RE2022-275: add boolean filter

Tianhao-Gu:


src/service/filtering/filters.py
```diff
@@ -92,6 +95,68 @@ def _gt(num: int, min_: int, name: str):
     return num
 
 
+class BooleanFilter(AbstractFilter):
```
MrCreosote, line 98: it occurs to me that for heatmaps we only need to use filters if there's a search on the display name. Otherwise we could probably just use regular AQL. I don't really see a strong reason to avoid arangosearch though, other than it's eventually consistent

MrCreosote, line 98: This filter needs to be added to FilterSet._FILTER_MAP

src/service/filtering/filters.py
```diff
@@ -19,6 +19,9 @@
```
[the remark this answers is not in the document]
Tianhao-Gu: 👍 added tests.

src/service/filtering/filters.py
Tianhao-Gu: I'd like to keep things consistent. It's more easier to maintain.

Tianhao-Gu: I know. Coming in future PRs. I think I need to merge the `ColumnType` from `heatmap_common_models` to `columnar_attribs_common_models`. So that we only have one `ColumnType` in the system. But `microtrait` runner is using `heatmap_common_models.ColumnType`. I need to do more testing to make sure `microtrait` can import `columnar_attribs_common_models`

MrCreosote: Hmm... I don't like the idea of mixing the columns types from heatmaps and attribs. A translator from heatmap -> attribs might be better, only since the filters understand attribs.

Or - have the filters understand both heat map and attribs column types

src/service/filtering/filters.py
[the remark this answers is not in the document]
MrCreosote: Looks like the equals method isn't tested against a different filter type / different object tye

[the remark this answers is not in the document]
MrCreosote: Also take a look at the round trip repr tests for the other filters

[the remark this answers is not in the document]
MrCreosote: And the constructor tests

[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/service/filtering/filters.py
Tianhao-Gu: 👍

src/common/product_models/heatmap_common_models.py
```diff
@@ -52,6 +53,26 @@ class ColumnType(str, Enum):
...
 
 
+def trans_column_type_heatmap_to_attribs(col_type: ColumnType) -> AttribsColumnType:
+    """
+    Translate a heatmap column type to an attributes column type.
+    This method is suitable for models like filters that exclusively operate on attributes column types
+
+    col_type: the heatmap column type
+    """
+    heatmap_to_attribs_mapping = {
```
MrCreosote, line 63: I would make this a module level constant

src/service/filtering/filters.py
```diff
@@ -92,6 +95,68 @@ def _gt(num: int, min_: int, name: str):
...
+                f"Invalid boolean specification; expected true or false: {string}")
+        return BooleanFilter(string == _TRUE_STR)
+
+    def to_arangosearch_aql(self, identifier: str, var_prefix: str) -> SearchQueryPart:
+        """
+        Convert the filter to lines of ArangoSearch AQL and bind variables.
+
+        identifier - the identifier for where the search is to take place, for example
+            `doc.classification`. This will be inserted verbatim into the search constraint, e.g.
+            `f"{identifer} > 47`
```
MrCreosote, line 149:
```suggestion
            `f"{identifer} == true`
```

test/src/service/filtering/filters_test.py
```diff
@@ -132,6 +134,51 @@ def test_rangefilter_from_string_fail():
...
+    return BooleanFilter(True)
+
+
+@pytest.fixture
+def boolean_filter_false():
+    return BooleanFilter(False)
+
+
+def test_boolean_filter_equality(boolean_filter_true, boolean_filter_false):
+    assert boolean_filter_true == boolean_filter_true
```
MrCreosote, line 148:
```suggestion
    assert boolean_filter_true == boolean_filter_true
    # test that non-identical objects also pass
    assert boolean_filter_true == BooleanFilter(true)
    
```

test/src/service/filtering/filters_test.py
```diff
@@ -132,6 +134,51 @@ def test_rangefilter_from_string_fail():
...
+    assert repr(boolean_filter_false) == "BooleanFilter(False)"
+
+    assert eval(repr(boolean_filter_true)) == boolean_filter_true
+    assert eval(repr(boolean_filter_false)) == boolean_filter_false
+
+
+def test_boolean_filter_from_string():
+    assert BooleanFilter.from_string(None, "true") == BooleanFilter(True)
+    assert BooleanFilter.from_string(None, "false") == BooleanFilter(False)
+    with raises(errors.IllegalParameterError):
```
MrCreosote, line 166: Can you test the exception message?

test/src/service/filtering/filters_test.py
```diff
@@ -132,6 +134,51 @@ def test_rangefilter_from_string_fail():
...
+        BooleanFilter.from_string(None, "invalid")
+
+
+def test_boolean_filter_to_arangosearch_aql(boolean_filter_true, boolean_filter_false):
+    identifier = "test_identifier"
+    var_prefix = "test_prefix"
```
MrCreosote, line 172: I would
a) make these different for the true and false cases, and
b) inline them to reduce recapitulation of the code under test

src/service/filtering/filters.py
MrCreosote: :+1:

MrCreosote: :+1:

src/service/filtering/filters.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/common/product_models/heatmap_common_models.py
Tianhao-Gu: 👍

test/src/service/filtering/filters_test.py
Tianhao-Gu: 👍

test/src/service/filtering/filters_test.py
Tianhao-Gu: 👍

MrCreosote: :+1:

test/src/service/filtering/filters_test.py
MrCreosote: :+1:

src/common/product_models/heatmap_common_models.py
MrCreosote: :+1:

src/common/product_models/heatmap_common_models.py
```diff
@@ -52,6 +53,29 @@ class ColumnType(str, Enum):
     """ A boolean. """
 
 
+# Maps a heatmap column type to a genome attributes column type
+HEATMAP_TO_ATTRIBS_MAPPING = {
```
MrCreosote, line 57: Does this need to be a public variable or should it be private?

Tianhao-Gu: 👍 make it private.

Tianhao-Gu: 👍 make it private.

MrCreosote: :+1:

MrCreosote (approved):

PR #585, opened 2023-12-19
RE2022-290: Add sample flatting experiment prep pt 1

MrCreosote:
Next: randomly assign genome IDs to sample groups and randomly generate a subset of sample attributes from the GROW data field frequencies


Tianhao-Gu (approved):
👍

PR #586, opened 2023-12-19
Sort heatmaps by their display name

MrCreosote:
rather than the kbase ID, which has even less meaning to the average user

Tested manually


Tianhao-Gu (approved):
👍

PR #603, opened 2024-01-03
RE2022-288: add sample display name

Tianhao-Gu:


MrCreosote:
Well that was trivially easy. And just that change means the name winds up in the `install_files` sample data file?

Tianhao-Gu:
> And just that change means the name winds up in th

yea. The display name will be in the `.prepared_sample` file and parser will pick up the info from that file.

MrCreosote (approved):

PR #605, opened 2024-01-06
RE2022-296: parse KB genome meta

Tianhao-Gu:


src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -136,6 +137,27 @@ def _read_metadata_file(meta_filename: str) -> dict:
     return meta_info
 
 
+def _get_genome_obj_meta(
+        genome_info: list[Any],
+) -> dict[str, Any]:
+    # retrieves workspace genome object metadata and convert it to a dict with parsed values
+
+    kb_genome_meta = genome_info[-1]
```
MrCreosote, line 145: Probably unlikely, but we could possibly add more items to the `object_info` tuple
```suggestion
    kb_genome_meta = genome_info[10]
```

src/loaders/genome_collection/parse_tool_results.py
```diff
@@ -163,6 +185,9 @@ def _update_docs_with_meta_info(res_dict, meta_lookup, check_genome):
 
         res_dict[genome_id].update({names.FLD_UPA_MAP: upa_dict})
 
+        # add Genome WS object metadata info
+        res_dict[genome_id].update(_get_genome_obj_meta(meta_info.get(loader_common_names.GENOME_OBJ_INFO_KEY, [{}])))
```
MrCreosote, line 189: Hm, this makes me a little nervous - if we neglected to redownload old data or forgot to download the genomes this would silently pass

src/common/collection_column_specs/genome_attribs-PMI.yml
```diff
@@ -23,6 +23,21 @@ columns:
      type: string
      filter_strategy: ngram
 
+   - key: kbase_genome_size
```
MrCreosote, line 26: If/when we do display names & categories we'll need updates for these

src/loaders/genome_collection/parse_tool_results.py
MrCreosote: Plus _get_genome_obj_meta will fail even without my change since the [-1] expects a non zero length array

src/loaders/genome_collection/parse_tool_results.py

src/loaders/genome_collection/parse_tool_results.py
Tianhao-Gu: 👍 raise error when we got empty genome meta.

src/common/collection_column_specs/genome_attribs-PMI.yml
Tianhao-Gu: hmm, unclear on what you mean.

MrCreosote: In the mockup discussions we've talked about adding display name and column categories for the fields. We have a list of those for the current fields, but we'll need them for these fields as well

src/loaders/genome_collection/parse_tool_results.py
MrCreosote: I guess this is ok, although in the future it's possible we might have assembly only collections, but I suppose we can deal with that then

MrCreosote (approved):

PR #606, opened 2024-01-09
RE2022-294: relax gtdb version matching

Tianhao-Gu:


MrCreosote:
Let's not do a fuzzy match, let's just make a list of the allowed versions like we discussed in standup

PR #607, opened 2024-01-10
RE2022-294: relax gtdb version matching

Tianhao-Gu:
comparing to fuzzy matching approach, we need to re-active all collections if we do it this way.

### DIFF
--- a/src/service/matchers/lineage_matcher.py
+++ b/src/service/matchers/lineage_matcher.py
@@ -3,9 +3,10 @@ Matches assemblies and genomes to collections based on the GTDB lineage string.
 """
 
 import logging
+import re
+from typing import Any, Self
 
-from pydantic import ConfigDict, BaseModel, Field
-from typing import Any
+from pydantic import ConfigDict, BaseModel, Field, model_validator
 
 from src.common.constants import GTDB_UNCLASSIFIED_PREFIX
 from src.common.gtdb_lineage import GTDBRank, GTDBLineage, GTDBLineageError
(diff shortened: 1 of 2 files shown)


src/service/matchers/lineage_matcher.py
```diff
@@ -3,39 +3,46 @@
...
-    gtdb_version: str = Field(
-        example="207.0",
-        description="The GTDB version of the collection in which the matcher is installed. " +
-            "Input data to the matcher must match this version of GTDB or the match will " +
+    gtdb_versions: list[str] = Field(
+        example=["214.0", "214.1"],
+        description="The allowed GTDB versions of the collection in which the matcher is installed. " +
+            "Input data to the matcher must match one of the allowed versions of GTDB or the match will " +
             "abort.",
-        pattern=r"^\d{2,4}\.\d{1,2}$"  # giving a little room for expansion
```
MrCreosote, line 35: https://stackoverflow.com/questions/58098925/use-regex-in-pydantic-model-that-works-with-set-of-strings

src/service/matchers/lineage_matcher.py
```diff
@@ -3,39 +3,46 @@
...
+_COLLECTION_PARAMS_GTDB_VERSIONS_KEY = "gtdb_versions"
 
 
 class GTDBLineageMatcherCollectionParameters(BaseModel):
     "Parameters for the GTDB lineage matcher."
-    gtdb_version: str = Field(
-        example="207.0",
-        description="The GTDB version of the collection in which the matcher is installed. " +
-            "Input data to the matcher must match this version of GTDB or the match will " +
+    gtdb_versions: list[str] = Field(
```
MrCreosote, line 30: I was thinking we could allow just a single string as well to allow for backwards compatibility... but maybe that's more trouble than it's worth

MrCreosote: And it would mean we don't have to redo the old collections and the old versions will still work... eh, maybe it is worth doing. We could keep the old field and document it as deprecated. Only one field should have a non-null value. What do you think?

src/service/matchers/lineage_matcher.py
Tianhao-Gu: 👍

src/service/matchers/lineage_matcher.py
Tianhao-Gu: if it's just for backwards compatibility, I don' t think it's worth it. There have been numerous recent changes (samples flattening, WS genome meta) that don't support backward compatibility.

MrCreosote: True, but this particular change will break everything since the collection docs themselves won't pass validation once this change goes live. We won't even be able to pull the old collections. But fair enough

src/service/matchers/lineage_matcher.py
```diff
@@ -3,36 +3,35 @@
...
+_COLLECTION_PARAMS_GTDB_VERSIONS_KEY = "gtdb_versions"
 
 
 class GTDBLineageMatcherCollectionParameters(BaseModel):
     "Parameters for the GTDB lineage matcher."
-    gtdb_version: str = Field(
-        example="207.0",
-        description="The GTDB version of the collection in which the matcher is installed. " +
-            "Input data to the matcher must match this version of GTDB or the match will " +
+    gtdb_versions: Set[constr(pattern=r"^\d{2,4}\.\d{1,2}$")] = Field(
```
MrCreosote, line 30: You can't specify a set in JSON so I don't think this will work

src/service/matchers/lineage_matcher.py
```diff
@@ -3,36 +3,35 @@
...
+        example={"214.0", "214.1"},
```
MrCreosote, line 31: can we specify that the list has to be > 0 items long? I think theres a conlist function that'll do that

src/service/matchers/lineage_matcher.py
Tianhao-Gu: 👍 revert to use list.

src/service/matchers/lineage_matcher.py
Tianhao-Gu: 👍

MrCreosote: Is it not possible to use a constr inside a conlist?

MrCreosote (approved):
LGTM. I'm assuming you fired up the server and checked that violating the matcher input contstraints threw errors and the matcher worked, yeah?

Tianhao-Gu:
> LGTM. I'm assuming you fired up the server and checked that violating the matcher input contstraints threw errors and the matcher worked, yeah?

yea. I tested locally. I will test on CI server for sure.

PR #609, opened 2024-01-10
fix no lineages error

Tianhao-Gu:


MrCreosote:
Hm, I'm thinking what should happen here is that the match should complete but show no matches rather than fail outright. Maybe we should discuss in the re team slack

Tianhao-Gu:
> Hm, I'm thinking what should happen here is that the match should complete but show no matches rather than fail outright. Maybe we should discuss in the re team slack

👍

src/service/data_products/genome_attributes.py
```diff
@@ -653,6 +653,13 @@ async def _mark_gtdb_matches_STARTS_WITH_strategy(
     # later
     # could also probably DRY up this and the above method
     # retries?
+    if not lineages:
+        # return no matches
+        await storage.update_match_state(
+            internal_match_id, models.ProcessState.COMPLETE, now_epoch_millis(), []
+        )
+        return
```
MrCreosote, line 661: You should do this in `perform_gtdb_lineage_match`, otherwise you'll miss some cases since the matcher chooses one of the two *strategy functions

MrCreosote: also maybe make this a function and use that function wherever the same code is called

Tianhao-Gu: 👍 I moved it to `perform_gtdb_lineage_match`. It's just calling `storage.update_match_state`. Making this single function call a function?

MrCreosote: Fair enough, I guess you'd only be getting rid of a couple args, and the millis function probably shouldn't be abstracted out anyway since it's not mockable and needs to be changed

MrCreosote: I guess it is mockable by monkey patching but I hate that

MrCreosote (approved):

PR #612, opened 2024-01-11
RE2022-238: update column model to include display name etc.

Tianhao-Gu:


src/common/product_models/columnar_attribs_common_models.py
```diff
@@ -66,6 +79,18 @@ class AttributesColumnSpec(BaseModel):
         description="The filter strategy for the column if any. Not all column types need "
             + "a filter strategy."
     )] = None
+    display_name: Annotated[str | None, Field(
+        example="Completeness",
+        description="The display name of the column."
+    )] = None
+    category: Annotated[ColumnCategory | None, Field(
```
MrCreosote, line 86: I would make this a string so that non-techie people can add new categories as they wish.

src/common/product_models/columnar_attribs_common_models.py
```diff
@@ -66,6 +79,18 @@ class AttributesColumnSpec(BaseModel):
         description="The filter strategy for the column if any. Not all column types need "
             + "a filter strategy."
     )] = None
+    display_name: Annotated[str | None, Field(
```
MrCreosote, line 82: I think that display name and category should be required, right?

src/common/product_models/columnar_attribs_common_models.py
Tianhao-Gu: 👍

src/common/product_models/columnar_attribs_common_models.py
Tianhao-Gu: there are genome attribute columns such as `_mtchsel`, `load_ver` and `coll` doesn't have display name and category. 
We also convert heatmap column to genome attrib col for heatmap filtering which doesn't require display_name and category fields.

MrCreosote: > _mtchsel, load_ver and coll doesn't have display name 

We could just put in arbitrary placeholders for those

> We also convert heatmap column to genome attrib col for heatmap filtering which doesn't require display_name and category fields.

Hmm, this is pointing out a code smell since we're making fields that should be required optional because of a filtering implementation detail. We should probably fix this long term, but what I would do for now is just add arbitrary placeholder text as a hack when making the conversion and add a TODO that we should think through how this works later and fix the hack

MrCreosote: maybe something like a non-visible bool flag that's default false. If the flag is set then display name and column are expected to be missing

MrCreosote: Actually maybe that's the way to go right now, and we could also update the genome meta endpoint to automatically remove non-visible columns

src/common/product_models/columnar_attribs_common_models.py
MrCreosote: :+1:

src/common/product_models/columnar_attribs_common_models.py
Tianhao-Gu: 👍

src/common/product_models/columnar_attribs_common_models.py
```diff
@@ -66,6 +66,22 @@ class AttributesColumnSpec(BaseModel):
         description="The filter strategy for the column if any. Not all column types need "
             + "a filter strategy."
     )] = None
+    non_visible: Annotated[bool, Field(
+        example=False,
+        description="Whether the column is visible to the user."
+    )] = False
+    display_name: Annotated[str | None, Field(
+        example="Completeness",
+        description="The display name of the column."
```
MrCreosote, line 75:
```suggestion
        description="The display name of the column. "
            + "Required unless the column is non-visible."
```

src/common/product_models/columnar_attribs_common_models.py
```diff
@@ -66,6 +66,22 @@ class AttributesColumnSpec(BaseModel):
         description="The filter strategy for the column if any. Not all column types need "
             + "a filter strategy."
     )] = None
+    non_visible: Annotated[bool, Field(
+        example=False,
+        description="Whether the column is visible to the user."
```
MrCreosote, line 71:
```suggestion
        description="Whether the column is visible to the user. "
             + "If True, the display name and category fields are not required"
```

src/common/product_models/columnar_attribs_common_models.py
```diff
@@ -66,6 +66,22 @@ class AttributesColumnSpec(BaseModel):
...
+    )] = False
+    display_name: Annotated[str | None, Field(
+        example="Completeness",
+        description="The display name of the column."
+    )] = None
+    category: Annotated[str | None, Field(
+        example="Quality",
+        description="The category of the column."
```
MrCreosote, line 79:
```suggestion
        description="The category of the column. "
            + "Required unless the column is non-visible."
```

src/common/product_models/columnar_attribs_common_models.py
MrCreosote: Can you update these lines to key off of column visibilty https://github.com/kbase/collections/blob/main/src/service/data_products/genome_attributes.py#L272-L273

Tianhao-Gu: 👍

MrCreosote: I think COLLECTION_KEYS is an unused import at this point

Tianhao-Gu: 👍 removed.

MrCreosote (approved):

PR #614, opened 2024-01-11
RE2022-290: Sample flattening

MrCreosote:
* Update indexes
* Update main sample return method to handle matches and selections
* Update get samples by ID method to handle flattened samples
* Update get missing IDs method to handle flattened samples
* Update get samples by location method to handle flattened samples

### DIFF
--- a/src/service/data_products/samples.py
+++ b/src/service/data_products/samples.py
@@ -470,7 +470,6 @@ async def get_missing_ids(
     selection_id: Annotated[str, Query(description="A selection ID.")] = None,
     user: kb_auth.KBaseUser = Depends(_OPT_AUTH),
 ) -> common_models.DataProductMissingIDs:
-    # TODO SAMPLES this needs to be updated to handle samples correctly
     return await _get_missing_ids(
         app_state.get_app_state(r),
         names.COLL_SAMPLES,
@@ -479,4 +478,5 @@ async def get_missing_ids(
         match_id=match_id,
         selection_id=selection_id,
         user=user,
+        multiple_ids=True,
     )
--- a/src/service/data_products/samples.py
+++ b/src/service/data_products/samples.py
@@ -78,7 +78,7 @@ SAMPLES_SPEC = SamplesSpec(
                 [
                     names.FLD_COLLECTION_ID,
                     names.FLD_LOAD_VERSION,
-                    names.FLD_KB_SAMPLE_ID,
+                    names.FLD_KB_DISPLAY_NAME,
                     # Since this is the default sort option (see below), we specify an index
                     # for fast sorts since every time the user hits the UI for the first time
                     # or without specifying a sort order it'll sort on this field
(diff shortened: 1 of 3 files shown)


MrCreosote:
Interesting codecov is telling me I added uncovered lines in files I didn't change

src/service/data_products/samples.py
```diff
@@ -117,6 +117,7 @@ def _remove_keys(doc):
     doc = remove_collection_keys(remove_arango_keys(doc))
     doc.pop(names.FLD_MATCHES_SELECTIONS, None)
     doc.pop(names.FLD_SAMPLE_GEO, None)
+    doc.pop(names.FLD_KBASE_IDS, None)
```
Tianhao-Gu, line 120: can you remind me why we are removing `FLD_SAMPLE_GEO` and `FLD_KBASE_IDS`?

src/service/data_products/samples.py
```diff
@@ -164,18 +165,19 @@ class Samples(BaseModel):
...
+    # TODO SAMPLES - how should we support creating a selection from samples?
+)
 async def get_samples(
     r: Request,
     collection_id: str = PATH_VALIDATOR_COLLECTION_ID,
-    sort_on: common_models.QUERY_VALIDATOR_SORT_ON = names.FLD_KBASE_ID,
+    sort_on: Annotated[str, Query(
+        example=names.FLD_KB_SAMPLE_ID,
+        description="The field to sort on."
+    )] = names.FLD_KB_SAMPLE_ID,
```
Tianhao-Gu, line 180: Do you think we should sort on `genome_count` by default?

src/service/data_products/samples.py
MrCreosote: GEO is removed because we already have lat / lon so it's redundant. It's just there so we could do geosearch later if we wanted.

I'm removing IDS because there could be hundreds of IDS in that field and so it gets really big and isn't particularly useful for users. We probably need to have a discussion about if / how we want to display the genome <-> sample relationships to users. If we do decide to keep the list it's a 1 liner to put back in.

src/service/data_products/samples.py
MrCreosote: Actually now I think about it we should search on the display name. I forgot about it because it's not in the flattened data, but that's because we haven't redownloaded the genomes because podman is still messed up

src/service/data_products/samples.py
Tianhao-Gu: 👍

src/service/data_products/samples.py
Tianhao-Gu: 👍 I do have a script to manually update the sample.prepare file. I can ensure the data has display_name when we deploy this change.

Tianhao-Gu (approved):
👍

Tianhao-Gu (approved):
LGTM

PR #619, opened 2024-01-12
RE2022-301: add collection metadata

Tianhao-Gu:


MrCreosote (approved):
LGTM, I'm assuming it tested fine

src/service/models.py
```diff
@@ -240,6 +240,10 @@ class Collection(BaseModel):
             }
         )
     ] = None
+    metadata: Annotated[dict[str, Any] | None, Field(
```
MrCreosote, line 243: Actually now I think about it, Any could include container types which I don't think we want
```suggestion
    metadata: Annotated[dict[str, str | int | float] | None, Field(
```

Tianhao-Gu:
> LGTM, I'm assuming it tested fine

yea. The metadata is saved/retrieved correctly locally.

MrCreosote (approved):

PR #623, opened 2024-01-29
RE2022-320: fix spec file bug

Tianhao-Gu:
A bug found during creating search view - spec entries need to be identical across different collections.

### DIFF
--- a/src/common/collection_column_specs/parse_sample_template.py
+++ b/src/common/collection_column_specs/parse_sample_template.py
@@ -15,6 +15,11 @@ usage: parse_sample_template.py [-h] --input_yaml INPUT_YAML --core_yaml CORE_YA
     In other words, the output of this script is to save the user's time when constructing the sample column
     specifications and is not expected to be completely accurate.
 
+    Observed discrepancies in the contextual information for certain sample fields,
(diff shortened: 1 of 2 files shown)


MrCreosote (commented):
What caused this bug?

Tianhao-Gu:
> What caused this bug?

The parsing script I used are from 2 different sample templates (enigma_template.yml and sesar_template.yml). Unfortunately, some fields have different context for the same key.

MrCreosote:
so you just have to manually fix it?

Tianhao-Gu:
> so you just have to manually fix it?

yea. For the current sample counts (only 24), I think manual fix is fine. If we see more samples in the future, it makes sense to update the parsing script to use the unified `metadata_validation.yml` instead of two template files.

MrCreosote:
In that case can you add a warning about this to the docs so people know they need to check the output for sample sources with the same key but different semantics

Tianhao-Gu:
> In that case can you add a warning about this to the docs so people know they need to check the output for sample sources with the same key but different semantics

👍

MrCreosote (approved):

PR #626, opened 2024-01-30
RE2022-320: add update samples filter set to enable filtering

Tianhao-Gu:
example local endpoint call:

```
(dev) tgu@collections (dev_sample_filtering)$curl -X 'GET' \
  'http://127.0.0.1:5000/collections/ENIGMA/data_products/samples/?output_table=false&limit=1&filter_genome_count=%5B3,4%5D' \
  -H 'accept: application/json' \
  -H "Authorization: Bearer $token" | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   487  100   487    0     0   1407      0 --:--:-- --:--:-- --:--:--  1419
{
    "match_state": null,
    "selection_state": null,
    "skip": 0,
    "limit": 1,
    "fields": null,
    "table": null,
    "data": [
        {
            "enigma:date": "2014-06-17",
            "enigma:experiment_name": "Groundwater",
            "enigma:time_zone": "UTC\u221204:00",
            "enigma:well_name": "DP16D",
            "env_package": "water",
            "genome_count": 3,
            "kbase_display_name": "DP16D-06-17-14",
            "kbase_sample_id": "11dd1e1e-c721-460b-9cc3-a517784fcdb8",
            "latitude": 35.97550000000001,
            "longitude": -84.27467799999998,
            "material": "ENVO:01001004",
            "sample_template": "ENIGMA"
        }
    ],
    "count": null
}
```

### DIFF
--- a/src/service/data_products/samples.py
+++ b/src/service/data_products/samples.py
@@ -241,7 +241,7 @@ async def get_samples(
     # we have a max limit of 1000, which means sorting is O(n log2 1000).
     # Otherwise we need indexes for every sort
     appstate = app_state.get_app_state(r)
-    load_ver, dp_match, dp_sel = await get_load_version_and_processes(
+    load_ver, dp_match, dp_sel, coll = await get_load_version_and_processes(
         appstate,
         user,
         names.COLL_SAMPLES,
@@ -254,7 +254,7 @@ async def get_samples(
     )
     if status_only:
         return _response(dp_match=dp_match, dp_sel=dp_sel)
-    coll = await appstate.arangostorage.get_collection_active(collection_id)
+
     filters = await get_filters(
         r,
         names.COLL_SAMPLES,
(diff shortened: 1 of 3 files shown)


src/service/data_products/samples.py
```diff
@@ -239,17 +254,28 @@ async def get_samples(
     )
     if status_only:
         return _response(dp_match=dp_match, dp_sel=dp_sel)
-    filters = FilterSet(
+    coll = await appstate.arangostorage.get_collection_active(collection_id)
```
MrCreosote, line 257: There's a subtle bug here - in the case of a load version override, a collection may not even exit yet with that load version, let alone be the active collection. In that case, `coll` should be `None` and if any filter parameters are supplied `get_filters` will throw an error. 

In `genome_attribs` this works correctly, but here, if a load ver override is supplied, coll will be pulled anyway and I'm not quite sure what'll happen. At the very least the data pulled may not be the data expected.

`get_load_version_and_processes` should probably be updated to return coll, then things should work ok, but that method is already pretty complex... what I'd really like to do is rethink the data product code structure in the service but there's no time for that

Tianhao-Gu: 👍 updated `get_load_version_and_processes` to return `coll`.

MrCreosote: At this point that function should probably return a named tuple or something, but this'll do

MrCreosote: :+1:

MrCreosote (approved):

PR #644, opened 2024-02-02
Give codecov v4 another chance

MrCreosote:
Maybe it was just codecov being codecov, as reverting to 3 didn't fix things immediately either


MrCreosote:
8 consistent failures over a couple hours = v4 is broke

PR #646, opened 2024-02-05
RE2022-327: add default sort taxa count

Tianhao-Gu:
example local endpoint return:

default sort:
```
(dev) tgu@collections (dev_taxa_count_ordering)$curl -X 'GET' \
  "$server_url/collections/PMI/data_products/taxa_count/counts/$rank/?match_id=$match_id&status_only=false" \
  -H 'accept: application/json' \
  -H "Authorization: Bearer $token" | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   333  100   333    0     0  26232      0 --:--:-- --:--:-- --:--:-- 33300
{
    "match_state": "complete",
    "selection_state": null,
    "data": [
        {
            "name": "Actinomycetota",
            "count": 82,
            "match_count": 82,
            "sel_count": null
        },
        {
            "name": "Pseudomonadota",
            "count": 367,
            "match_count": 0,
            "sel_count": null
        },
        {
            "name": "Bacteroidota",
            "count": 53,
            "match_count": 0,
            "sel_count": null
        },
        {
            "name": "Bacillota",
            "count": 48,
            "match_count": 0,
            "sel_count": null
        }
    ]
}
```

specify sort order:
```
(dev) tgu@collections (dev_taxa_count_ordering)$sort_priority=standard%2C%20matched
(dev) tgu@collections (dev_taxa_count_ordering)$curl -X 'GET' \
  "$server_url/collections/PMI/data_products/taxa_count/counts/$rank/?match_id=$match_id&status_only=false&sort_priority=$sort_priority" \
  -H 'accept: application/json' \
  -H "Authorization: Bearer $token" | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   333  100   333    0     0  25848      0 --:--:-- --:--:-- --:--:-- 33300
{
    "match_state": "complete",
    "selection_state": null,
    "data": [
        {
            "name": "Pseudomonadota",
            "count": 367,
            "match_count": 0,
            "sel_count": null
        },
        {
            "name": "Actinomycetota",
            "count": 82,
            "match_count": 82,
            "sel_count": null
        },
        {
            "name": "Bacteroidota",
            "count": 53,
            "match_count": 0,
            "sel_count": null
        },
        {
            "name": "Bacillota",
            "count": 48,
            "match_count": 0,
            "sel_count": null
        }
    ]
}
```

### DIFF
--- a/src/service/data_products/taxa_count.py
+++ b/src/service/data_products/taxa_count.py
@@ -226,6 +226,7 @@ async def get_taxa_counts(
             + "Note that if a selection ID is set, any load version override is ignored."),
     sort_priority: str = Query(
         default=None,
+        example="standard",  # or make a constant for this. That's probably better
         description=f"A comma separated list of sort priorities. Valid values are: "
                     f"{', '.join(_SORT_PRIORITY_ORDER_MAP.keys())}. ",
     ),
--- a/src/service/data_products/taxa_count.py
+++ b/src/service/data_products/taxa_count.py
@@ -44,6 +44,15 @@ _TYPE2PREFIX = {
 
 _MAX_COUNT = 20  # max number of taxa count records to return
 
+# TaxaCount fields. These need to match the field names in the TaxaCount class below
+_FLD_TAXA_COUNT_MATCH_COUNT = "match_count"
+_FLD_TAXA_COUNT_SEL_COUNT = "sel_count"
+
+# The default sorting order for taxa counts results
+_DEFAULT_SORT_ORDER = [names.FLD_TAXA_COUNT_COUNT,
+                       _FLD_TAXA_COUNT_MATCH_COUNT,
+                       _FLD_TAXA_COUNT_SEL_COUNT]
+
 
 class TaxaCountSpec(DataProductSpec):
 
@@ -134,8 +143,8 @@ class TaxaCount(BaseModel):
 
 # these need to match the field names in TaxaCount above
 _TYPE2FIELD = {
-    models.SubsetType.MATCH: "match_count",
-    models.SubsetType.SELECTION: "sel_count",
+    models.SubsetType.MATCH: _FLD_TAXA_COUNT_MATCH_COUNT,
+    models.SubsetType.SELECTION: _FLD_TAXA_COUNT_SEL_COUNT,
 }
 
 
@@ -234,7 +243,45 @@ async def get_taxa_counts(
         await _add_subset_data_in_place(q, store, collection_id, load_ver, rank, dp_proc)
         # For now always sort by the std data. See if ppl want sort by match/selection data
         # before implementing something more sophisticated.
+    _sort_taxa_counts(q)
+
     return _taxa_counts(dp_match=dp_match, dp_sel=dp_sel, data=q)
+
+
+def _sort_dict_list(
+        dict_list: list[dict],
+        key: str,
+        reverse: bool = True):
+    # Sort taxa count records, a list of dicts, in place by the key of dictionary with None values last.
+    # Key values should be either numeric or None.
+    # Certain keys, such as 'sel_count' or 'match_count,' may not exist in all dictionaries.
+    # For instance, these keys might be absent if the corresponding match or selection process did not occur/complete.
+    # In such cases, the default value of float('-inf') is used to ensure that the dictionaries are sorted correctly.
+    dict_list.sort(key=lambda x: (int(x.get(key)) if x.get(key) is not None else float('-inf')), reverse=reverse)
+
+
+def _fill_missing_orders(sort_order: list[str]):
+    # fill in missing orders with the default sort order
+    if isinstance(sort_order, list):
+        raise ValueError(f"sort_order must be a list of strings, provided: {sort_order}")
+
+    return sort_order + [order for order in _DEFAULT_SORT_ORDER if order not in sort_order]
+
+
+def _sort_taxa_counts(
+        q: list[dict],
+        sort_order: list[str] = None):
+    # sort the taxa counts result in place by the sort order
+    sort_order = _fill_missing_orders(sort_order) if sort_order else _DEFAULT_SORT_ORDER
+
+    # Remove the `count` order if it is the first element since the records are already sorted by count in the
+    # _query function.
+    sort_order.pop(0) if sort_order[0] == names.FLD_TAXA_COUNT_COUNT else None
+
+    for k in sort_order:
+        if k not in _DEFAULT_SORT_ORDER:
+            raise errors.IllegalParameterError(f"Invalid sort key: {k}")
+        _sort_dict_list(q, k)
 
 
 def _taxa_counts(
--- a/src/service/data_products/taxa_count.py
+++ b/src/service/data_products/taxa_count.py
@@ -261,18 +261,23 @@ async def get_taxa_counts(
     return _taxa_counts(dp_match=dp_match, dp_sel=dp_sel, data=q)
 
 
-def _fill_missing_orders(sort_order: list[str]):
-    # fill in missing orders with the default sort order
+def _fill_missing_orders(sort_order: list[str], processed_count: list[str]):
+    # fill in missing orders with the default sort order if order exists in the processed count
     if not isinstance(sort_order, list):
         raise ValueError(f"sort_order must be a list of strings, provided: {sort_order}")
 
-    return [order for order in _DEFAULT_SORT_ORDER if order not in sort_order] + sort_order
+    missing_orders = [order for order in _DEFAULT_SORT_ORDER
+                      if order not in sort_order and
+                      (order in processed_count or order == names.FLD_TAXA_COUNT_COUNT)]
+
+    return missing_orders + sort_order
 
 
 def _sort_taxa_counts(
         q: list[dict[str, Any]],
         sort_priority: str,
-        dp_list: list[models.DataProductProcess]):
+        dp_list: list[models.DataProductProcess],
+        sort_by_count_already: bool = True):
     # Sort taxa count records in place by the sort_priority list.
 
     processed_count = [_TYPE2FIELD[dp.type] for dp in dp_list if dp]
@@ -297,12 +302,18 @@ def _sort_taxa_counts(
                                            f"taxa count records")
 
     # fill in missing orders with the default precedence order
-    sort_order = _fill_missing_orders(sort_order_rev[::-1])
-    # remove any sort orders that are not processed (i.e. match or selection counts)
-    sort_order = [order for order in sort_order if order in processed_count or order == names.FLD_TAXA_COUNT_COUNT]
+    sort_order = _fill_missing_orders(sort_order_rev[::-1], processed_count)
+
+    if sort_by_count_already:
+        # remove unnecessary sort order if records are already sorted by count
+        sort_order.pop(0) if sort_order[0] == names.FLD_TAXA_COUNT_COUNT else None
 
     for k in sort_order:
         q.sort(key=lambda x: x[k], reverse=True)
+
+    if not sort_by_count_already:
+        # indicating that we have added more taxa count records in _add_subset_data_in_place step.
+        q[:] = q[:_MAX_COUNT]
 
 
 def _taxa_counts(
--- a/src/service/data_products/taxa_count.py
+++ b/src/service/data_products/taxa_count.py
@@ -265,7 +265,7 @@ def _fill_missing_orders(sort_order: list[str]):
     if not isinstance(sort_order, list):
         raise ValueError(f"sort_order must be a list of strings, provided: {sort_order}")
 
-    return sort_order + [order for order in _DEFAULT_SORT_ORDER if order not in sort_order]
+    return [order for order in _DEFAULT_SORT_ORDER if order not in sort_order] + sort_order
 
 
 def _sort_taxa_counts(
--- a/src/service/data_products/taxa_count.py
+++ b/src/service/data_products/taxa_count.py
@@ -261,17 +261,6 @@ async def get_taxa_counts(
     return _taxa_counts(dp_match=dp_match, dp_sel=dp_sel, data=q)
 
 
-def _sort_dict_list(
-        dict_list: list[dict],
-        key: str):
-    # Sort taxa count records, a list of dicts, in place by the key of dictionary with None values last.
-    # Key values should be either integer or None.
-    # Certain keys, such as 'sel_count' or 'match_count,' may not exist in all dictionaries.
-    # For instance, these keys might be absent if the corresponding match or selection process did not occur/complete.
-    # In such cases, the default value of float('-inf') is used to ensure that the dictionaries are sorted correctly.
-    dict_list.sort(key=lambda x: x.get(key, _INF_NEG), reverse=True)
-
-
 def _fill_missing_orders(sort_order: list[str]):
     # fill in missing orders with the default sort order
     if not isinstance(sort_order, list):
@@ -306,7 +295,7 @@ def _sort_taxa_counts(
         sort_order = [names.FLD_TAXA_COUNT_COUNT]
 
     for k in sort_order:
-        _sort_dict_list(q, k)
+        q.sort(key=lambda x: x[k], reverse=True)
 
 
 def _taxa_counts(
--- a/src/service/data_products/taxa_count.py
+++ b/src/service/data_products/taxa_count.py
@@ -269,14 +269,10 @@ def _fill_missing_orders(sort_order: list[str]):
 
 
 def _sort_taxa_counts(
-        q: list[dict],
+        q: list[dict[str, Any]],
         sort_order: list[str] = None):
     # sort the taxa counts result in place by the sort order
     sort_order = _fill_missing_orders(sort_order) if sort_order else _DEFAULT_SORT_ORDER
-
-    # Remove the `count` order if it is the first element since the records are already sorted by count in the
-    # _query function.
-    sort_order.pop(0) if sort_order[0] == names.FLD_TAXA_COUNT_COUNT else None
 
     for k in sort_order:
         if k not in _DEFAULT_SORT_ORDER:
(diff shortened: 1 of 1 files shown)


src/service/data_products/taxa_count.py
```diff
@@ -234,9 +243,43 @@ async def get_taxa_counts(
...
+    # In such cases, the default value of float('-inf') is used to ensure that the dictionaries are sorted correctly.
+    dict_list.sort(key=lambda x: (int(x.get(key)) if x.get(key) is not None else float('-inf')), reverse=reverse)
+
+
+def _fill_missing_orders(sort_order: list[str]):
+    # fill in missing orders with the default sort order
+    if not isinstance(sort_order, list):
+        raise ValueError(f"sort_order must be a list of strings, provided: {sort_order}")
+
+    return [order for order in _DEFAULT_SORT_ORDER if order not in sort_order] + sort_order
```
MrCreosote, line 268: So this means, IIUC, that if the sort order I provide is "match_count,sel_count" then the sort order will be std->match->sel? That seems wrong, I'd expect match->sel->std.

src/service/data_products/taxa_count.py
```diff
@@ -234,9 +243,43 @@ async def get_taxa_counts(
...
+def _sort_dict_list(
+        dict_list: list[dict],
+        key: str,
+        reverse: bool = True):
+    # Sort taxa count records, a list of dicts, in place by the key of dictionary with None values last.
+    # Key values should be either numeric or None.
+    # Certain keys, such as 'sel_count' or 'match_count,' may not exist in all dictionaries.
+    # For instance, these keys might be absent if the corresponding match or selection process did not occur/complete.
+    # In such cases, the default value of float('-inf') is used to ensure that the dictionaries are sorted correctly.
+    dict_list.sort(key=lambda x: (int(x.get(key)) if x.get(key) is not None else float('-inf')), reverse=reverse)
```
MrCreosote, line 260: Is this a stable sort?

src/service/data_products/taxa_count.py
```diff
@@ -234,9 +243,43 @@ async def get_taxa_counts(
...
         # before implementing something more sophisticated.
+    _sort_taxa_counts(q)
+
     return _taxa_counts(dp_match=dp_match, dp_sel=dp_sel, data=q)
 
 
+def _sort_dict_list(
+        dict_list: list[dict],
+        key: str,
+        reverse: bool = True):
```
MrCreosote, line 254: It looks like this is never used by calling code and I don't see a use case for it - is it necessary?

src/service/data_products/taxa_count.py
MrCreosote, line 260: Maybe make `float("-inf")` a constant

MrCreosote, line 260: Why is the int conversion necessary?

Tianhao-Gu: 👍

Tianhao-Gu: yea. `.sort` is stable. https://docs.python.org/3.11/howto/sorting.html#decorate-sort-undecorate

stable or not matters here?

Tianhao-Gu: I just want to catch in case the field value cannot be casted to int which should not happen. But just in case.

src/service/data_products/taxa_count.py
Tianhao-Gu: removed.

src/service/data_products/taxa_count.py
Tianhao-Gu: `sort order` is the order we sort. I think that's correct. If you provide ['match_count', 'sel_cout'], the sorting function will sort 'match_count' first, then 'sel_cout'. I am planning to introduce `sort_priority` as endpoint input. And something making `sort_order = sort_priority[::-1]`. Thus, if user inputs `['matched', 'selected']`, the sort_order is assigned to `['count', 'sel_count', 'match_count']`.

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
MrCreosote: Ok, I wasn't thinking this through correctly. This function is returning the *reverse* of the users's desired sort so that the last sort is the user's first priority.

MrCreosote: That could use a comment in the code I think

src/service/data_products/taxa_count.py
MrCreosote: If you're doing multiple sorts and want the earlier sorts to be reflected in the final sort then the sort needs to be stable, yeah. Otherwise there's no guarantee that sort #1 won't be completely undone by sort #2

MrCreosote: Alright, can you add a comment explaining why it's there?

MrCreosote: Wait, hold on - do you want it to throw an error? Because floats will just get silently truncated which seems wrong

src/service/data_products/taxa_count.py
MrCreosote: Hold on again - shouldn't sort_order be reversed in that case?

MrCreosote: And _DEFAULT_SORT_ORDER for that matter

Tianhao-Gu: No. I was planning to call this func with reverse of user’s priority.

MrCreosote: Then at  _DEFAULT_SORT_ORDER needs to be reversed too, right? Otherwise the default sort order is really match->sel->count

Tianhao-Gu: No. I don't think so. Why?

src/service/data_products/taxa_count.py
Tianhao-Gu: I see. yea. `.sort` is stable.

src/service/data_products/taxa_count.py
MrCreosote: Ok, I finally grok this. I managed to confuse myself on part of it, but I think there are two confusing parts about the implementation:

* I would reverse the user's sort order inside the function rather than in the calling function, unless there a reason that it's important it's in the calling function
* Sorts happen on the match and selection fields even if there's no data to sort, so the sort is effectively a no op and a waste of cycles. I don't think the latter is significant but the next person that comes along might, like me, go "why the heck is match in the sort order when no match was provided?"  I would write things so that the sort order only contains, and therefore sorts only occur on, data that is actually in the data structure

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍 it makes `_sort_dict_list` unnecessarily complicated. Given our use case, it's probably not worth it.

MrCreosote: :+1:

MrCreosote: Sorry, what are you saying isn't worth it?

MrCreosote: And should floats be allowed here since these are counts?

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍 updated func to take sort_priority and only sort on existing processes.

src/service/data_products/taxa_count.py
Tianhao-Gu: updated to  `x[k]` since key should always exist now.

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
MrCreosote: :

src/service/data_products/taxa_count.py
```diff
@@ -208,7 +226,12 @@ async def get_taxa_counts(
             + "Note that if a selection ID is set, any load version override is ignored."),
     status_only: QUERY_VALIDATOR_STATUS_ONLY = False,
     load_ver_override: QUERY_VALIDATOR_LOAD_VERSION_OVERRIDE = None,
-    user: kb_auth.KBaseUser = Depends(_OPT_AUTH)
+    user: kb_auth.KBaseUser = Depends(_OPT_AUTH),
+    sort_priority: list[str] = Query(
+        default=list(_SORT_PRIORITY_ORDER_MAP.keys()),
+        description="Specify the priority order for sorting taxa counts based on the provided fields and "
+                    "their corresponding order.",
```
MrCreosote, line 233: I would put this prior to status_only as it's a control users are likely to use, not just admins or stuff that's only really of interest to UI coders

src/service/data_products/taxa_count.py
```diff
@@ -208,7 +226,12 @@ async def get_taxa_counts(
             + "Note that if a selection ID is set, any load version override is ignored."),
     status_only: QUERY_VALIDATOR_STATUS_ONLY = False,
     load_ver_override: QUERY_VALIDATOR_LOAD_VERSION_OVERRIDE = None,
-    user: kb_auth.KBaseUser = Depends(_OPT_AUTH)
+    user: kb_auth.KBaseUser = Depends(_OPT_AUTH),
+    sort_priority: list[str] = Query(
+        default=list(_SORT_PRIORITY_ORDER_MAP.keys()),
```
MrCreosote, line 231: The way that fastapi handles lists as query parameters, IIRC, is that for every list entry you have to put another k/v pair into the query like `?sort_priority=match&sort_priority=sel`. The way I've done it in other places is just make it a string and split on commas

src/service/data_products/taxa_count.py
```diff
@@ -232,11 +255,49 @@ async def get_taxa_counts(
...
+    # Sort taxa count records in place by the sort_priority list.
+    processed_count = [_TYPE2FIELD[dp.type] for dp in dp_list if dp]
+
+    if processed_count:
+
+        if len(sort_priority) != len(set(sort_priority)):
+            raise errors.IllegalParameterError(f"Duplicate sort priority found: {sort_priority}")
+
+        try:
+            sort_order_rev = [_SORT_PRIORITY_ORDER_MAP[k] for k in sort_priority]
```
MrCreosote, line 285:
```suggestion
            sort_order_rev = [_SORT_PRIORITY_ORDER_MAP[k] for k in sort_priority
                                         if _SORT_PRIORITY_ORDER_MAP[k] in processed_count]
```
Or we could throw an error if the user asks for a sort where the order includes a match or selection they didn't provide

src/service/data_products/taxa_count.py
```diff
@@ -232,11 +255,49 @@ async def get_taxa_counts(
...
+        except KeyError as e:
+            raise errors.IllegalParameterError(f"Invalid sort priority: {e}, "
+                                               f"valid priorities are: {list(_SORT_PRIORITY_ORDER_MAP.keys())}")
+
+        # fill in missing orders with the default precedence order
+        sort_order = _fill_missing_orders(sort_order_rev[::-1])
+        # remove any sort orders that are not processed (i.e. match or selection counts)
+        sort_order = [order for order in sort_order if order in processed_count or order == names.FLD_TAXA_COUNT_COUNT]
```
MrCreosote, line 293: Assuming the above change, the only thing that can be missing at this point is the regular count, right? So can't you just do something like

```
if count_str not in sort_order_rev:
    sort_order_rev += [count_str]
sort_order = sort_order_rev.reverse()
```

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍

src/service/data_products/taxa_count.py
Tianhao-Gu: if we are doing that, we are not enforcing the default precedence right? For example, if `sort_priority='count'`, we still want to sort records by selected then by matched.

src/service/data_products/taxa_count.py
Tianhao-Gu: I think we want to set sort_priority to "selected, matched, standard" by default. That way, we shouldn't raise if the process hasn't started.

Tianhao-Gu: 👍 set sort_priority to None by default and raise error if sort_priority has count that's not processed.

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
MrCreosote: Yep, you're right

src/service/data_products/taxa_count.py
```diff
@@ -232,11 +255,67 @@ async def get_taxa_counts(
...
+                      (order in processed_count or order == names.FLD_TAXA_COUNT_COUNT)]
+
+    return missing_orders + sort_order
+
+
+def _sort_taxa_counts(
+        q: list[dict[str, Any]],
+        sort_priority: str,
+        dp_list: list[models.DataProductProcess],
+        sort_by_count_already: bool = True):
```
MrCreosote, line 280: This isn't currently used - what's the use case for it?

src/service/data_products/taxa_count.py
```diff
@@ -232,11 +255,67 @@ async def get_taxa_counts(
     q = await _query(store, collection_id, load_ver, rank)
     for dp_proc in [dp_match, dp_sel]:
         await _add_subset_data_in_place(q, store, collection_id, load_ver, rank, dp_proc)
-        # For now always sort by the std data. See if ppl want sort by match/selection data
-        # before implementing something more sophisticated.
+
+    _sort_taxa_counts(q, sort_priority, [dp_match, dp_sel])
```
MrCreosote, line 259: I'm assuming the next PR is to update _add_subset_data_in_place so it doesn't throw away match & selection data?

src/service/data_products/taxa_count.py
```diff
@@ -232,11 +255,67 @@ async def get_taxa_counts(
...
+    try:
+        sort_order_rev = [_SORT_PRIORITY_ORDER_MAP[k] for k in sort_priority]
+    except KeyError as e:
+        raise errors.IllegalParameterError(f"Invalid sort priority: {e}, "
+                                           f"valid priorities are: {list(_SORT_PRIORITY_ORDER_MAP.keys())}")
+    missing_processes = [p for p in sort_order_rev if p not in processed_count and p != names.FLD_TAXA_COUNT_COUNT]
+    if missing_processes:
+        missing_priority = [key for key, value in _SORT_PRIORITY_ORDER_MAP.items() if value in missing_processes]
+        raise errors.IllegalParameterError(f"No occurrences of the {missing_priority} data were found in the "
+                                           f"taxa count records")
```
MrCreosote, line 302: This is a confusing error message - the error is that the user put match or selection in the sort priority without providing a match or selection ID

Tianhao-Gu: 👍

src/service/data_products/taxa_count.py
Tianhao-Gu: yea. That's the plan.

src/service/data_products/taxa_count.py
Tianhao-Gu: I intended to us it for situations where we found missing records linked to a match/selection and attach them to `q` causes the original sorting of `q` by count is no longer accurate.

MrCreosote: Ok, so it's a small optimization to avoid a resort when the match and select bars are already represented in the standard count bars, right?

src/service/data_products/taxa_count.py
MrCreosote: :

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
Tianhao-Gu: that's correct. If no new records are added, we don't necessarily need to resort by `count` since it's already sorted by `count`.

src/service/data_products/taxa_count.py
```diff
@@ -44,6 +44,24 @@
...
 _MAX_COUNT = 20  # max number of taxa count records to return
 
+# TaxaCount fields. These need to match the field names in the TaxaCount class below
+_FLD_TAXA_COUNT_MATCH_COUNT = "match_count"
+_FLD_TAXA_COUNT_SEL_COUNT = "sel_count"
+
+# The default sorting order for taxa counts results
+_DEFAULT_SORT_ORDER = [names.FLD_TAXA_COUNT_COUNT,
+                       _FLD_TAXA_COUNT_MATCH_COUNT,
+                       _FLD_TAXA_COUNT_SEL_COUNT]
```
MrCreosote, line 54: Technically this is just `list(_SORT_PRIORITY_VALUE_MAP.values()).reverse()`

src/service/data_products/taxa_count.py
```diff
@@ -206,6 +224,11 @@ async def get_taxa_counts(
         default = None,
         description="A selection ID to include the selection count in the taxa count data. "
             + "Note that if a selection ID is set, any load version override is ignored."),
+    sort_priority: str = Query(
+        default=None,
+        description=f"A comma separated list of sort priorities. Valid values are: "
```
MrCreosote, line 229:
```suggestion
        default=None,
        example="standard",  # or make a constant for this. That's probably better
        description=f"A comma separated list of sort priorities. Valid values are: "
```

src/service/data_products/taxa_count.py
MrCreosote: :+1:

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍

MrCreosote: :+1:

MrCreosote (approved):

PR #648, opened 2024-02-06
RE2022-327: add missing records from match/selection

Tianhao-Gu:


src/service/data_products/taxa_count.py
```diff
@@ -250,10 +250,13 @@ async def get_taxa_counts(
     if rank not in ranks.data:
         raise errors.IllegalParameterError(f"Invalid rank: {rank}")
     q = await _query(store, collection_id, load_ver, rank)
+    retrieved_new = False
     for dp_proc in [dp_match, dp_sel]:
-        await _add_subset_data_in_place(q, store, collection_id, load_ver, rank, dp_proc)
+        retrieved_new_proc = await _add_subset_data_in_place(q, store, collection_id, load_ver, rank, dp_proc)
+        if retrieved_new_proc:
+            retrieved_new = True
```
MrCreosote, line 257: retrieved_new
```suggestion
        retrieved_new = await _add_subset_data_in_place(
            q, store, collection_id, load_ver, rank, dp_proc) or retrieved_new

```

src/service/data_products/taxa_count.py
```diff
@@ -250,10 +250,13 @@ async def get_taxa_counts(
     if rank not in ranks.data:
         raise errors.IllegalParameterError(f"Invalid rank: {rank}")
     q = await _query(store, collection_id, load_ver, rank)
```
MrCreosote, line 252: Can we change `q` to something else? I keep reading it as query rather than what it is, the result

src/service/data_products/taxa_count.py
```diff
@@ -344,9 +348,22 @@ async def _add_subset_data_in_place(
         )
         name, count = names.FLD_TAXA_COUNT_NAME, names.FLD_TAXA_COUNT_COUNT
         mqd = {d[name]: d[count] for d in matchq}
+
+        missing_names = set(mqd.keys()) - set(d[name] for d in q)
```
MrCreosote, line 352: There's a potential optimization here to skip this step if the last sort is `standard` since any missing names will be cut off of the end of the list, but I think we can worry about that when it becomes an issue

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍

src/service/data_products/taxa_count.py
Tianhao-Gu: 👍 added a todo item. Given the size of taxa count I think it should be okay for now.

src/service/data_products/taxa_count.py
MrCreosote: :+1:

MrCreosote (approved):

PR #650, opened 2024-02-08
RE2022-328: renaming assembly variables

Tianhao-Gu:


test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -98,7 +98,7 @@ def test_update_upload_status_yaml_file(setup_and_teardown):
     params = setup_and_teardown
     assembly_dir = params.assembly_dirs[0]
     assembly_name = ASSEMBLY_NAMES[0]
-    assembly_tuple = workspace_uploader._AssemblyTuple(
+    assembly_tuple = workspace_uploader._WSObjTuple(
```
Xiangs18, line 101: Might change assembly_* to WsObject_* as well while you are at it?

Tianhao-Gu: I think for testing files, we are probably fine since as all those tests are specific to Assembly.

Xiangs18: Okay, we can change later when we test genomes.

MrCreosote: plus variable names should be snake_case in Python

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -88,15 +88,19 @@
 _UPLOADED_YAML = "uploaded.yaml"
 _WS_MAX_BATCH_SIZE = 10000
 
-_AssemblyTuple = namedtuple(
-    "AssemblyTuple",
-    ["assembly_name", "host_assembly_dir", "container_internal_assembly_path"],
+# fields of _WSObjTuple
+_OBJ_NAME = "obj_name"
+_OBJ_DIR = "host_obj_dir"
+_CONTAINER_OBJ_PATH = "container_internal_obj_path"
```
MrCreosote, line 94: These are actually paths to files, not objects, but they get uploaded and linked to from the objects. Maybe s/obj/file/?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -200,44 +204,44 @@ def _get_parser():
...
     Returns the list of workspace UPAs for the created objects in the same order as
-    `assembly_tuples`.
+    `ws_obj_tuples`.
     """
     inputs = [
         {
-            "file": assembly_tuple.container_internal_assembly_path,
-            "assembly_name": assembly_tuple.assembly_name,
+            "file": getattr(obj_tuple, _CONTAINER_OBJ_PATH),
+            "assembly_name": getattr(obj_tuple, _OBJ_NAME),
```
MrCreosote, line 241: The point of making a named tuple is so you can use dot syntax to access its members - why are we using getattr?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -498,36 +504,36 @@ def _upload_assembly_files_in_parallel(
         workspace_id: target workspace id
         load_id: load id
         upload_dir: a directory in collectionssource that creates new directories linking to sourcedata
-        wait_to_upload_assemblies: a dictionary that maps assembly file name to assembly directory
+        wait_to_upload_objs: a dictionary that maps object(assembly) file name to object(assembly) directory
```
MrCreosote, line 507: Why the parentheses here but no where else?

Tianhao-Gu: it's in `_upload_assembly_files_in_parallel` methods. It's specific to Assembly.

Tianhao-Gu: it's in `_upload_assembly_files_in_parallel` methods. It's specific to Assembly.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: > Maybe s/obj/file/?
I didn't get what you mean here.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: in case we need to change member names, I don't need to go through the whole file again.

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: A refactoring tool should be able to do that for you in one shot

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: E.g. host_file_dir, container_internal_file_dir

src/loaders/workspace_uploader/workspace_uploader.py

MrCreosote (approved):

PR #651, opened 2024-02-08
RE2022-328: update uploaded.yaml file struct

Tianhao-Gu:
An overwrite for the 'file_name' will happen when uploading multiple files using the previous structure.

### DIFF
--- a/src/loaders/workspace_uploader/workspace_uploader.py
+++ b/src/loaders/workspace_uploader/workspace_uploader.py
@@ -274,8 +274,8 @@ def _read_upload_status_yaml_file(
             <load_id>:
                 assembly_upa: <assembly_upa>
                 assembly_filename: <assembly_filename>
-                genome_upa: <genome_upa>
-                genome_filename: <genome_filename>
+                genome_upa: <genome_upa> (can be None)
+                genome_filename: <genome_filename> (can be None)
     """
 
     if upload_env_key not in loader_common_names.KB_ENV:
--- a/src/loaders/workspace_uploader/workspace_uploader.py
+++ b/src/loaders/workspace_uploader/workspace_uploader.py
@@ -264,6 +264,7 @@ def _read_upload_status_yaml_file(
     workspace_id: int,
     load_id: str,
     obj_dir: str,
+    create_assembly_only: bool = True,
 ) -> tuple[dict[str, dict[int, list[str]]], bool]:
     """
     Get metadata and upload status of a WS object from the uploaded.yaml file.
@@ -290,8 +291,11 @@ def _read_upload_status_yaml_file(
 
     workspace_dict = data.setdefault(upload_env_key, {}).setdefault(workspace_id, {})
 
-    # TODO - we might want to check that upa and filename matches
     uploaded = load_id in workspace_dict
+
+    # In the event of genome upload and an assembly with the identical load_id has already been uploaded.
+    if not create_assembly_only:
+        uploaded = uploaded and workspace_dict[load_id].get(_KEY_GENOME_UPA)
 
     return data, uploaded
 
--- a/src/loaders/workspace_uploader/workspace_uploader.py
+++ b/src/loaders/workspace_uploader/workspace_uploader.py
@@ -93,6 +93,10 @@ _WSObjTuple = namedtuple(
     ["obj_name", "host_file_dir", "container_internal_file_dir"],
 )
 
+# keys for the uploaded.yaml file
+_KEY_LOADS = "loads"
+_KEY_UPA = "upa"
+
 
 def _get_parser():
     parser = argparse.ArgumentParser(
@@ -262,9 +266,16 @@ def _read_upload_status_yaml_file(
 ) -> tuple[dict[str, dict[int, list[str]]], bool]:
     """
     Get metadata and upload status of a WS object from the uploaded.yaml file.
+
+    Structure of result yaml file:
+    <env>:
+        <workspace_id>:
+            <obj_name>:
+                loads:
+                    <load_id>:
+                        upa: <upa>
     """
 
-    uploaded = False
     if upload_env_key not in loader_common_names.KB_ENV:
         raise ValueError(
             f"Currently only support these {loader_common_names.KB_ENV} envs for upload"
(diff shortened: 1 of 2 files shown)


test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -88,7 +88,7 @@ def test_read_upload_status_yaml_file(setup_and_teardown):
         "CI", 12345, "214", assembly_dir, assembly_name
     )
     expected_data = {
-        "CI": {12345: {"file_name": assembly_name, "loads": {}}}
+        "CI": {12345: {assembly_name: {workspace_uploader._KEY_LOADS: {}}}}
```
MrCreosote, line 91:
```suggestion
        "CI": {12345: {assembly_name: {"loads": {}}}}
```

We should use literals as much as possible in tests. If someone fatfingers `loads` to `loaeds`, for example, we want the tests to break as that means the code will behave unexpectedly when dealing with existing files

MrCreosote: Same thing in other places

MrCreosote (commented):
@Xiangs18 should def approve this

test/src/loaders/workspace_uploader/workspace_uploader_test.py

Xiangs18: The original is like this:
```
CI:
  69046:
    file_name: GCF_001559985.1_GG12_C01_04_genomic.fna.gz
    loads:
         12:
            upa: '69046_311_1'
         13:
            upa: "69046_400_1"
```
It seems you changed file_name as a key now and loads as its val.

MrCreosote (commented):
LGTM, but as per above @Xiangs18 should approve

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: yea. now for a source genome, we might upload genomic.gbff.gz file as Genome in addition to Assembly. One file_name per genome no longer works.

Xiangs18: If we add more meta data inside other than file_name in future, original structure seems more flexible to extend.

Xiangs18: so we need to consider 4 cases?
1. empty
2. assembly, no genome
3. no assembly, genome
4. both

Tianhao-Gu: it's won't support multiple files, unless I misunderstand the code. If we upload `xxx.genomic.fna.gz` as Assembly first and then `xxx.genomic.gbff.gz` as Genome. `file_name` will be overwritten. And `loads` associated with `file_name` are mixed from Assembly and Genome.

Xiangs18: will not overwrite because is diff name right? They get diff extensions.

Xiangs18: wait a second, are you saying both xxx.genomic.fna.gz and xxx.genomic.gbff.gz in the same folder?

Tianhao-Gu: yea they are downloaded in the same folder.

Xiangs18: Let's say we add the genome. On the same load id, we want to add the corresponding assembly in that folder. What will happen?

Tianhao-Gu: I am not sure I get what you mean. I don't think we are adding Assembly to Genome. We now supports both Genome uploads and Assembly uploads. They are completely separate. Each object type should have its own load ids.

Tianhao-Gu: with this PR, if you uploaded both objects, the yml file should looks like:
```
env:
  ws_id:
      xxx.gbff.gz:
            loads:
                1:
                    upa: 'xxx'
                2:
                   upa: 'xxx'
     xxx.fna.gz:
             loads:
                 1: 
                    upa: 'xxx'
```
But with current structure with `file_name` as a key, we cannot support both files.

Tianhao-Gu: with this PR, if you uploaded both objects, the yml file should looks like:
```
env:
  ws_id:
      xxx.gbff.gz:
            loads:
                1:
                    upa: 'xxx'
                2:
                   upa: 'xxx'
     xxx.fna.gz:
             loads:
                 1: 
                    upa: 'xxx'
```
But with current structure with `file_name` as a key, we cannot support both files.

Xiangs18: I somehow feel genome and assembly are not completely separate. If we upload genome already, maybe not need to redo the upload of assembly? Need some clarification here @MrCreosote 

If genome and assembly are completely separate you current design would work.

Tianhao-Gu: @Xiangs18 @MrCreosote updated file structure per our discussion.

MrCreosote (commented):
LGTM but as usual @Xiangs18 should reivew

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -299,26 +304,34 @@ def _update_upload_status_yaml_file(
     upload_env_key: str,
     workspace_id: int,
     load_id: str,
-    upa: str,
-    obj_tuple: _WSObjTuple,
+    assembly_upa: str,
+    assembly_tuple: _WSObjTuple,
```
Xiangs18, line 308: assembly_upa and assembly_tuple here are positional argument. So I assume if we upload genome, you will build assembly_tuple based on genbanks_to_genomes return?

Tianhao-Gu: yea. I think we said genbanks_to_genomes will return assembly_ref as well. So we will always build assembly_tuple.

Xiangs18 (approved):
LGTM!

PR #652, opened 2024-02-12
RE2022-328: add create_assembly_only flag

Tianhao-Gu:


src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -620,7 +634,8 @@ def main():
     source_version = getattr(args, loader_common_names.SOURCE_VER_ARG_NAME)
     root_dir = getattr(args, loader_common_names.ROOT_DIR_ARG_NAME)
     token_filepath = args.token_filepath
-    upload_file_ext = args.upload_file_ext
+    create_assembly_only = args.create_assembly_only
+    upload_file_ext = args.upload_file_ext or _get_default_file_ext(create_assembly_only)
```
Xiangs18, line 638: Also need to check if user input file ext are ones we are currently supporting?

Xiangs18:
usage: workspace_uploader.py [-h] also needs to be updated.

Tianhao-Gu:
> usage: workspace_uploader.py [-h] also needs to be updated.

👍

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: can you clarify more? File extension can be anything. For example, users can just name their genbank file as `.genbank` or `.abc`.

Xiangs18: Aren't we only support "genomic.fna.gz" and "genomic.gbff.gz" for now?

Tianhao-Gu: those are the extension from downloaded NCBI files. That's our current only use case. You are suggesting we should constraint this script to only accept those 2 extensions? In that case, we don't even need a user defined `upload_file_ext`. We can just use `create_assembly_only` to choose one or the other.

Xiangs18: I personally think it is good to keep upload_file_ext but also let the user know what use cases we are currently supporting.

Tianhao-Gu: I am not sure I am following you. Can you give me an example of what upload_file_ext values we should throw an error?

Xiangs18: I think it should throw an error if upload_file_ext does not belong to a list of extensions
(["genomic.fna.gz", "genomic.gbff.gz"]) we are currently supporting.

But I am open to other thoughts.

Tianhao-Gu: hmmm `upload_file_ext ` provides users the flexibility to use their own downloaded file extension. I don't think `genomic.fna.gz` and `genomic.gbff.gz` are the only extension we are supporting. It's just one of our current use cases. I can image we will have `.fa` or `.fna` files. The uploader script should still work for those extensions.

Xiangs18: 👍

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -34,8 +34,11 @@
                         If not provided, the token must be provided in the `KB_AUTH_TOKEN` environment variable.
   --env {CI,NEXT,APPDEV,PROD}
                         KBase environment (default: PROD)
+  --create_assembly_only
+                        Create only assembly object. If not set create Genome object using genbank files by default.
   --upload_file_ext UPLOAD_FILE_EXT [UPLOAD_FILE_EXT ...]
-                        Upload only files that match given extensions (default: ['genomic.fna.gz'])
+                        Upload only files that match given extensions. If not provided, the default extension determined by the
+                        create_assembly_only flag
```
MrCreosote, line 41:
```suggestion
                        Upload only files that match given extensions. If not provided, uses the appropriate default extension
                        depending on the create_assembly_only flag
```
minor wordsmithing

MrCreosote (approved):

MrCreosote (approved):

PR #654, opened 2024-02-13
RE2022-328: refactor post process to handle genome uploads

Tianhao-Gu:


src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,75 @@ def _post_process(
...
+    # Create a standard entry in sourcedata/workspace
+    # hardlink to the original object file in sourcedata
+    src_file = Path(_get_source_file(assembly_tuple.host_file_dir, assembly_tuple.obj_name))
+    target_dir = os.path.join(output_dir, assembly_upa)
+    os.makedirs(target_dir, exist_ok=True)
+
+    suffixes = src_file.suffixes
+    # TODO - handle extension other than .gz
+    dest_suffix = suffixes[-1] if suffixes[-1] != ".gz" else "".join(suffixes[-2:])
+    dest_file = os.path.join(target_dir, f"{assembly_upa}{dest_suffix}")
```
MrCreosote, line 530: We're assuming no archive files here right? So no *.tar.gz?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,75 @@ def _post_process(
...
+    loader_helper.create_hardlink_between_files(dest_file, src_file)
+
+    # create metadata file used by parser
+    meta = loader_helper.generate_import_dir_meta(assembly_obj_info, genome_obj_info)
+    metafile = os.path.join(target_dir, f"{assembly_upa}.meta")
+    loader_helper.dump_json_to_file(metafile, meta)
+
     # Creates a softlink from new_dir to the contents of upa_dir.
```
MrCreosote, line 538: This comment seems like it's way out of date, neither of the args match

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,75 @@ def _post_process(
...
+    suffixes = src_file.suffixes
+    # TODO - handle extension other than .gz
+    dest_suffix = suffixes[-1] if suffixes[-1] != ".gz" else "".join(suffixes[-2:])
+    dest_file = os.path.join(target_dir, f"{assembly_upa}{dest_suffix}")
+    loader_helper.create_hardlink_between_files(dest_file, src_file)
+
+    # create metadata file used by parser
+    meta = loader_helper.generate_import_dir_meta(assembly_obj_info, genome_obj_info)
+    metafile = os.path.join(target_dir, f"{assembly_upa}.meta")
+    loader_helper.dump_json_to_file(metafile, meta)
```
MrCreosote, line 536: Isn't this already created somewhere else?

MrCreosote (commented):
haven't looked at the tests yet. @Xiangs18 should also take a look at this one

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,75 @@ def _post_process(
...
-    # hardlink to the original object file in sourcedata
-    src_file = _get_source_file(
-        obj_tuple.host_file_dir,
-        obj_tuple.obj_name,
-    )
-    target_dir = os.path.join(output_dir, upa)
-    os.makedirs(target_dir, exist_ok=True)
-    dest_file = os.path.join(target_dir, f"{upa}.fna.gz")
-    loader_helper.create_hardlink_between_files(dest_file, src_file)
-
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
Tianhao-Gu: as discussed, we are moving towards supporting Genome uploads only. So I am guessing this is fine for now.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: no. only `.gz` file for now.

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: no. As per my understanding, appears that only the fasta file is linked. But no `.meta` file.

MrCreosote: It must be - the pipeline depends on the meta file to get the genome upa at least

src/loaders/workspace_uploader/workspace_uploader.py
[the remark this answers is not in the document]
MrCreosote: Seems like it should error out if any of the genome or assembly arguments aren't provided

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: hmm no. We are at the upload step. We haven't touched WS yet.

src/loaders/workspace_uploader/workspace_uploader.py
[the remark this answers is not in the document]
Tianhao-Gu: maybe once everything is refactored. If I make both arguments positional now, everything will break.

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: But the uploader is supposed to generate the same input directory contents as the downloader, since it has all the info the downloader would once the object are uploaed

src/loaders/workspace_uploader/workspace_uploader.py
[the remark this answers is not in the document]
MrCreosote: Ok, TODO then?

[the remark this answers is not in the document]

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: yea. That's why I added it here. It must be overlooked before.

MrCreosote: So is there a downloader fn somewhere we're duplicating?

Tianhao-Gu: oh I see what you meant now. 

The only difference is the downloader inlined arguments of dump_json_to_file.  
https://github.com/kbase/collections/blob/main/src/loaders/workspace_downloader/workspace_downloader.py#L153

MrCreosote: Ok we shouldn't duplicate the fn, so whichever way you choose re the dump call is fine with me

src/loaders/common/loader_helper.py
```diff
@@ -586,3 +586,26 @@ def dump_json_to_file(json_file_path: str | Path, json_data: dict[str, Any] | li
     """
     with open(json_file_path, "w", encoding="utf8") as json_file:
         json.dump(json_data, json_file, indent=2)
+
+
+def create_meta_file(
+        output_dir: Path | str,
+        upa_format: str,
```
Xiangs18, line 593: Maybe just `upa`?

Xiangs18 (approved):
LGTM!

src/loaders/common/loader_helper.py
Tianhao-Gu: For many KBasers, the initial impression of a UPA is in the format of `a/b/c` as a workspace object reference.  However, for collection, UPAs are typically structured as 'a_b_c' to align with the file structure. That's why I named `upa` in formate of `a_b_c` differently.

Tianhao-Gu: I updated it to `structured_upa`. Maybe that's a less confusing name.

Xiangs18 (dismissed):
LGTM!

src/loaders/workspace_uploader/workspace_uploader.py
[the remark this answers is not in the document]

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,75 @@ def _post_process(
...
         load_id,
-        upa,
-        obj_tuple,
+        assembly_upa,
+        assembly_tuple,
+        genome_upa=genome_upa,
+        genome_tuple=genome_tuple,
     )
 
+    if genome_tuple and genome_upa:
```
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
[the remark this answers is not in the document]
MrCreosote: :

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_downloader/workspace_downloader.py
```diff
@@ -150,7 +150,7 @@ def _process_input(conf: Conf):
             os.link(cfn, dst)
 
             # save meta file with relevant object_info
-            loader_helper.dump_json_to_file(metafile, loader_helper.generate_import_dir_meta(obj_info, genome_info))
+            loader_helper.create_meta_file(conf.output_dir, upa_format, obj_info, genome_info)
```
MrCreosote, line 153: This should return the meta file location since it's used below, rather than having it declared above

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,81 @@ def _post_process(
...
-        obj_tuple.host_file_dir,
-        obj_tuple.obj_name,
-    )
-    target_dir = os.path.join(output_dir, upa)
-    os.makedirs(target_dir, exist_ok=True)
-    dest_file = os.path.join(target_dir, f"{upa}.fna.gz")
-    loader_helper.create_hardlink_between_files(dest_file, src_file)
+    # TODO: make all parameters positional arguments
+
+    if (genome_tuple is None) != (genome_upa is None):
```
MrCreosote, line 490:
```suggestion
    if bool(genome_tuple) != bool(genome_upa):  # xor
```

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,81 @@ def _post_process(
...
+        if not assembly_obj_info or not genome_obj_info:
+            raise ValueError(
+                "Both assembly_obj_info and genome_obj_info must be provided"
+            )
+
+        _process_genome_objects(
+            upload_dir, output_dir, assembly_tuple, assembly_upa, assembly_obj_info, genome_obj_info)
+
+    # Update the 'uploaded.yaml' file, serving as a marker to indicate the successful upload of the object.
+    # Ensure that this operation is the final step in the post-processing workflow
```
MrCreosote, line 506: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,81 @@ def _post_process(
     upload_env_key: str,
     workspace_id: int,
     load_id: str,
-    obj_tuple: _WSObjTuple,
     upload_dir: str,
```
MrCreosote, line 471: upload_dir is kind of a confusing name, maybe collections_source_dir?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -471,39 +468,81 @@ def _post_process(
...
     output_dir: str,
```
MrCreosote, line 472: Same for output dir, maybe source_dir? Although that's a bit too close to collections_source_dir

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -203,7 +212,7 @@ def test_post_process(setup_and_teardown):
 
     host_assembly_dir = params.assembly_dirs[0]
     assembly_name = ASSEMBLY_NAMES[0]
-    src_file = params.target_files[0]
```
MrCreosote, line 206: is this cruft? Seems like it's not used

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -227,18 +236,67 @@ def test_post_process(setup_and_teardown):
                                "genome_upa": None,
                                "genome_filename": None}}}}
 
-    dest_file = os.path.join(
-        os.path.join(output_dir, "12345_58_1"), f"12345_58_1.fna.gz"
+    assert uploaded
+    assert expected_data == data
+
+    # test with genome_tuple and genome_upa
```
MrCreosote, line 242: This seems like it should be a separate test

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -356,25 +414,6 @@ def test_upload_assembly_files_in_parallel(setup_and_teardown):
...
-    assert os.path.samefile(
-        src_files[0],
-        os.path.join(output_dir, "12345_58_1", "12345_58_1.fna.gz")
-    )
-
-    assert os.path.samefile(
-        src_files[1],
-        os.path.join(output_dir, "12345_60_1", "12345_60_1.fna.gz")
-    )
-
```
MrCreosote, line 377: This is deleted because if only assemblies are uploaded nothing happens in the WS source dir now? Presumably when we upload genomes there will be a new test along these lines?

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: This step is in `if` condition and can be skipped.

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: it's not used since this test only test assembly. But used in genome cases in the separate test.

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: 👍

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: yup. that's my plan.

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: I do think it makes sense to return the metadata file for the helper function. But we cannot use it here.

MrCreosote: Maybe there should be a function for just getting the file location then?

What happens if the file isn't created and the function keeps going?

src/loaders/workspace_uploader/workspace_uploader.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

src/loaders/workspace_downloader/workspace_downloader.py
Tianhao-Gu: > Maybe there should be a function for just getting the file location then?

👍 

> What happens if the file isn't created and the function keeps going?

It won't. The file existence and content are checked by the if condition.

MrCreosote: :+1:

src/loaders/common/loader_helper.py
```diff
@@ -586,3 +586,43 @@ def dump_json_to_file(json_file_path: str | Path, json_data: dict[str, Any] | li
...
+        structured_upa: str
+) -> Path:
+    """
+    Get the metadata file path for a specific workspace id under sourcedata/WS/<env>.
+
+    source_dir - The directory for a specific workspace id under sourcedata/WS/<env>
+    structured_upa - The UPA of a workspace object in the format of "<wsid>_<objid>_<ver>"
+    """
+    source_dir = Path(source_dir)
+    metafile = source_dir / structured_upa / f"{structured_upa}.meta"
```
Xiangs18, line 602: This seems only work if `source_dir` is type Path?

Tianhao-Gu: I cast it to Path in case if it's not.

Tianhao-Gu: I inline the casting. I think that makes more sense.

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -227,18 +236,74 @@ def test_post_process(setup_and_teardown):
...
+    collections_source_dir = Path(params.tmp_dir) / "collections_source_dir"
+    collections_source_dir.mkdir()
+    source_dir = Path(params.tmp_dir) / "source_dir"
+    source_dir.mkdir()
+
+    host_assembly_dir2 = params.assembly_dirs[1]
+    assembly_name2 = ASSEMBLY_NAMES[1]
+    src_file2 = params.target_files[1]
+    assembly_tuple2 = workspace_uploader._WSObjTuple(
+        assembly_name2, host_assembly_dir2, "/path/to/file/in/AssembilyUtil"
```
MrCreosote, line 254: Why are these all *2?

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -227,18 +236,74 @@ def test_post_process(setup_and_teardown):
...
+
+    # check metadata file
+    metadata_file = source_dir / "42_4_75" / "42_4_75.meta"
+    with open(metadata_file, 'r') as file:
+        data = json.load(file)
+
+    assert data['upa'] == '42/4/75'
+    assert data['genome_upa'] == '42/7/1'
+    assert data['assembly_object_info'] == assembly_obj_info
+    assert data['genome_object_info'] == genome_obj_info
```
MrCreosote, line 306: Why not compare the entire dicts like above?

MrCreosote, line 306: Should there be a softlink check?

src/loaders/common/loader_helper.py
Xiangs18: 👍

src/loaders/common/loader_helper.py
```diff
@@ -335,7 +335,7 @@ def is_upa_info_complete(upa_dir: str):
     """
     upa = os.path.basename(upa_dir)
     fa_path = os.path.join(upa_dir, upa + ".fa")
-    meta_path = os.path.join(upa_dir, upa + ".meta")
+    meta_path = get_meta_file_path(os.path.dirname(upa_dir), upa)
```
Xiangs18, line 338: Is this necessary?

Tianhao-Gu: you are referring to `os.path.dirname(upa_dir)`? Yes. the function gets source_dir which is the parent of upa_dir.

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: fixed.

test/src/loaders/workspace_uploader/workspace_uploader_test.py

Tianhao-Gu: it's already checked above right?
```
    assert os.readlink(os.path.join(collections_source_dir, "42_4_75")) == os.path.join(
        source_dir, "42_4_75"
    )
```

MrCreosote: Whoops, :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

MrCreosote (approved):

src/loaders/common/loader_helper.py
Xiangs18: 👍

PR #655, opened 2024-02-15
RE2022-328: refactor ws retrieval obj to also return obj info

Tianhao-Gu:


src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -410,24 +410,50 @@ def _query_workspace_with_load_id(
     workspace_id: int,
     load_id: str,
     obj_names: list[str],
-) -> tuple[list[str], list[str]]:
+    assembly_objs_only: bool = True,
+) -> tuple[list[str], list[str], list[str], list[str]]:
```
Xiangs18, line 414: assembly_objs_info and genome_objs_info not necessarily return a list of string?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -410,24 +410,50 @@ def _query_workspace_with_load_id(
...
+        _check_obj_type(workspace_id, load_id, uploaded_objs_info, {loader_common_names.OBJECTS_NAME_GENOME})
+        genome_objs_info = uploaded_objs_info
+        try:
+            assembly_objs_ref = [info[10]['Assembly Object'] for info in genome_objs_info]
+        except KeyError:
+            raise ValueError(
+                f"Genome objects must have an 'Assembly Object' field in their metadata"
+            )
+        assembly_objs_spec = [{"ref": ref} for ref in assembly_objs_ref]
+        assembly_objs_info = ws.get_object_info3({"objects": assembly_objs_spec, "includeMetadata": 1})["infos"]
```
MrCreosote, line 441: We're assuming here that the assemblies are in an accessible workspace and we don't need to provide a reference path, which I guess is ok?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -410,24 +410,50 @@ def _query_workspace_with_load_id(
...
+
+    return uploaded_obj_names_batch, uploaded_obj_upas_batch, assembly_objs_info, genome_objs_info
```
MrCreosote, line 443: Returning 4 things seems excessive, but since it's an internal method only...

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -587,3 +590,89 @@ def test_query_workspace_with_load_id_mass(setup_and_teardown):
...
+    assert assembly_objs_info == assembly_objs_response['infos']
+    assert genome_objs_info == genome_objs_response['infos']
+
+    # Assert expected calls to ws.get_object_info3
+    expected_calls = [
+        call({"objects": [{"wsid": 69046, "name": "genome_1"},
+                          {"wsid": 69046, "name": "genome_2"}], "ignoreErrors": 1, "includeMetadata": 1}),
+        call({"objects": [{"ref": "1/1/1"}, {"ref": "2/2"}], "includeMetadata": 1}),
+    ]
+    ws.get_object_info3.assert_has_calls(expected_calls, any_order=False)
```
Xiangs18, line 635: Does this check how many times ws.get_object_info3 get called?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -753,9 +789,15 @@ def main():
...
-        )
+        (uploaded_obj_names,
+         uploaded_obj_upas,
+         uploaded_assembly_objs_info,
+         uploaded_genome_objs_info) = _query_workspace_with_load_id_mass(ws,
+                                                                         workspace_id,
+                                                                         load_id,
+                                                                         list(wait_to_upload_objs.keys()),
+                                                                         assembly_objs_only=create_assembly_only)
+
```
MrCreosote, line 800: At this point the main function is 175 lines long, which is way too long. In a PR pretty soon we need to split this up into cohesive functions.

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -436,18 +462,26 @@ def _query_workspace_with_load_id_mass(
...
-            ws, workspace_id, load_id, obj_names[idx: idx + batch_size]
-        )
-        uploaded_obj_names.extend(obj_names_batch)
-        uploaded_obj_upas.extend(obj_upas_batch)
-
-    return uploaded_obj_names, uploaded_obj_upas
+    assembly_objs_only: bool = True,
+) -> tuple[list[str], list[str], list[str], list[str]]:
+
+    # TODO: run _query_workspace_with_load_id in parallel instead of in serial
```
MrCreosote, line 468: Just checking - it's serial right now, right? We're not downgrading here?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -436,18 +462,26 @@ def _query_workspace_with_load_id_mass(
...
+                                                                      obj_names[i:i + batch_size],
+                                                                      assembly_objs_only=assembly_objs_only)
+                                        for i in range(0, len(obj_names), batch_size)])
+
+    return (
+        [item for sublist in uploaded_obj_names for item in sublist],
+        [item for sublist in uploaded_obj_upas for item in sublist],
+        [item for sublist in uploaded_assembly_objs_info for item in sublist],
+        [item for sublist in uploaded_genome_objs_info for item in sublist]
+    )
```
MrCreosote, line 484: Here I think the loop is a lot more readable

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -587,3 +590,89 @@ def test_query_workspace_with_load_id_mass(setup_and_teardown):
...
+            [2, 'assembly_2', 'KBaseGenomeAnnotations.Assembly-10', 'time', 75, 'user', 42, 'wsname', 'md5', 78, {}]
+        ]
+    }
+
+    genome_objs_response = {
+        "infos": [
+            [1, 'genome_1', 'KBaseGenomes.Genome-6', 'time', 75, 'user', 42, 'wsname', 'md5', 78,
+             {"load_id": load_id, "Assembly Object": "1/1/1"}],
+            [4, 'genome_2', 'KBaseGenomes.Genome-9.3', 'time', 75, 'user', 42, 'wsname', 'md5', 78,
+             {"load_id": load_id, "Assembly Object": "2/2"}]
```
MrCreosote, line 611: The assembly object ref is always pulled from the object and is annotated as a WS id so it will always be fully qualified

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -587,3 +590,89 @@ def test_query_workspace_with_load_id_mass(setup_and_teardown):
...
+
+    # Assert expected calls to ws.get_object_info3
+    expected_calls = [
+        call({"objects": [{"wsid": 69046, "name": "genome_1"},
+                          {"wsid": 69046, "name": "genome_2"}], "ignoreErrors": 1, "includeMetadata": 1}),
+        call({"objects": [{"ref": "1/1/1"}, {"ref": "2/2"}], "includeMetadata": 1}),
+    ]
+    ws.get_object_info3.assert_has_calls(expected_calls, any_order=False)
+
+    # Test case 2: Invalid scenario - genome object does not have an 'Assembly Object' field in its metadata
```
MrCreosote, line 637: I would split these into separate tests with a shared setup method if there's lots of setup. Same for the test below

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -587,3 +590,89 @@ def test_query_workspace_with_load_id_mass(setup_and_teardown):
...
+    ws.get_object_info3.side_effect = [genome_objs_response]
+    with pytest.raises(ValueError) as excinfo:
+        workspace_uploader._query_workspace_with_load_id_mass(
+            ws,
+            69046,
+            load_id,
+            ["genome_1"],
+            assembly_objs_only=False
+        )
+    assert "Genome objects must have an 'Assembly Object' field in their metadata" == str(excinfo.value)
```
MrCreosote, line 654: It'd be helpful if it gave the ref for the offending object

test/src/loaders/workspace_uploader/workspace_uploader_test.py
```diff
@@ -587,3 +590,89 @@ def test_query_workspace_with_load_id_mass(setup_and_teardown):
...
+    ]
+    expected_obj_types = {'TypeA', 'TypeB'}
+
+    workspace_uploader._check_obj_type(workspace_id, load_id, obj_infos, expected_obj_types)
+
+    # Test case 2: Invalid scenario
+    obj_infos = [
+        [1, 'abc-123', 'TypeA', 'time', 75, 'user', 42, 'wsname', 'md5', 78, {'foo': 'bar'}],
+        [2, 'def-456', 'TypeC', 'time', 75, 'user', 42, 'wsname', 'md5', 78, {}]  # TypeC is not expected
+    ]
```
MrCreosote, line 673: Less code to just change expected_obj_types, unless there's a reason not to do that

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: It doesn't, it just has to have the given calls. To ensure it only has those calls you can do 

`mock.call_count = 2`

src/loaders/workspace_uploader/workspace_uploader.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: 👍

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: yea. I think that would be okay for now.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: yea. next PR will fix this.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: coming up next.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: oh. I mean the `for` loop. Parallel within `_query_workspace_with_load_id`.

src/loaders/workspace_uploader/workspace_uploader.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: hmmm it was caught in `assembly_objs_ref = [info[10]['Assembly Object'] for info in genome_objs_info]` making it hard to find just return the bad genome ref. Unless we run it in a for loop.

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: hmmm can you clarity more?

MrCreosote: It'll never be `2/2` since the workspace rewrites any `@id ws` annotated fields to full UPAs, and the `assembly_ref` field, which is where the `Assembly Object` meta value comes from, is an `@id ws` field

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Sorry, not grokking the response here - is there a part missing?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: yeah, we should probably do that, clear error messages > code brevity

Although maybe you could make a case here that it's not needed if you can ensure that all Genomes should always have that field. I think newer ones do but not sure about older versions of the type

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: hmm I got confused with your question now too.

MrCreosote: There's a TODO to parallelize something. I'm wondering if that's because it just needs to be parallelized or because this change broke the current parallelization

test/src/loaders/workspace_uploader/workspace_uploader_test.py
Tianhao-Gu: 👍 fixed.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: oh I see. Parallelization refers to `for i in range(0, len(obj_names), batch_size)` to bump performance. We don't need to do anything if performance is good.

test/src/loaders/workspace_uploader/workspace_uploader_test.py

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Ok, so this is just a note about a potential future improvement

Tianhao-Gu: yea. removed the todo. Probably don't need to worry about performance for now.

src/loaders/workspace_uploader/workspace_uploader.py

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: :+1:

test/src/loaders/workspace_uploader/workspace_uploader_test.py
MrCreosote: Can you do the same thing for the test below?

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -410,24 +410,54 @@ def _query_workspace_with_load_id(
...
-    uploaded_obj_upas_batch = [
-        path[0].replace("/", "_") for path in res["paths"] if path is not None
-    ]
-    return uploaded_obj_names_batch, uploaded_obj_upas_batch
+    res = ws.get_object_info3({"objects": refs, "ignoreErrors": 1, "includeMetadata": 1})
+    uploaded_objs_info = [info for info in res["infos"] if info and info[10].get("load_id") == load_id]
+    if not uploaded_objs_info:
+        return list(), list(), list(), list()
+    uploaded_obj_names_batch = [info[1] for info in uploaded_objs_info]
+    uploaded_obj_upas_batch = ["{6}_{0}_{4}".format(*info) for info in uploaded_objs_info]
```
MrCreosote, line 425: There's a couple more info -> upa implementations below and I'm sure there's more in other places around the the code base. We should just make a function in src/common like

```
def obj_info_to_upa(upa, underscore_sep=False) -> str:
    sep = "_" if underscore_sep else "/"
    return f"{info[6]}{sep}{info[0]}{sep}{info[4]}
```

...and start replacing the zillion independent implementations

test/src/loaders/workspace_uploader/workspace_uploader_test.py

MrCreosote: :+1:

MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: 👍 updated a few places I found in the uploader and downloader. Will replace others as I encounter them.

MrCreosote (commented):
LGTM but @Xiangs18 should approve

src/loaders/common/loader_helper.py
```diff
@@ -626,3 +626,14 @@ def create_meta_file(
 
     return metafile
 
+
+def obj_info_to_upa(obj_info: list[Any], underscore_sep=False) -> str:
```
MrCreosote, line 630: Actually can we put this in src/common vs src/loaders? I suspect we use it in the service as well

Tianhao-Gu: 👍

MrCreosote: :+1:

Xiangs18 (approved):
LGTM!

PR #657, opened 2024-02-23
RE2022-330: Fix missing sample table entries bug

MrCreosote:
Not all sample entries in the DB have the same keys, which can mean the table view of the samples table may have out of order columns for samples with missing keys. Fill in missing values with `null`.

### DIFF
--- a/src/service/data_products/common_functions.py
+++ b/src/service/data_products/common_functions.py
@@ -279,20 +279,22 @@ async def query_simple_collection_list(
 
 
 def _query_acceptor(
+    fields: list[str],
     data: list[dict[str, Any]],
-    last: list[dict[str, Any]],
     doc: dict[str, Any],
     output_table: bool,
     document_mutator: Callable[[dict[str, Any]], dict[str, Any]],
     count: bool,
 ):
-    last[0] = doc
     if count:
         data.append(doc)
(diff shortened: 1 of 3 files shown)


src/service/data_products/common_functions.py
```diff
@@ -337,36 +340,31 @@ async def query_table(
...
 
     storage - the storage system.
+    columns - the expected columns in the table data.
     filters - the filters to apply to the search
     output_table - whether to return the results as a list of lists (e.g. a table) with a separate
         fields entry defining the key for each table column, or a list of key / value dictionaries.
     document_mutator - a function applied to a document retrieved from the database before
         returning the results.
     """
+    fields = [c.key for c in columns]
```
Tianhao-Gu, line 350: should we enforce `sort` here rather than relying on caller to pass a `sorted(cols)`?

MrCreosote: I'm assuming that in the future we'll probably want to define a sort order in the column specs, in which case the calling code will need to provide the sort order to `query_table`

Tianhao-Gu: 👍

Tianhao-Gu (approved):
💯

PR #658, opened 2024-02-28
Fix match / selections bug

MrCreosote:
Previous change to specified fields left out the special match and selection fields

### DIFF
--- a/src/service/data_products/common_functions.py
+++ b/src/service/data_products/common_functions.py
@@ -13,6 +13,7 @@ from src.common.storage.db_doc_conversions import (
 )
 from src.service import errors, kb_auth, models, app_state
 from src.service.filtering.filters import FilterSet
+from src.service.processing import SubsetSpecification
 from src.service.storage_arango import ArangoStorage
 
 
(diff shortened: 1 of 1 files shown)


src/service/data_products/common_functions.py
```diff
@@ -268,16 +269,20 @@ async def query_simple_collection_list(
...
-                        names.FLD_MATCHES_SELECTIONS]
+                    d[selection_field] = _get_matchsel(filters.selection_spec, d)
             acceptor(d)
     finally:
         await cur.close(ignore_missing=True)
 
 
+def _get_matchsel(spec: SubsetSpecification, doc: dict[str, Any]):
+    if not doc.get(names.FLD_MATCHES_SELECTIONS):
+        return False
```
Tianhao-Gu, line 282: not `list()` but `False` here?

src/service/data_products/common_functions.py
```diff
@@ -352,6 +357,10 @@ async def query_table(
         raise errors.IllegalParameterError(
                 f"No such field for collection {filters.collection_id} load version "
                 + f"{filters.load_ver}: {filters.sort_on}")
+    if not filters.selection_spec.is_null_subset():
+        fields = [names.FLD_SELECTED_SAFE] + fields
```
Tianhao-Gu, line 361: it's ok to have duplicate `names.FLD_SELECTED_SAFE`/ `names.FLD_MATCHED_SAFE` in `fields` array?

MrCreosote: I'm not understanding, those aren't duplicate names

src/service/data_products/common_functions.py
MrCreosote: No, `__match__` and `__sel__` field values are expected to be `True` or `False`

src/service/data_products/common_functions.py
Tianhao-Gu: can `fields` array already have `names.FLD_SELECTED_SAFE`? unlikely?

src/service/data_products/common_functions.py
MrCreosote: I added the missing return type to make this clearer

src/service/data_products/common_functions.py
MrCreosote: They don't actually exist in the database, so unless the calling code does something really weird, no.

MrCreosote: Currently the field names are coming from the column specs, which don't have those names.

src/service/data_products/common_functions.py
Tianhao-Gu: 👍

src/service/data_products/common_functions.py
Tianhao-Gu: 👍

Tianhao-Gu (approved):
👍

PR #678, opened 2024-03-12
RE2022-328: retrive fasta file from assemblyutil

Tianhao-Gu:


src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -639,11 +639,69 @@ def _process_batch_upload(
...
+    assembly_tuples = list()
+
+    for genome_tuple, genome_obj_info in zip(genome_tuples, genome_obj_infos):
+        assembly_file = genome_tuple.obj_name + "_assembly.fasta"  # matches the FASTA file name returned from GFU during successful upload process
+        coll_src_assembly_path = os.path.join(genome_tuple.obj_coll_src_dir, assembly_file)
+        if not os.path.exists(coll_src_assembly_path):
+            _download_assembly_fasta_file(asu_client,
+                                          genome_obj_info,
+                                          coll_src_assembly_path,
+                                          job_data_dir)
```
MrCreosote, line 688: You could batch these up to speed things up but probably not worth the trouble

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -639,11 +639,69 @@ def _process_batch_upload(
...
+
+        assembly_tuple = WSObjTuple(assembly_file,
+                                    genome_tuple.obj_coll_src_dir,
+                                    None)  # container_internal_file is fine to be None here as it's not used during failure recovery
```
MrCreosote, line 692: This goes into _post_process just like normal uploads, that's not a problem?

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -639,11 +639,69 @@ def _process_batch_upload(
...
+    # remove prefix of the assembly_path
+    relative_assembly_path = containerized_assembly_file.relative_to(_JOB_DIR_IN_ASSEMBLYUTIL_CONTAINER)
+    local_assembly_file = Path(job_data_dir) / relative_assembly_path
+
+    loader_helper.create_hardlink_between_files(coll_src_assembly_path,
+                                                local_assembly_file)
+
+    if not os.path.exists(coll_src_assembly_path):
+        raise ValueError(f"Retrieving assembly file from Genome {genome_obj_info[1]} failed.")
+
```
MrCreosote, line 666: Can this actually happen? I think the hardlink would've failed before this error

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -745,6 +806,8 @@ def _upload_objects_in_parallel(
                                                          workspace_id,
                                                          load_id,
                                                          obj_tuples,
+                                                         asu_client,
```
MrCreosote, line 809: What happens if an assembly is uploaded but the genome upload doesn't complete?

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: It won't be a problem. `container_internal_file` is used to build WS object. We won't build Assembly object. The assembly object associate with genome is build from the genbank file.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: yea. We can leave it as is for now. I think (hope) there won't be too many failed cases for recovery.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: hmmm. the last call in hardlink is `os.link(target_file, new_file)`. In theory, this shouldn't happen. But just in case? or do you think removing this check makes more sense?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: it first does `_query_workspace_with_load_id_mass` to check successfully uploaded Genomes. If Genome hasn't created, it will try to create Genome in the second run.

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: I'm not sure I understand - container_internal_file is the location for the file inside the docker container. That's not used to build the ws object

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: What happens if target_file doesn't exist in the link?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: What if there is an assembly though?

Tianhao-Gu: `_query_workspace_with_load_id_mass` won't find previous uploaded Assemblies since the name will be different. Genome obj name is exactly the same as genbank file name.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: if target_file doesn't exist, `os.path.samefile(target_file, new_file)` will throw an error in `create_hardlink_between_files`.

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Ok, so if an assembly gets uploaded but it's genome doesn't, the assembly will be duplicated next load?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: So it seems like that raise line is impossible to reach

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: 👍 removed the condition.

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: yup. The assembly will be duplicated during the next load.

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Ok, I guess that's ok for now.

MrCreosote (commented):
LGTM but @Xiangs18 should take a look

Xiangs18:
Will we add more tests in future PR?

Xiangs18 (approved):
LGTM!
Ideally we should add more tests in future PR.

PR #680, opened 2024-03-13
RE2022-328: fix compatibility with GFU

Tianhao-Gu:


src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -1025,8 +1025,8 @@ def main():
         if proc:
             proc.terminate()
 
-    if not keep_job_dir:
-        shutil.rmtree(job_dir)
+        if not keep_job_dir:
+            shutil.rmtree(job_dir)
```
Tianhao-Gu, line 1029: when above in try returns, `job_dir` might not be deleted if it's outside of finally.

MrCreosote (commented):
LGTM but @Xiangs18 should review

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: no. in try, function can return. 

```
In [4]: def example_function():
   ...:     try:
   ...:         return
   ...:     finally:
   ...:         print('foo')
   ...: 
   ...:     print('bar')
   ...: 
   ...: example_function()
foo

In [5]:
```

Xiangs18: yeah, I made a toy example. 👍

Xiangs18 (approved):
LGTM!

PR #682, opened 2024-03-13
add eggnog

Tianhao-Gu:


src/loaders/common/loader_common_names.py
```diff
@@ -94,6 +94,9 @@
 # The metadata file name created during the Mash run
 MASH_METADATA = 'mash_run_metadata.json'
 
+# The metadata file name created during the Mash run
```
MrCreosote, line 97:
```suggestion
# The metadata file name created during the Eggnog run
```

src/loaders/compute_tools/eggnog/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN conda config --add channels bioconda
+RUN conda config --add channels conda-forge
+
+RUN conda create -n $CONDA_ENV python=$PYTHON_VER
+RUN conda install -n $CONDA_ENV -c conda-forge -c bioconda eggnog-mapper=$EGGNOG_VER
+RUN conda install -n $CONDA_ENV pandas=2.2.1 jsonlines=2.0.0
+
+RUN echo "source activate $CONDA_ENV" >> ~/.bashrc
+
+# eggNOG annotation DB is pre-downloaded at /global/cfs/cdirs/kbase/collections/libraries/eggnog/5.0.2
```
MrCreosote, line 16: 5.0.2 is the DB version?

src/loaders/compute_tools/eggnog/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# following instructions at https://github.com/eggnogdb/eggnog-mapper/wiki/eggNOG-mapper-v2.1.5-to-v2.1.12#setup
+# Mount the annotation DB directory to /reference_data when running the container
```
MrCreosote, line 18: Can you remind me how much of the DB management / mounting / etc. is manual and how much is automated

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -49,10 +49,11 @@
 256 cores.
 '''
 
-TOOLS_AVAILABLE = ['gtdb_tk', 'checkm2', 'microtrait', 'mash']
+TOOLS_AVAILABLE = ['gtdb_tk', 'checkm2', 'microtrait', 'mash', 'eggnog']
```
MrCreosote, line 52: Is this going to confuse / interact with the rest of the pipeline?

src/loaders/compute_tools/eggnog/Dockerfile
MrCreosote, line 1: Can you add some sort of readme.md explaining the cdm / collections stuff we talked about

src/loaders/compute_tools/eggnog/eggnog.py
```diff
@@ -0,0 +1,58 @@
...
+        print(f"Skipping {source_file} as it has already been processed.")
+        return
+
+    # RUN eggNOG for a single genome
+    command = ['emapper.py',
+               '-i', source_file,  # Input file.
+               '-o', output_dir / source_file.name,  # Output prefix.
+                                                     # Save result file to collectiondata directory. Expecting 'emapper.annotations', 'emapper.hits' and  'emapper.seed_orthologs' files.
+               '--itype', f'{INPUT_TYPE}',
+               '--cpu', f'{THREADS}',
```
MrCreosote, line 36: I didn't realize we were using 16 threads per instance. Any idea of how that scales? If we reduced the threads per instance and it doesn't make a difference maybe we could increase the instances / node

src/loaders/compute_tools/eggnog/eggnog.py
```diff
@@ -0,0 +1,58 @@
...
+               '--excel',
+               '--sensmode', 'fast',
+               '--dmnd_iterate', 'no',
```
MrCreosote, line 39: We don't have any mechanism to submit custom arguments to these tools do we?

src/loaders/compute_tools/eggnog/Dockerfile
Tianhao-Gu: yea. That's the eggNOG DB version. There are also diamond DB and MMseqs DB. But I use the eggNOG DB version.
```
(eggnog-2.1.12) root@919d2e0de08e:/app# emapper.py --version
emapper-2.1.12 / Expected eggNOG DB version: 5.0.2 / Installed eggNOG DB version: 5.0.2 / Diamond version found: diamond version 2.1.9 / MMseqs2 version found: 15.6f452 / Compatible novel families DB version: 1.0.1
```

MrCreosote: And the instructions you reference explain how to install all those dbs and they're all in the same dir?

src/loaders/common/loader_common_names.py

src/loaders/compute_tools/eggnog/Dockerfile
Tianhao-Gu: We need to manually download the DB to collections/libraies/<tool_name>/<db_version>.
If in the versions.yaml file a `reference_db_version` field is filled, the code will automatically mount the db to /reference_data directory.

MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: no, I don't think so. The parser has its own defined tools. Even if we run eggnog on legit collections, the parser will skip this tool.

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: yea. We didn't build that. Something should consider if we were going to rewrite this for CDM.

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: and running the tasks won't be affected?

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: I can def see us wanting to be able to change parameter sets

MrCreosote: Anyway :+1:

src/loaders/compute_tools/eggnog/Dockerfile
Tianhao-Gu: yea. The I followed the instructions to install the DB. They are all in the same dir by the download_eggnog_data.py program comes with eggnog installation.

src/loaders/compute_tools/eggnog/Dockerfile
Tianhao-Gu: 👍

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: testing different setups. threads per instance doesn't seem to be a big factor here.

src/loaders/compute_tools/eggnog/Dockerfile
MrCreosote: :+1:

src/loaders/compute_tools/eggnog/Dockerfile
MrCreosote: :+1:

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: Hm. I wonder if we reduced that you might see better performances with more instances per node.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: No. Parser will just skip the result of this tool. And plus, we are running this tool on a custom collection.

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: yea. I am planning to try that.

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: :+1:

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: Set the internal threads to 4 for now. Running 16 parallel processes concurrently will exceed 70% in memory usage. Running 32 or 64 processes intermittently disrupts the program. I am guessing it's due to memory. 

16 parallelization with 4 threads per instance will give us. 
`Used 16.2 minutes to execute eggnog for 147 data units`

I think that's pretty good.

MrCreosote: In the slack thread you said 8 threads each resulted in ~14m runs - any reason you went with 4 over 8? 128 threads doesn't seem egregious at all.

How many cores are there per instance?

Tianhao-Gu: 👍  Increased threads to 8. There are 256 cores in total per node.

MrCreosote: Hm. Suprised that 16 threads / instance slows things down then, odd.

Anyway, one other question - I could've sworn we had 2 thread related CLI options - one for total threads and one for threads per instance. is that right? If so, why doesn't eggnog use the threads per instance option?

Tianhao-Gu: yea. We have those options but those are designed for batch parallel tools. For single parallel tools, we are not passing any threads info to each instance (bad design probably because single parallelization tools are not expected when we did this originally). As a result, all we did was to run number of threads in parallel for single parallel tools. 
https://github.com/kbase/collections/blob/develop/src/loaders/compute_tools/tool_common.py#L304

MrCreosote: So for single parallel tools is one of the arguments ignored?

Tianhao-Gu: yes. total number of threads are not used.

MrCreosote: Ok. yeah, seems like we should use the total # of threads / threads per instance to set instance count and provide the threads per instance to the container

Or just set instance count and threads per instance, that seems a bit simpler, although it makes it a bit easier for the user to specify thread count >> core count

MrCreosote: Anyway :+1:

MrCreosote (approved):

PR #683, opened 2024-03-13
RE2022-335: Fix service manager failure to create views

MrCreosote:
If the service hasn't been started yet, the custom analyzers won't be installed and the views will fail to be created.


Tianhao-Gu (approved):
👍

PR #688, opened 2024-03-15
pass program threads to tools

Tianhao-Gu:


src/loaders/compute_tools/tool_common.py
```diff
@@ -298,6 +298,7 @@ def parallel_single_execution(self, tool_callable: Callable[[str, str, Path, Pat
                  meta.get(loader_common_names.META_UNCOMPRESSED_FILE,
                           meta[loader_common_names.META_SOURCE_FILE]),
                  output_dir,
+                 self._program_threads,
```
MrCreosote, line 301: Doesn't this make the thread count different for single and batch run programs? Batch programs will never use more than `_threads` threads since they divide that by `_program_threads` to get the instance count IIUC. Single program threads will now use `_threads` * `_program_threads` threads

Tianhao-Gu: I will need to unify the logic for single and batch programs. 

For batch program, I designed to compute the threads number by _threads/_program_threads. However, later we decided to have Taskfarmer handle parallelization. 
https://github.com/kbase/collections/blob/develop/src/loaders/jobs/taskfarmer/task_generator.py#L208-L209

For single program, total threads are _threads * _program_threads * NODE_THREADS. When I schedule the job, I manually set NODE_THREADS to 1. 

Does that make sense?

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -186,7 +186,13 @@ def _process_trait_counts(
     return heatmap_row, cells_meta, traits_meta
 
 
-def _run_microtrait(tool_safe_data_id: str, data_id: str, fna_file: Path, genome_dir: Path, debug: bool):
+def _run_microtrait(
+        tool_safe_data_id: str,
+        data_id: str,
+        fna_file: Path,
+        genome_dir: Path,
+        program_threads: int,
```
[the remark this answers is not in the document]
Tianhao-Gu: I don't think so. extract_traits func doesn't take threads num. 

https://github.com/ukaraoz/microtrait/blob/master/R/extract_traits.R#L22-L26

src/loaders/compute_tools/mash/mash.py
```diff
@@ -25,6 +26,7 @@ def _run_mash_single(
                                    # Save result file to source file directory. The suffix '.msh' will be appended.
                '-k', f'{kmer_size}',
                '-s', f'{sketch_size}',
+               '-p', program_threads,
```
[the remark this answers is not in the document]
Tianhao-Gu: I didn't find anywhere online the default for mash.

src/loaders/compute_tools/microtrait/versions.yaml
```diff
@@ -12,4 +12,8 @@ versions:
   - version: 0.1.3
     date: 2023-10-16
     notes: |
-      - fix cells value data type
\ No newline at end of file
+      - fix cells value data type
+  - version: 0.1.4
+    date: 2024-03-15
+    notes: |
+      - add ability to specify thread number for execution
```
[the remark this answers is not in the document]
Tianhao-Gu: yea. add note regarding to that?

src/loaders/compute_tools/mash/mash.py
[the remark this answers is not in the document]
MrCreosote:
```
~/kbase/mash/mash-Linux64-v2.3$ ./mash sketch -h

Version: 2.3

Usage:

  mash sketch [options] <input> [<input>] ...

Description:

  Create a sketch file, which is a reduced representation of a sequence or set
  of sequences (based on min-hashes) that can be used for fast distance
  estimations. Inputs can be fasta or fastq files (gzipped or not), and "-" can
  be given to read from standard input. Input files can also be files of file
  names (see -l). For output, one sketch file will be generated, but it can have
  multiple sketches within it, divided by sequences or files (see -i). By
  default, the output file name will be the first input file with a '.msh'
  extension, or 'stdin.msh' if standard input is used (see -o).

Options:

  Option     Description (range) [default]

  -h         Help

  -p <int>   Parallelism. This many threads will be spawned for processing. [1]```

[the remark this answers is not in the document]
MrCreosote: Was just worried we might be screwing things up by allowing ppl to set low threads, but this seems ok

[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/microtrait/versions.yaml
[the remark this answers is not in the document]
MrCreosote: I don't really see the need to make a release at all if there's no behavior change, but maybe I'm missing something?

src/loaders/compute_tools/microtrait/microtrait.py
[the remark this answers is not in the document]
MrCreosote: Can you add a comment saying so here in that case?

src/loaders/compute_tools/microtrait/versions.yaml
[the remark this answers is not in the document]
Tianhao-Gu: no behavior changes. But `_run_microtrait` needs take the `program_threads` additional argument now.

src/loaders/compute_tools/tool_common.py
MrCreosote: > threads number by _threads/_program_threads.

you mean instances number?

> For single program, total threads are _threads * _program_threads * NODE_THREADS. When I schedule the job, I manually set NODE_THREADS to 1.

I'm kind of lost here - is NODE_THREADS a slurm variable?

src/loaders/compute_tools/microtrait/microtrait.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍

src/loaders/compute_tools/microtrait/versions.yaml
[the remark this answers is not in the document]
MrCreosote: Won't the existing image just ignore it?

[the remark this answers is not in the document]
MrCreosote: yeah, all the changes are in compute_tools so they're built into the current image.

src/loaders/compute_tools/microtrait/microtrait.py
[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: yea. `NODE_THREADS` is a slurm variable that controls how many threads each node can run. We set it to 4 now. Because it gives us the best performance for gtdb_tk. 

And that's why we set _program_threads == _threads
https://github.com/kbase/collections/blob/develop/src/loaders/jobs/taskfarmer/task_generator.py#L208-L209

MrCreosote: Are those actual OS threads or does it really mean jobs / node?

src/loaders/compute_tools/microtrait/versions.yaml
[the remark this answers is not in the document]
Tianhao-Gu: oh you are right. remove release.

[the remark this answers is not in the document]
MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: OS threads. For example, if we set `NODE_THREADS` to 2, and there are 6 tasks, it will run 2 tasks at a time.

Tianhao-Gu: when I run eggnog, I manually change NODE_THREADS to 1.

MrCreosote: And therefore those tasks will only get one os thread each? In that case what's the point of specifying threads per program at all?

MrCreosote: > when I run eggnog, I manually change NODE_THREADS to 1.

If NODE_THREADS == os threads then all the tests you did trying different thread counts per program are invalid since the 1 instance only got one os thread

MrCreosote: This can't be right I think we must be talking past each other

Tianhao-Gu: oh no. that's not what I meant. `The THREADS variable controls the number of tasks per node`

https://docs.nersc.gov/jobs/workflow/taskfarmer/#step-3-create-a-batch-script-submit_taskfarmersl

MrCreosote: Ok, good. So for a batch job _theads is divided by _program_threads, which are equal, and therefore one instance is run per NODE_THREAD, with the instance using _threads os threads via some CLI argument.

For a single job, _threads instances are run per NODE_THREAD and each instance uses _program_threads os threads via a CLI argument, which is why you set NODE_THREADS to 1.

Right?

Tianhao-Gu: yea. That's correct. I haven't fully thought through how to streamline the logic yet, but I'll make some improvements in task_generator.py to enhance clarity.

MrCreosote: yeah, you and I should talk through a refactor before we start anything , it's a pretty odd setup at this point.

I wouldn't bother with it for now unless you have cycles other than maybe adding some docs in an appropriate place explaining how it works now

Tianhao-Gu: cool. Sounds good. I will add some notes in task_generator in my next PR. I need to do some modification there as well.

MrCreosote: :+1:

MrCreosote (approved):

PR #689, opened 2024-03-16
refactor task_generator

Tianhao-Gu:


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -52,14 +52,13 @@
 TOOLS_AVAILABLE = ['gtdb_tk', 'checkm2', 'microtrait', 'mash', 'eggnog']
 
 # estimated execution time (in minutes) for each tool to process a chunk of data
-TASK_META = {'gtdb_tk': {'chunk_size': 1000, 'exe_time': 65},
-             'eggnog': {'chunk_size': 1000, 'exe_time': 65},  # TODO: update this value after performance testing
+TASK_META = {'gtdb_tk': {'chunk_size': 1000, 'exe_time': 65, 'tasks_per_node': 4},
+             'eggnog': {'chunk_size': 100, 'exe_time': 15, 'node_time_limit': 0.5},  # Memory intensive tool - reserve more nodes with less node reservation time
```
MrCreosote, line 56: does this make it harder to reserve a set of nodes, as my understanding is that they all have to be available simultaneously as a block?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -52,14 +52,13 @@
...
              'default': {'chunk_size': 5000, 'exe_time': 60}}
 NODE_TIME_LIMIT = 5  # hours  # TODO: automatically calculate this based on tool execution time and NODE_THREADS
 MAX_NODE_NUM = 100  # maximum number of nodes to use
-# The THREADS variable controls the number of parallel tasks per node
-# TODO: make this configurable based on tool used. At present, we have set the value to 4 for optimal performance with GTDB-Tk.
-NODE_THREADS = 4
+# Used as THREADS variable in the batch script which controls the number of parallel tasks per node
+TASKS_PER_NODE = 1
```
MrCreosote, line 61:
```suggestion
TASKS_PER_NODE_DEFAULT = 1
```

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -205,8 +204,16 @@ def _create_task_list(
     program_threads: number of threads to use per task
     For instance, if "threads" is set to 128 and "program_threads" to 32, then each task will run 4 batches in parallel.
 
-    We have chosen to use "taskfarmer" for parallelization, which means that "threads" and "program_threads" should
-    have the same value. This ensures that parallelization only happens between tasks, and not within them.
+    For tools capable of processing a batch of genomes, such as GTDB-TK and checkm2,
+    we have chosen to utilize the THREADS variable within Taskfarmer batch script (submit_taskfarmer.sl) for parallelization,
+    which means that "threads" and "program_threads" should have the same value.
+    This ensures that parallelization only happens between tasks, and not within them.
```
MrCreosote, line 210: This sentence is confusing to me - we're running GTDB_tk with multiple threads aren't we?

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -52,14 +52,13 @@
 TOOLS_AVAILABLE = ['gtdb_tk', 'checkm2', 'microtrait', 'mash', 'eggnog']
 
 # estimated execution time (in minutes) for each tool to process a chunk of data
-TASK_META = {'gtdb_tk': {'chunk_size': 1000, 'exe_time': 65},
-             'eggnog': {'chunk_size': 1000, 'exe_time': 65},  # TODO: update this value after performance testing
+TASK_META = {'gtdb_tk': {'chunk_size': 1000, 'exe_time': 65, 'tasks_per_node': 4},
+             'eggnog': {'chunk_size': 100, 'exe_time': 15, 'node_time_limit': 0.5},  # Memory intensive tool - reserve more nodes with less node reservation time
              'default': {'chunk_size': 5000, 'exe_time': 60}}
 NODE_TIME_LIMIT = 5  # hours  # TODO: automatically calculate this based on tool execution time and NODE_THREADS
```
MrCreosote, line 58:
```suggestion
NODE_TIME_LIMIT_DEFAULT = 5  # hours  # TODO: automatically calculate this based on tool execution time and NODE_THREADS
```

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -205,8 +204,16 @@ def _create_task_list(
...
-    have the same value. This ensures that parallelization only happens between tasks, and not within them.
+    For tools capable of processing a batch of genomes, such as GTDB-TK and checkm2,
+    we have chosen to utilize the THREADS variable within Taskfarmer batch script (submit_taskfarmer.sl) for parallelization,
+    which means that "threads" and "program_threads" should have the same value.
+    This ensures that parallelization only happens between tasks, and not within them.
+
+    For tools that can only handle genomes individually, such as microtrait, mash and eggNOG,
+    we still use the THREADS variable for task parallelization. However, within each task, we further parallelize tool
+    execution using the 'threads' variable. For instance, with 'threads' set to 16 and 'tasks_per_node' set to 4,
+    each node will concurrently execute 4 tasks, with each task executing 16 parallel tool operations.
```
MrCreosote, line 215: I thought you said you manually set NODE_THREADS to 1 for these tools

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: just for eggNOG. Previously, we I run tools with collections data, I never need to manually update anything.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: yea. we run GTDB_TK with num of `program_threads`.

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: that's true. So we might have to come back and update this again once we have run large data sets.

MrCreosote: Yeah, we might need to make it submit several jobs with their own blocks

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: This sentence seems to contradict that

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: it won't. I still need to manually update default `threads` and `program_threads` arg in `_create_task_list` to 16. I think those should be supplied by CLI.

MrCreosote: Ok, so this PR is just a small cleanup basically?

Tianhao-Gu: yea. I haven't completely thought through how we want to treat `threads` and `NODE_THREADS_DEFAULT`

MrCreosote: yeah, let's talk though that at some point. I def have opinions on how this should work

MrCreosote: :+1:

src/loaders/jobs/taskfarmer/task_generator.py
Tianhao-Gu: 👍 how about now?

MrCreosote: I'm still confused I'm afraid. I think the problem is it's too easy to confuse "threads" because it could mean NODE_THREADS (e.g.. jobs) or os threads (e.g. _threads and _program_threads). Also, the number of threads is not necessarily correlated with the number of CPUs.

Also 
> This ensures that parallelization only happens between tasks, and not within them.

...but given _threads / _program_threads > 1 parallelization *is* happening within the task?

Tianhao-Gu: yea. if _threads / _program_threads > 1 within task, there will be <_threads / _program_threads> number of parallelizations. That was actually my original design that we use `_threads` and `_program_threads` to control parallelization. 

https://github.com/kbase/collections/blob/develop/src/loaders/compute_tools/tool_common.py#L334

MrCreosote: Sorry, by `/` I meant "given _threads and _program_threads are both set to the same number greater than 1", which obviously is not how anyone reasonable is going to interpret `/`

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -205,8 +204,19 @@ def _create_task_list(
...
+    The 'program_threads' is configured to match the CPU count per execution. To ensure each thread handles only one
+    task at a time, both 'threads' and 'program_threads' should be set to the same value. Due to the fact the underlying
+    program utilizes 'threads'/'program_threads' to determine the quantity of parallel execution on a thread.
+    Setting 'threads' and 'program_threads' the same same number ensures that parallelization only happens between tasks,
+    and not within tasks.
+
+    For tools that can only handle genomes individually, such as microtrait, mash and eggNOG,
+    we still use the THREADS variable for task parallelization. However, within each task, we further parallelize tool
+    execution using the 'threads' variable. For instance, with 'threads' set to 16 and 'tasks_per_node' set to 4,
+    each node will concurrently execute 4 tasks, with each task executing 16 parallel tool operations.
```
MrCreosote, line 218:
```suggestion
    For tools capable of processing a batch of genomes, such as GTDB-TK and checkm2,
    we have chosen to utilize the Slurm THREADS variable within Taskfarmer batch script (submit_taskfarmer.sl) for tool parallelization.
    From here on we will call this value 'tasks' rather than 'threads' to avoid confusion with operating system
    threads.
    It determines the number of simultaneous tool executions per node.
    'program_threads' is configured to match the CPU count per task execution. To ensure each task handles only one
    tool execution at a time, both 'threads' and 'program_threads' should be set to the same value, due to the fact the task generator
    program utilizes 'threads' divided 'program_threads' to determine the quantity of parallel execution in a task.
    Setting 'threads' and 'program_threads' the same same number ensures that parallelization pf tool executions
    only happens between tasks, and not within tasks.
    Each tool execution is specified to use 'threads' operating system threads.

    For tools that can only handle genomes individually, such as microtrait, mash and eggNOG,
    we still use the THREADS variable for task parallelization. However, within each task, we further parallelize tool
    execution using the 'threads' variable. For instance, with 'threads' set to 16 and 'tasks_per_node' set to 4,
    each node will concurrently execute 4 tasks, with each task executing 16 parallel tool operations.
    
    TODO: how does program_threads affect single genome tools?
```

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: I made a comment below with edits with how I *think* things work, although I'm not sure I got things right

Tianhao-Gu: 👍 Thanks. I think I will start fixing it next anyways.

MrCreosote: There's also a TODO in there that needs addressing... and some grammar and spelling mistakes I just noticed, sigh

Tianhao-Gu: 👍 fixed grammar and also add explanation of `TODO: how does program_threads affect single genome tools?`.

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -205,11 +204,27 @@ def _create_task_list(
...
-    We have chosen to use "taskfarmer" for parallelization, which means that "threads" and "program_threads" should
-    have the same value. This ensures that parallelization only happens between tasks, and not within them.
+    For tools capable of processing a batch of genomes, such as GTDB-TK and checkm2,
+    we have chosen to utilize the Slurm THREADS variable within Taskfarmer batch script (submit_taskfarmer.sl) for tool
+    parallelization.
+    From here on we will call this value 'tasks' rather than 'threads' to avoid confusion with operating system
+    threads. It determines the number of simultaneous tool executions per node.
+    'program_threads' is configured to match the CPU count per task execution. To ensure each task handles only one
+    tool execution at a time, both 'threads' and 'program_threads' should be set to the same value, due to the fact the
+    task generator program utilizes 'threads' divided 'program_threads' to determine the quantity of parallel execution
```
MrCreosote, line 214:
```suggestion
    task generator program utilizes 'threads' divided by 'program_threads' to determine the quantity of parallel execution
```

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -205,11 +204,27 @@ def _create_task_list(
...
+    in a task.
+    Setting 'threads' and 'program_threads' the same same number ensures that parallelization of tool executions
+    only happens between tasks, and not within tasks.
+    Each tool execution is specified to use 'program_threads' operating system threads.
+
+    For tools that can only handle genomes individually, such as microtrait, mash and eggNOG,
+    we still use the THREADS variable for task parallelization. However, within each task, we further parallelize tool
+    execution using the 'threads' variable. For instance, with 'threads' set to 16 and 'tasks_per_node' set to 4,
+    each node will concurrently execute 4 tasks, with each task executing 16 parallel tool operations.
+    'program_threads' is also utilized to match the operating system threads per tool execution.
```
MrCreosote, line 224: So how is this different from the batch case? We're still using THREADS (tasks) to determine how many simultaneous copies of the tool run, and threads and program threads need to be set to the same value so that only one tool instance runs per THREAD, right? Then the number of of threads each task uses is set by threads == program_threads.

Tianhao-Gu: in batch case, total num of execution parallelization is `(threads/program_threads) * THREADS ` , we want `<threads/program_threads>` to be 1. So that THREAD controls the num of parallelization. But if threads = 4, program_threads = 2 and THREAD = 3, for each task, there will be 2 (4/2) parallelization of execution and each execution takes 2 cpus. And in total there are 3 tasks in parallel. 

in comparison, in single cases, total num of execution parallelization is `threads * THREADS `.  In each task, there will be threads number of tool execution. And the node run THREADS number of tasks in parallel. 

For both cases, each execution uses program_threads number of OS threads to run.

Tianhao-Gu: As another example, in eggNOG case, we want to run 16 execution in parallel. I temporarily set THREADS = 1 and threads=16. But setting THREADS = 4 and threads=4 will give us the same performance.

MrCreosote: Is there any reason we shouldn't, in the future, make the single and batch work the same way and get rid of _threads?

Tianhao-Gu: yea. Not at all. I think that's the next thing I will be working.

src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -205,11 +204,27 @@ def _create_task_list(
...
```
MrCreosote, line 224:
```suggestion
    For tools that can only handle genomes individually, such as microtrait, mash and eggNOG,
    we still use the THREADS variable for task parallelization. However, within each task, we further parallelize tool
    execution using the 'threads' variable. For instance, with 'threads' set to 16 and 'tasks_per_node' set to 4,
    each node will concurrently execute 4 tasks, with each task executing 16 parallel tool operations.
    Each tool will also use 'program_threads' operating threads, and so in the current example if 'program_threads'
    is set to 8, the total number of operating system threads will be 4 * 16 * 8 = 512 threads. This is a significant
    difference from the batch tasks described above
```

src/loaders/jobs/taskfarmer/task_generator.py
MrCreosote: Make another stab and making this more clear

MrCreosote: :+1:

MrCreosote (approved):

PR #690, opened 2024-03-18
reduce main() length

Tianhao-Gu:


src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -788,64 +931,28 @@ def main():
...
     env = args.env
     kb_base_url = loader_common_names.KB_BASE_URL_MAP[env]
 
-    if workspace_id <= 0:
-        parser.error(f"workspace_id needs to be > 0")
-    if batch_size <= 0:
-        parser.error(f"batch_size needs to be > 0")
-    if cbs_max_tasks <= 0:
-        parser.error(f"cbs_max_tasks needs to be > 0")
+    _validate_arguments(workspace_id, batch_size, cbs_max_tasks)
```
MrCreosote, line 942: Hmm, parser.error seems more user friendly than throwing an exception

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -788,64 +931,28 @@ def main():
...
-    uid = os.getuid()
     username = os.getlogin()
-
-    job_dir = loader_helper.make_job_dir(root_dir, username)
-    collection_source_dir = loader_helper.make_collection_source_dir(
-        root_dir, loader_common_names.DEFAULT_ENV, kbase_collection, source_version
-    )
-    collections_source_dir = loader_helper.make_collection_source_dir(
-        root_dir, env, kbase_collection, source_version
+    job_dir, collection_source_dir, collections_source_dir, source_dir = _prepare_directories(
```
MrCreosote, line 945: Would it be too difficult to rename collection_source_dir / collections_source_dir here? Those var names are way too close.

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -926,16 +1020,7 @@ def main():
...
-        if conf:
-            conf.stop_callback_server()
-
-        # stop podman service if it is on
-        if proc:
-            proc.terminate()
-
-        if not keep_job_dir:
-            shutil.rmtree(job_dir)
+        _cleanup_resources(conf, proc, job_dir, keep_job_dir)
```
MrCreosote, line 1023: Even with these changes main is still > 100 LOC, which is way too long

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Maybe just pass in the parser?

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: If you have the parser you don't need to pass in the other args as well, since they're in the parser, right?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Is it guaranteed the source dir is NCBI?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Although maybe we should merge this and then take pass 2 to keep the PR small

Tianhao-Gu: I actually already did further reduce the main().

src/loaders/workspace_uploader/workspace_uploader.py
Tianhao-Gu: source_dir is the WS source dir. (e.g. sourcedata/WS/<UPA>). Rename it as something like `ws_source_dir`?

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: Now it's 60 lines, which is still on the large side but manageable

MrCreosote: :+1:

src/loaders/workspace_uploader/workspace_uploader.py
MrCreosote: I'm confused now. If I understand the renaming:

collection_source_dir -> ncbi_coll_src_dir  (e.g. collectionssource/None/GTDB/...)
collections_source_dir -> ws_coll_src_dir (e.g. collectionssource/<env>/GTDB/...)

Is that right? If so ncbi_coll_src_dir doesn't seem appropriate

Tianhao-Gu: that's correct. we use collectoin_source representing collectionssource dir. And source representing sourcedata dir. `ncbi_coll_src_dir` is confusing? what's your suggestion?

MrCreosote: yeah, ncbi doesn't appear in the directory path, and there's no guarantee in the future that the only external source will be ncbi, so at least to me the var name is confusing

Maybe `external coll_src_dir`?

Tianhao-Gu: I see. 👍

MrCreosote (commented):
LGTM but @Xiangs18 should take a look

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -807,113 +910,141 @@ def main():
...
+    try:
+        proc, conf = _setup_and_start_services(job_dir, kb_base_url, token_filepath, cbs_max_tasks, catalog_admin)
+        if not proc or not conf:
+            print("Failed to start services. Exiting ...")
             return
 
-        wtus_len = len(wait_to_upload_objs)
-        print(f"Originally planned to upload {count} object files")
-        print(f"Detected {count - wtus_len} object files already exist in workspace")
+        ws, asu_client, gfu_client = _create_ws_clients(
```
Xiangs18, line 1025: maybe rename to `_create_clients` if asu and gfu are not `ws` clients

src/loaders/workspace_uploader/workspace_uploader.py
```diff
@@ -807,113 +910,141 @@ def main():
...
+            kb_base_url, conf.callback_url, conf.token, au_service_ver, gfu_service_ver)
 
-        _prepare_skd_job_dir_to_upload(conf, wait_to_upload_objs)
-        print(
-            f"{wtus_len} objects in {conf.job_data_dir} are ready to upload to workspace {workspace_id}"
+        wait_to_upload_objs, count = _fetch_objs_to_upload(
```
Xiangs18, line 1028: if var `count` is not used below, then no need to rtn?

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py

src/loaders/workspace_uploader/workspace_uploader.py
Xiangs18: 👍

src/loaders/workspace_uploader/workspace_uploader.py
Xiangs18: 👍

Xiangs18 (approved):
LGTM!

PR #691, opened 2024-03-19
RE2022-328: refactor WSObjTuple

Tianhao-Gu:


src/loaders/workspace_uploader/upload_result.py
```diff
@@ -12,11 +12,12 @@
 
 WSObjTuple = namedtuple(
     "WSObjTuple",
-    ["obj_name", "obj_coll_src_dir", "container_internal_file"],
+    ["obj_name", "obj_file_name", "obj_coll_src_dir", "container_internal_file"],
```
MrCreosote, line 15: I could've sworn there was somewhere in the code that was really confusing / bug prone due to having the object name be the same as the file name, but I'm not seeing any fixes along those lines - I was expecting something like that.

MrCreosote (commented):
LGTM other than a question about the reason for this change

src/loaders/workspace_uploader/upload_result.py
Tianhao-Gu: Yea. There are actually several fixes wherever `.obj_name` was used, replacing it with `.obj_file_name` where appropriate.

MrCreosote: Right, but there wasn't actually a place where we had to do some weird munging in order to get things to work - I was expecting some munging code to be deleted

Tianhao-Gu: yea. I think you are referring to during `_upload_genomes_to_workspace`, when we create WSObjTuple, it used `obj_file_name` as `obj_name`. Now it should be fixed with separate `obj_name` and `obj_file_name`. It mistakenly presumed that `obj_name` and `obj_file_name` were identical, leading to confusion.

Tianhao-Gu: also fixing places like `_KEY_ASSEMBLY_FILENAME: upload_result.assembly_tuple.obj_name,`.  where `obj_name` should be  `obj_file_name`.

MrCreosote (commented):
LGTM but standard @Xiangs18 clause applies

src/loaders/workspace_uploader/upload_result.py
Xiangs18: I am still confused about what's the difference between `obj_name` and `obj_file_name`.
Also, in what place the other one cannot be used?

Tianhao-Gu: `obj_name` is currently used as workspace object name and used as the name of file to build the object. This assumption leads confusion. And it's not always true. 

For example, when we download a file associated with Assembly, `obj_name` and `obj_file_name` are no long the same. 
assembly object name: 'GCF_000979555.1_gtlEnvA5udCFS_genomic.gbff.gz_assembly'
downloaded fasta file name: 'GCF_000979555.1_gtlEnvA5udCFS_genomic.gbff.gz_assembly.fasta'.

Xiangs18: so is basically help you handle name in one place than many even though you don't always need both on later stage. Is my understanding correct?

Tianhao-Gu: yea. And most importantly we no longer assume `obj_name` and `obj_file_name` are always the same.

Xiangs18 (approved):
LGTM!

PR #708, opened 2024-04-04
Dev update deps

Tianhao-Gu:


MrCreosote:
Oh, it was pytest that was the issue. I forgot about that

MrCreosote (approved):

PR #709, opened 2024-04-04
RE2022-336: remove threads from tool execution

Tianhao-Gu:


src/loaders/compute_tools/tool_common.py
```diff
@@ -93,12 +87,8 @@ def __init__(
                                 the source version)
           --root_dir ROOT_DIR   Root directory for the collections project.
                                 (default: /global/cfs/cdirs/kbase/collections)
-          --threads THREADS     Total number of threads used by the script. (default:
-                                half of system cpu count)
-          --program_threads PROGRAM_THREADS
+          --threads_per_tool_run THREADS_PER_TOOL_RUN
```
MrCreosote, line 89: This seems like it won't work with the current task generation code, aren't changes needed there?

src/loaders/compute_tools/tool_common.py
```diff
@@ -246,46 +226,67 @@ def _get_data_ids(self):
...
                   each individual genome directory. Parser program will parse the result file in each individual genome
                   directory.
 
-        tool_callable - the callable for the tool that takes 5 arguments:
+        tool_callable - the callable for the tool that takes 6 arguments:
             * The tool safe data ID
             * The data ID
             * The input file
             * The output directory
+            * The number of threads to use for the tool run
```
MrCreosote, line 277: The docs were missing last time when adding threads to single tool runs?

src/loaders/compute_tools/tool_common.py
```diff
@@ -246,46 +226,67 @@ def _get_data_ids(self):
             data_ids = all_data_ids
         return list(set(data_ids))
 
+    def _prepare_execution(
+            self,
+            unzip: bool
+    ) -> Tuple[Path, Dict[str, Dict[str, Union[str, Path]]], List[str]]:
```
MrCreosote, line 231: Is there any reason genomes_meta can't just always be a dict[str, dict[str, Path]]

src/loaders/compute_tools/tool_common.py
```diff
@@ -298,27 +299,20 @@ def parallel_single_execution(self, tool_callable: Callable[[str, str, Path, Pat
...
-            self._execute(self._threads, tool_callable, args_list, start, False)
+            self._execute(tool_callable, args_list, start, False)
         finally:
-            if unzipped_files_to_delete:
-                print(f"Deleting {len(unzipped_files_to_delete)} unzipped files: {unzipped_files_to_delete[:5]}...")
-                for file in unzipped_files_to_delete:
-                    os.remove(file)
-
-        create_metadata_file(genomes_meta, batch_dir)
+            self._finalize_execution(unzipped_files_to_delete, genomes_meta, batch_dir)
```
MrCreosote: This seems like a logic error. With the previous set up, if _execute throws an error the unzipped files are cleaned up and the metadata file isn't created (as you would expect if there were errors). Now if an exception is thrown the metadata file will be created

src/loaders/compute_tools/tool_common.py
```diff
@@ -327,63 +321,35 @@ def parallel_batch_execution(self, tool_callable: Callable[[Dict[str, GenomeTupl
...
 
     def _execute(
             self,
-            threads: int,
             tool_callable: Callable[..., None],
             args: List[Tuple[Dict[str, GenomeTuple], Path, int, bool]] | List[Tuple[str, str, Path, Path, int, bool]],
             start: datetime.datetime,
             total: bool,
+            process_count: int = 1,
```
MrCreosote: Is there a reason not to just hardcode this into the Pool call? It seems like we don't even need the pool anymore

Tianhao-Gu: 👍 hardcoded it.

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: genomes_meta can have `tool_identifier` str field.

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: yea. I probably missed that.

MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py
MrCreosote: In which cases does it have tool id and which cases a path?

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: yea. I will need to update taskfarmer task generation code and update entrypoint script for each tool container in order of this to work. But since we didn't update containers yet, it won't break anything at least.

src/loaders/compute_tools/tool_common.py

src/loaders/compute_tools/tool_common.py
MrCreosote: is there a simpler way of using multiprocess to run serial processes than with a Pool?

src/loaders/compute_tools/tool_common.py
MrCreosote: Doesn't changes in tool common cause the images to get built? I don't really remember how that works

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: it will always have a mix of strings (`tool_identifier`) and Path (`data_file`)
https://github.com/kbase/collections/blob/develop/src/loaders/compute_tools/tool_common.py#L520-L523

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: no. Only when we update the yaml file for each tool.

MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py
MrCreosote: :+1:

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: yea. I think we don't need to use multiprocess at all. Just `tool_callable(*arg)` should work.

MrCreosote (approved):
lgtm

PR #711, opened 2024-04-05
RE2022-336: update tool containers to use threads_per_tool_run

Tianhao-Gu:


src/loaders/compute_tools/checkm2/versions.yaml
```diff
@@ -19,4 +19,11 @@ versions:
     date: 2023-11-08
     notes: |
       - create fatal error file for CheckM2
+    reference_db_version: 1.0.1
+  - version: 0.1.4
+    date: 2024-04-05
+    notes: |
+      - Remove the 'threads' and 'program_threads' parameters
+      - and introduce the capability to specify the number of threads per tool run
+      - pass job_id to the tool container and remove node_id
```
MrCreosote: I don't think you want the hyphen on 2nd line do you? I think that starts a new string

src/loaders/compute_tools/tool_common.py
```diff
@@ -449,21 +450,26 @@ def create_metadata_file(
...
-    # create working directory for each batch
-    batch_dir = work_dir / (f'{loader_common_names.COMPUTE_OUTPUT_PREFIX}{batch_id}_size_'
-                            + f'{len(data_ids)}_node_{node_id}')
+    # create a working directory for each job
+    # parser program will need `COMPUTE_OUTPUT_PREFIX` as dir name prefix to identify compute directories
+    # and `COMPUTE_OUTPUT_NO_BATCH` as dir name suffix to identify non-batch directories
+    batch_job_dir_identifier = '' if batch_tool else loader_common_names.COMPUTE_OUTPUT_NO_BATCH
+    batch_dir_name = (f'{loader_common_names.COMPUTE_OUTPUT_PREFIX}_{job_id}'
+                      f'_size_{len(data_ids)}{batch_job_dir_identifier}')
+    batch_dir = work_dir / batch_dir_name
```
MrCreosote, line 471: Does this mean the parser can no longer handle data from older runs?

src/loaders/compute_tools/checkm2/versions.yaml
Tianhao-Gu: 👍

src/loaders/compute_tools/tool_common.py
Tianhao-Gu: yes. But if we need to, we can quickly update `COMPUTE_OUTPUT_PREFIX` and `COMPUTE_OUTPUT_NO_BATCH`. Then run the parser program. Or we can keep them as is. The dir name will be something like `batch_0_size_1000_no_batch` or `batch_0_size_1000` which is also acceptable. I assume that we are unlikely to need to re-parse the current data.

Tianhao-Gu: vs `job_0_size_1000_non_batch` or `job_0_size_1000`

src/loaders/compute_tools/checkm2/versions.yaml
MrCreosote: Not sure if indentation needs to be added or not

src/loaders/compute_tools/tool_common.py
MrCreosote: Would it be difficult to rename the directories of older runs?

Tianhao-Gu: oh yea. not at all. We can rename old dirs.

MrCreosote: :+1:

src/loaders/compute_tools/checkm2/versions.yaml
Tianhao-Gu: 👍 sure.

MrCreosote (approved):

PR #712, opened 2024-04-05
RE2022-336: add confirm step befor submit job

Tianhao-Gu:


MrCreosote (approved):
We might need to make the number of cpu cores configurable but this looks ok for now

PR #746, opened 2024-07-11
use dbmen for eggnog

Tianhao-Gu:


MrCreosote:
do you think this is causing the slow performance on the worker nodes? If so why wouldn't the login nodes be the same?

Tianhao-Gu:
> do you think this is causing the slow performance on the worker nodes? If so why wouldn't the login nodes be the same?

Still testing the performance. I built the image in login node and test run it in an interactive node. The performance was even worse with this option. With this PR, I just want to build it and run via taskfarmer to see if it improves performance. The performance in general is worse in the worker nodes than in the login node. I used to cut the path to test directly with the login node. Moving forward, I probably will test things in an interactive node to get a better estimation.

Tianhao-Gu:
But I went over issues on eggnog repo. This option is what the eggnog team suggested in one of the issues.
https://github.com/eggnogdb/eggnog-mapper/issues/80

MrCreosote:
ok, sounds like you can test with the pr image so we don't need to merge this, but lmk if I'm wrong

Tianhao-Gu:
> ok, sounds like you can test with the pr image so we don't need to merge this, but lmk if I'm wrong

yea. I can test it with PR image. I will get back to you once I finished testing the image.

Tianhao-Gu:
> > ok, sounds like you can test with the pr image so we don't need to merge this, but lmk if I'm wrong
> 
> yea. I can test it with PR image. I will get back to you once I finished testing the image.

closing PR. Surpassingly, a single genome takes more than 30 minutes to finish with this option for both 2 nodes and 5 nodes running mode.

PR #747, opened 2024-08-16
add tool metadata file

Tianhao-Gu:


src/loaders/common/loader_common_names.py
MrCreosote, line 1: I assume none of this will break the collections pipline?

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -68,6 +69,13 @@ def _run_checkm2(
         fatal_tuples.append(fatal_tuple)
     write_fatal_tuples_to_dict(fatal_tuples, output_dir)
 
+    metadata = {'tool': 'checkm2',
+                'version': '1.0.1',
+                'command': command,
+                'reference_db': 'diamond_db ver unknown - saved copy in NERSC at global/cfs/cdirs/kbase/collections/libraries/checkm2/1.0.1/uniref100.KO.1.dmnd',
+                'ids_to_files': ids_to_files}
```
MrCreosote, line 79: Remind me what ids_to_files is again?

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -68,6 +69,13 @@ def _run_checkm2(
         fatal_tuples.append(fatal_tuple)
     write_fatal_tuples_to_dict(fatal_tuples, output_dir)
 
+    metadata = {'tool': 'checkm2',
+                'version': '1.0.1',
+                'command': command,
+                'reference_db': 'diamond_db ver unknown - saved copy in NERSC at global/cfs/cdirs/kbase/collections/libraries/checkm2/1.0.1/uniref100.KO.1.dmnd',
```
MrCreosote:
```suggestion
                 "reference_db": {
                     "path": "/global/cfs/cdirs/kbase/collections/libraries/checkm2/1.0.1/uniref100.KO.1.dmnd",
                     "location": "NERSC",
                     "version": None,
                     "comment": "diamond_db, ver unknown",
                 },
```

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -68,6 +69,13 @@ def _run_checkm2(
         fatal_tuples.append(fatal_tuple)
     write_fatal_tuples_to_dict(fatal_tuples, output_dir)
 
+    metadata = {'tool': 'checkm2',
+                'version': '1.0.1',
```
MrCreosote, line 73: It seems like it'd be unlikely for this to be updated when the tool version in the Dockerfile is updated. Maybe it should go in a file somewhere that both places read?

src/loaders/compute_tools/eggnog/eggnog.py
```diff
@@ -22,7 +21,7 @@ def _run_eggnog_single(
         threads_per_tool_run: int,
         debug: bool) -> None:
 
-    metadata_file = output_dir / EGGNOG_METADATA
+    metadata_file = output_dir / TOOL_METADATA
```
MrCreosote, line 24: Why does eggnog already have a metadata file while checkm2 doesn't?

src/loaders/compute_tools/eggnog/eggnog.py
```diff
@@ -22,7 +21,7 @@ def _run_eggnog_single(
...
     if metadata_file.exists():
         print(f"Skipping {source_file} as it has already been processed.")
         return
```
MrCreosote, line 27: When would this case occur?

src/loaders/compute_tools/eggnog/eggnog.py
```diff
@@ -44,9 +43,13 @@ def _run_eggnog_single(
...
     metadata = {'source_file': str(source_file),
-                'input_type': INPUT_TYPE}
-    with open(metadata_file, 'w') as f:
-        json.dump(metadata, f, indent=4)
+                'input_type': INPUT_TYPE,
+                'data_id': data_id,
+                'tool_name': 'eggnog',
+                'version': '2.1.12',
+                'command': command,
+                'reference_db': '5.0.2',}
```
MrCreosote: Both version and reference_db version seem like they could easly not be updated when they're updated elsewhere

MrCreosote: I would use the same reference db structure as I suggested earlier

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Although it seems like the path has the same issue as the version in that it could change and not be updated here. Also I suppose the container could be run at a different facility with a different path so maybe those 2 fields are ill advised

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: This comment applies to all the tools

src/loaders/common/loader_common_names.py
MrCreosote: One thing is the metadata file name is changing - will that cause problems with any earlier runs or later steps in the process?

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: that's true. Grabbing them from the Dockerfile/versions file instead of manual input? I didn't build that logic because I think we are not changing versions very often. And I feel those metadata information will unlikely to be used at all. I don't see in the near future, anyone will run the same tool with different options and analyze the diffs.

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: during a rerun - in the event of node timeout or just want to retry the failure ones. But never used before.

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: I believe we have metadata files for individual executable tools that can help us identify successful `data_ids` runs. But for batched tools (`checkm2`, `gtdb_tk`), the metadata file won't indicate which `data_ids` failed so we never add one.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: that's `data_id: source_file_path`.

src/loaders/common/loader_common_names.py
Tianhao-Gu: parser for `mash` will fail. I don't see it will affects any other tools. I think it should be fine. We will very unlikely to run parser on mash again.

MrCreosote: you mean for old mash runs only?

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Why is that valuable to put in the meta? I'm not saying it isn't, I'm just not quite sure.

src/loaders/common/loader_common_names.py
MrCreosote: Same deal for eggnog?

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: Isn't here a file which indicates which data ids have been run? I seem to remember something like that?

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: And since we're just replacing the old metadata file with the new one, which means a rerun on the old metadata file would rerun everything?

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: > Grabbing them from the Dockerfile/versions file instead of manual input?

that's what I was thinking, yeah. Is it possible for the versions file to get out of sync with the Dockerfile? Where are the version and refdb versions specified?

I think the minimum would be documenting the locations that the versions need to be changed and referring to that documentation in places where the version is likely to be changed

src/loaders/common/loader_common_names.py
Tianhao-Gu: yea parser will fail for old mash runs only. No other metadata files are being used now.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: I want to use the `data_id` as a foreign key in CDM. We can still parse it from tool results. But this is better and direct I think.

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: for batched tools? I don't think so. There is a failure file created after the tool has been run to make sure all result files are in place. But if the tool execution failed, we have no way to identify which id has been run or not.

src/loaders/common/loader_common_names.py
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: So you can look up the tool run from the genome id?

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: yea. But I am not sure why we want to do that. The current finished collections are all in good place. If we want to rerun any collection means we mean a full rerun I guess. I am not aware of any failures in any of current collections.

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: ok, it's been a long time since I looked at this code and I'm not super familiar with it anyway.

There's no code to look at partial results and figure out if any are successful? Is that even possible, or for all the tools if the execution didn't finish normally the whole batch is dead?

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: We can. But I was thinking to join tool result tables with downloaded metadata tables.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: added reminder comments in docker file and versions.yaml

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: 👍 added reminder comments in docker file and versions.yaml. I don't think we'll need to update tool versions very often. Since we built all of these tools, it hasn't happened yet for any tools.

Tianhao-Gu: 👍

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: It just never became a priority for us to develop a strategy for handling reruns and failures. There is some logic to verify that result files are in place. For failures, the parser will simply skip them. However, no rerun logic has been implemented yet.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: I think that's probably fine at the moment.

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Won't that duplicate a ton of data since it's the same for every data id?

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: If we're talking about using Lawrencium in the future I don't think we should add data that we know isn't portable

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: :+1:

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: :+1:

MrCreosote: :+1:

src/loaders/compute_tools/eggnog/eggnog.py
```diff
@@ -44,9 +43,18 @@ def _run_eggnog_single(
...
+                'data_id': data_id,
+                'tool_name': 'eggnog',
+                'version': '2.1.12',
+                'command': command,
+                "reference_db": {
+                    "path": "global/cfs/cdirs/kbase/collections/libraries/eggnog/5.0.2",
+                    "location": "NERSC",
+                    "version": "5.0.2",
+                },
+                }
```
MrCreosote:
```suggestion
                    "version": "5.0.2",
                    },
                }
```

MrCreosote (commented):
Minor formatting

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: no. it won't. Only batched tool has this. So it will be one per batch.

src/loaders/compute_tools/eggnog/eggnog.py
Tianhao-Gu: 👍

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: 👍 removed location and path.

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: I don't understand what you were saying then. Are you talking about having some sort of run ID that's associated with each row of the tool results that's linked to a single row of tool metadata in another table?

src/loaders/compute_tools/eggnog/eggnog.py
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: e.g join eggnog result to checkm2 results; join exiting metadata (the csv files we imported already) to tool results.

MrCreosote: By join do you mean a SQL JOIN or are you talking about adding more rows to a csv file?

Tianhao-Gu: SQL join. I am planning to create table for each tool. e.g. fastgenomics.eggnog_result; fastgenomics.checkm2_results. And using this data_id as foreign key.

MrCreosote: Ok, you might want to think abuot a run id since you can run the same tool multiple times on the same data_id, but :+1:

MrCreosote (approved):

PR #748, opened 2024-08-17
add bbmap tool

Tianhao-Gu:


src/loaders/jobs/taskfarmer/task_generator.py
```diff
@@ -41,7 +41,7 @@
```
Tianhao-Gu, line 41: BBMap tool operates very quickly, processing each genome under a second. Using the default configuration for task scheduling is sufficient.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,13 @@
 # KBase Collections Release Notes
 
+## 0.1.3
+
+* Added BBMap tool to the CDM pipeline.
+* Included metadata file generation after each tool's execution.
+* Updated Python library dependencies to the latest versions.
+* Standardized thread management logic across all tools.
+* Pass `job_id` to the tool container and remove `node_id`.
```
MrCreosote, line 9: Seems like there's more changes than that, for example looking at git log I see
* a int genome folder bugfix
* a biolog bugfix
* doc updates

just from the first few lines

src/loaders/compute_tools/bbmap/bbmap.py
```diff
@@ -0,0 +1,66 @@
+"""
+Run BBMap tool on a set of fna files.
+
+This tool serves a distinct purpose separate from collection tools; instead, it is suited for CDM work.
+Therefore, the parser program is not compatible with data generated by this tool.
```
MrCreosote, line 5: Can this be added at the top of the version file as well?

src/loaders/compute_tools/bbmap/bbmap.py
```diff
@@ -0,0 +1,66 @@
...
+    start = time.time()
+    print(f'Start executing BBMap for {data_id}')
+
+    metadata_file = output_dir / TOOL_METADATA
+    if metadata_file.exists():
+        print(f"Skipping {source_file} as it has already been processed.")
+        return
+
+    command = [
+        'stats.sh',
```
MrCreosote, line 31: My understanding is that bbtools does a lot of stuff. We're not interested in anything except this?

RELEASE_NOTES.md
Tianhao-Gu: 👍 But I think we should keep release notes on a very high level. Otherwise we should just auto generate the release note with every PR in between of releases.

src/loaders/compute_tools/bbmap/bbmap.py
Tianhao-Gu: 👍

src/loaders/compute_tools/bbmap/bbmap.py
Tianhao-Gu: that's the only thing the science need for now.

MrCreosote: Would it be difficult to handle if we needed to use other bbtools functionality?

src/loaders/compute_tools/bbmap/bbmap.py
MrCreosote: :+1:

src/loaders/compute_tools/bbmap/bbmap.py
Tianhao-Gu: I don't see an alternative. With the current setup we would have to build a new image for even an argument update.

RELEASE_NOTES.md
MrCreosote: Release notes should describe any bugfixes or changes in behavior that the user might encounter, and be readable to the user, which the auto generated release notes are not.

what we should do is just keep the release notes up to date with each PR, then we don't have to spelunk through the logs later

A couple other things I see in the logs

* the task generator now asks for confirmation before submitting a job
* fix working dir name prefix
  * not sure if this is visible to the user or not

src/loaders/compute_tools/bbmap/bbmap.py
MrCreosote: Hmm, we should discuss this in the standup tomorrow. 

Anyway, nothing to be done for now I guess

:+1:

RELEASE_NOTES.md
Tianhao-Gu: We've never taken release notes seriously before, but this is the first time I've written them for this repo, and now suddenly, detailed release notes have become a requirement.

I understand your point, but could you clarify why release note PRs like this was acceptable to you before but isn't now for other people?
https://github.com/kbase/collections/commit/c31fd5daa51c8fb2aa3ddf1be7e5c0fa9d831f7b

MrCreosote: > We've never taken release notes seriously before

I absolutely take release notes seriously.

The reason they're now required is that collections is released in prod

MrCreosote: The release notes you link are the initial release, so there's no changes to document

MrCreosote: Anyway, LGTM now
:+1:

MrCreosote (approved):

PR #749, opened 2024-08-19
fix a typo in metadata field

Tianhao-Gu:


src/loaders/common/loader_common_names.py
MrCreosote, line 1: Release notes for the metadata change

While you're doing that, could you update release notes for all the changes since the last release?

src/loaders/compute_tools/microtrait/microtrait.py
```diff
@@ -262,6 +272,22 @@ def _run_microtrait(
...
+    end_time = time.time()
+    run_time = end_time - start
+    print(
+        f'Used {round(run_time / 60, 2)} minutes to execute Microtrait for {data_id}')
+
+    # Save run info to a metadata file in the output directory for parsing later
+    metadata = {'source_file': str(fna_file),
+                'data_id': data_id,
+                'tool_name': 'microtrait',
+                'version': 'None',
```
MrCreosote: There must be at least a git hash right

src/loaders/compute_tools/checkm2/checkm2.py
```diff
@@ -69,14 +66,23 @@ def _run_checkm2(
...
                 'command': command,
                 "reference_db": {
                     "version": None,
                     "comment": "diamond_db, ver unknown",
                     },
-                'ids_to_files': ids_to_files}
+                'ids_to_files': make_json_serializable(ids_to_files),
+                'run_time': round(run_time, 2),
+                'batch_size': size,
+                }
```
MrCreosote: Wouldn't it make more sense, now I think about it, to handle this outside the container? Then the code would live in one place rather than having to duplicate it for every container

src/loaders/common/loader_common_names.py
Tianhao-Gu: In the main release note? Yea. I am planning to do it with the BBMap PR.

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: hmmm which block of code are you referring to? Generate the `metadata` outside of container? something like making `create_tool_metadata` takes `command, run_time, size, etc.`?

src/loaders/compute_tools/microtrait/microtrait.py
Tianhao-Gu: 👍 I updated to `ukaraoz/microtrait - release tag kb`. Do you think that makes sense?

src/loaders/common/loader_common_names.py
MrCreosote: ok, as long as the changes from this pr are included

src/loaders/compute_tools/microtrait/microtrait.py
MrCreosote: Seems like the best we can do, although maybe you should split it into git_url and release_tag?

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Generate the metadata outside of container?

right. The running code should know all this stuff, so you could abstract it all out of the containers

src/loaders/compute_tools/microtrait/microtrait.py
Tianhao-Gu: 👍

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: This is something that could probably be done later though, as I assume it'd be a big lift

src/loaders/compute_tools/microtrait/microtrait.py
MrCreosote: :+1:

src/loaders/compute_tools/checkm2/checkm2.py
Tianhao-Gu: 👍 I moved everything to `create_tool_metadata`.

src/loaders/common/loader_common_names.py
Tianhao-Gu: 👍

src/loaders/compute_tools/checkm2/checkm2.py
MrCreosote: Sorry, missed this.

I mean the metadata code doesn't run in the container at all, and is written by code after the container exits. Then you don't need to add it to every tool, it Just Works

Tianhao-Gu: That probably won’t be a good idea. If we move this logic completely outside of container, we will need if statement for each tool for tool specific metadata. And For ‘command’ metadata, we also need to return the command.

MrCreosote: I don't think so, most of the additional metadata is either source file, data id, ref db version, or ids_to_files, which is either known or can be looked up by looking in the tool directory. The only anomaly is mash, and for that there could be an method in each tool dir, that if it exists, takes the input args and returns the extra metadata.

For the command, the command = fn(input_args) so there could also be a method that does that translation that is used both in the container to start the executable and outside the container to generate the metadata.

Anyway, that's a different PR if it gets done at all.

MrCreosote (approved):

src/loaders/compute_tools/checkm2/checkm2.py

